PR #63, opened 2026-05-20
feat: US-6.5 sample from song (extract a loop and build around it)

frankbria:
## Summary

Adds the `acemusic sample <clip_id>` command (US-6.5, #37). The selected time range is extracted from the source clip and **physically combined** with a freshly generated track based on a musical role — so the sample is always audible. Each of the four roles produces a distinct placement, and a `{filename}.meta.json` sidecar records full attribution (source clip ID, source path, start/end ms, role, prompt, backend, timestamp).

The new clip is registered in the clips DB with `parent_clip_id` and `generation_mode='sample'`, so it composes with the rest of the Stage 6 iterative-generation workflow.

### Acceptance criteria (all green)
- ✅ Sample is audible in the generated output — verified by direct WAV combination + integration tests
- ✅ Different roles produce different placements — `test_roles_produce_different_output_audio` hashes all 4 outputs and asserts uniqueness
- ✅ Metadata includes attribution to the source clip and time range — `test_metadata_sidecar_written` validates every field

### Notable implementation decisions

- **Used `clip_id: int` (not file path)** to match `extend`/`cover`/`mashup` patterns. The CodeRabbit plan assumed no clip system existed; one already does via `acemusic.db.get_clip`.
- **Reused existing utilities**: `parse_time_string()` and `crop_audio()` were already in place from US-5.1, so the new code is just `combine_sample()` + `write_sample_metadata()` + the CLI command.
- **Post-processing combination over backend conditioning** — the only way to guarantee the sample is audible.
- **`format=` hint on every `AudioSegment.from_file()`** — CI lacks ffprobe; passing the format explicitly avoids the probe step (per project memory).

## Test plan

- [x] `uv run pytest tests/test_sample.py -v` — 19 new tests pass
- [x] `uv run pytest tests/test_audio.py::TestCombineSample tests/test_audio.py::TestWriteSampleMetadata -v` — 8 new unit tests pass
- [x] `uv run pytest` — full suite 538 passed, 1 skipped, 7 deselected
- [x] `uv run ruff check src/ tests/` — clean
- [x] `uv run black --check src/ tests/` — clean
- [x] CLI surface verified: `acemusic --help` lists `sample`; `acemusic sample --help` shows expected options
- [x] End-to-end demo against real WAV data — all 4 role outputs distinct, sidecar JSON correct (see `demo-us-6.5.md`)

## Summary by CodeRabbit

* **New Features**
  * New sample CLI to extract audio segments, generate role-based variations, combine sample + generated audio with role-specific placement strategies, and save outputs.
  * Metadata sidecars written alongside outputs capturing source, time range (ms), role, prompt, backend, and timestamp.

* **Documentation**
  * New demo and updated acceptance criteria describing usage, options, and expected outputs.

* **Tests**
  * Added unit and end-to-end tests for composition, metadata, backend routing, validation, and CLI behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/frankbria/auto-music-studio/pull/63?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/demo-us-6.5.md
+++ b/demo-us-6.5.md
@@ -72,7 +72,7 @@ tests/test_sample.py::TestSampleValidation::test_missing_clip_exits_one PASSED
 tests/test_sample.py::TestSampleValidation::test_invalid_role_exits_one PASSED
 tests/test_sample.py::TestSampleValidation::test_end_past_duration_exits_one PASSED
 tests/test_sample.py::TestSampleValidation::test_end_before_start_exits_one PASSED
-tests/test_sample.py::TestSampleValidation::test_negative_start_exits_one PASSED
+tests/test_sample.py::TestSampleValidation::test_invalid_time_format_exits_one PASSED
 tests/test_sample.py::TestSampleRoles::test_each_role_succeeds[loop-bed] PASSED
 tests/test_sample.py::TestSampleRoles::test_each_role_succeeds[intro-outro] PASSED
 tests/test_sample.py::TestSampleRoles::test_each_role_succeeds[rhythmic-element] PASSED
--- a/demo-us-6.5.md
+++ b/demo-us-6.5.md
@@ -1,0 +1,169 @@
+# US-6.5: Sample from song (extract a loop and build around it)
+
+*2026-05-20T19:53:46Z*
+
+## What was built
+
+US-6.5 adds the `acemusic sample <clip_id>` command, which extracts a time range from an existing clip and combines it with a freshly generated track based on a musical role. The sample is physically present in the output (so it is always audible), each role produces a distinct placement, and a `{filename}.meta.json` sidecar records the source clip, time range, role, and prompt for later tooling.
+
+This demo verifies the implementation against the three acceptance criteria from issue #37. There is no live ACE-Step server in the loop — acceptance criteria #1 and #3 are exercised end-to-end with the real `combine_sample`/`write_sample_metadata` utilities against real WAV data, and #2 is exercised both by direct utility invocation and by integration tests against a mocked backend.
+
+## Step 1: The new `sample` subcommand is wired into the CLI
+
+`acemusic --help` now lists `sample` alongside `generate`, `extend`, `cover`, `mashup`, etc.
+
+```bash
+ACEMUSIC_BASE_URL=http://localhost:8001 uv run acemusic --help 2>&1 | grep -E "^.*sample.*Extract"
+```
+
+```output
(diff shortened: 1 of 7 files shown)


frankbria:
## Response to CodeRabbit review

Applied in 94ab30f:
- ✅ Dropped unsupported `500ms` example from `--start` help
- ✅ Metadata sidecar write failure now aborts the command (provenance contract)
- ✅ `--num-clips` contract enforced: fail when ACE-Step returns fewer URLs than requested
- ✅ Renamed `test_negative_start_exits_one` → `test_invalid_time_format_exits_one`

Not applied, with reasons:
- **`@pytest.mark.integration` on all CLI test classes** — Per project convention, `@pytest.mark.integration` is reserved for tests that hit a live ACE-Step server (see `test_cover.py::TestCoverIntegration`). The CLI test classes here are mocked end-to-end tests of the Typer command surface, matching the existing pattern in `test_extend.py`, `test_mashup.py`, `test_repaint.py`, none of which use the marker. Marking them as integration would deselect them from the default `pytest` run (configured via `addopts = "-m 'not integration'"` in `pyproject.toml`), which would actually reduce coverage of the new code.
- **Lowercase `--backend` before validation** — Skipping for consistency with sibling commands: `mashup --blend`, `sounds --type`, and `generate --model` are all case-sensitive. Adding case normalization to `sample` alone would create surface inconsistency. Happy to revisit if the project decides to normalize across the board.

frankbria:
## Response to second-pass CodeRabbit review

**On the ACEMUSIC_BASE_URL gate blocking ElevenLabs-only `sample` runs:** Verified — this is a real issue, but it is a pre-existing cross-command pattern. The same gate in `main()` blocks `generate --backend elevenlabs`, `cover`, etc. with the identical error when `ACEMUSIC_BASE_URL` is unset, even though those commands also support ElevenLabs. Adding a special-case exemption for `sample` alone would create surface inconsistency — `generate --backend elevenlabs` would still fail the same way after this PR.

I'd rather track this as a follow-up issue covering all backend-routed commands at once. Not in scope for US-6.5.

**On the `@pytest.mark.integration` marker (duplicate from first pass):** Already addressed in [earlier comment](https://github.com/frankbria/auto-music-studio/pull/63#issuecomment-4502822305) — project convention reserves that marker for tests that hit a live ACE-Step server (`test_cover.py::TestCoverIntegration`). All other CLI test files (`test_extend.py`, `test_mashup.py`, `test_repaint.py`) use the same mocked pattern without the marker. Applying it here would deselect these tests from the default run.

PR #64, opened 2026-05-21
feat: US-6.6 add vocal layering and section replacement (#38)

frankbria:
## Summary

Closes #38 — implements two new CLI commands for refining specific parts of a composition:

- **`acemusic add-vocal <clip_id> --lyrics ...`** — Layers vocals onto an instrumental clip via ACE-Step's `complete` mode. Supports `--voice`, `--style`, `--output`, `--name`.
- **`acemusic replace <clip_id> --start 30s --end 45s --prompt ...`** — Regenerates a specific time range via ACE-Step's `repaint` mode. `--lock-context` (default on) crossfade-stitches the new section into the original so surrounding audio is preserved. `--no-lock-context` writes the model output directly without stitching.

Both commands create child clips with `parent_clip_id` set and the appropriate `generation_mode` (`add_vocal` / `replace`).

## Design notes

- **No client.py changes** — `AceStepClient.submit_task()` already supports `task_type="complete"` and `"repaint"` (added by US-6.1 / US-6.3). The new commands reuse those code paths.
- **Reuses US-6.3 stitching logic** — `replace` with `--lock-context` shares the same `crossfade_stitch` helper as `repaint`, only differing at the CLI/UX layer.
- **Time parsing** — uses the existing `parse_time_string()` utility from US-5.1 / US-6.3 (supports `30s`, `1m30s`, `1:30`).

## Acceptance criteria

- [x] Add-vocal produces a clip with vocals layered on the instrumental
- [x] Replace section changes only the target time range
- [x] Surrounding audio is preserved with smooth transitions (verified by `test_outside_region_preserved_when_locked`)

## Test plan

- [x] 24 new tests added (`tests/test_add_vocal.py` × 11, `tests/test_replace.py` × 13)
- [x] All 562 pre-existing tests still pass
- [x] `ruff check` and `black --check` clean
- [ ] Manual demo with a live ACE-Step server (will follow once integration env is available)

## Summary by CodeRabbit

* **New Features**
  * add-vocal: layer vocals onto instrumental clips with required lyrics; supports style, voice warning for non-default, output/name overrides and duration-probing fallback
  * replace: regenerate a specified time range with optional context-locked stitching, crossfade control, and validation of time bounds; truncated model output surfaces as an error

* **Tests**
  * Comprehensive suites for both commands covering success paths, validations/failures, duration probing, output/name overrides, lineage, locking, and stitching

* **Documentation**
  * Project status, acceptance criteria, and CLI docs updated to include the new commands and examples

[Review Change Stack](https://app.coderabbit.ai/change-stack/frankbria/auto-music-studio/pull/64?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,7 +9,7 @@ AI-powered music generation platform built on **ACE-Step-1.5** (fork: `github.co
 3. **Layer 3 — Web UI** (Stages 15–21): Next.js frontend
 4. **Layer 4 — Advanced Integrations** (Stages 22–28): VST3 plugin, music video, voice models, credits, moderation
 
-**Current progress:** Stages 1–4 complete; Stage 6 in progress (US-6.1 extend, US-6.2 cover, US-6.3 repaint, and US-6.4 mashup implemented). CLI entry point, health check, basic generation, musical parameters, quality/speed tradeoffs, model selection, sound modes, workspace management, clip extension, cover-mode restyle, section repaint with crossfade blending, and two-clip mashup with BPM alignment all implemented.
+**Current progress:** Stages 1–4 complete; Stage 6 in progress (US-6.1 extend, US-6.2 cover, US-6.3 repaint, US-6.4 mashup, US-6.5 sample, and US-6.6 add-vocal/replace implemented). CLI entry point, health check, basic generation, musical parameters, quality/speed tradeoffs, model selection, sound modes, workspace management, clip extension, cover-mode restyle, section repaint with crossfade blending, two-clip mashup with BPM alignment, loop sampling, vocal layering, and section replacement all implemented.
 
 ## Commands
 
@@ -64,6 +64,14 @@ uv run acemusic mashup 42 43                                                    
 uv run acemusic mashup 42 43 --blend sequential                                 # section-by-section blend
 uv run acemusic mashup 42 43 --blend ai-guided --style "lo-fi hip hop"          # model chooses; unifying style applied
 
+# Add vocal (US-6.6) — layer vocals onto an instrumental clip
+uv run acemusic add-vocal 42 --lyrics "[Verse]\nSing this line"                                    # default voice
+uv run acemusic add-vocal 42 --lyrics "[Chorus]\nLouder now" --voice soulful --style "breathy"     # styled vocal
+
+# Replace section (US-6.6) — regenerate a specific time range with new instructions
+uv run acemusic replace 42 --start 30s --end 45s --prompt "make this section more energetic"      # locks context (default)
+uv run acemusic replace 42 --start 1m --end 1m30s --prompt "soft strings" --no-lock-context        # use model output as-is
+
 # Workspace management (US-4.1)
 uv run acemusic workspace list                        # list all workspaces (auto-creates Default)
 uv run acemusic workspace create "My Album"           # create a new workspace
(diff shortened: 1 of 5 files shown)


frankbria:
Thanks for the thorough review — addressed in 72ca869. Item-by-item:

**1. add-vocal prompt/style redundancy** ✅ Fixed. The prompt now derives from `source.style_tags or source.title or "layer vocals over the instrumental"` (parallel to how `extend` builds its prompt), so it describes the instrumental backdrop. `--style` independently controls the vocal performance style. Added `test_submits_complete_task_with_correct_params` assertion that `kwargs["prompt"] != kwargs["style"]`.

**2. replace truncation tolerance** ✅ Fixed. Added a post-slice guard: after `middle = replace_full[start_ms:end_ms]`, we verify `len(middle) >= expected_middle_ms - 1`. The existing pre-slice check (`< end_ms - 10`) stays as a faster-failing early guard. New test `test_truncated_model_output_exits_when_locked`.

**3. Poll loop duplication** ⏭️ Skipped for this PR. Agreed it's worth a `_poll_until_complete()` helper — filing as a follow-up; touching it here would balloon the diff across `extend`, `repaint`, `mashup`, `sample`, `add-vocal`, and `replace`.

**4. voice silent no-op** ✅ Fixed help text: `"Voice identifier (Stage 25 stub — value is currently accepted but not forwarded to the model)."`

**5. pydub lazy import** ⏭️ Skipped — pydub is *not* imported at module level in `cli.py`. Both `repaint` (L2292) and `replace` use the same lazy-import pattern, so removing it would diverge from the established convention.

**6. test_default_voice_is_default** ✅ Split into `test_default_voice_suppresses_stage_25_warning` and `test_non_default_voice_emits_stage_25_warning` with concrete output assertions.

**7. Missing add-vocal duration-fallback test** ✅ Added `test_falls_back_to_probing_when_duration_missing` covering the `source.duration is None → get_duration()` path.

**Minor**:
- `source.format` ext: confirmed via grep — the DB stores `"wav"` (no leading dot) consistently across `import`, `crop`, etc., so `f"...{ext}"` is correct.
- Title naming `(vocal)` vs `(replace)`: keeping these — they read naturally and match each command's verb.

frankbria:
**Reply to CodeRabbit inline comments:**

**`--voice` is currently a no-op (L2420)** — Acknowledged. This is intentionally a Stage 25 stub, consistent with the existing pattern in `cover` (cli.py:2015–2016), which uses the same "yellow warning + no-op" pattern. The help text now explicitly states *"Stage 25 stub — value is currently accepted but not forwarded to the model"* so the user-facing contract is clear. Wiring voice IDs into the API will land with Stage 25.

**Honor `config.output_dir` before falling back (L2426)** — Intentional. The "honor `config.output_dir`" pattern is used in `generate` (L428) and `sounds` (L838), but **not** in `extend` (L1843), `cover` (L2024), `repaint` (L2224), `mashup`, `sample`, or these new commands. The deliberate split:
- `generate`/`sounds` produce fresh clips and respect `config.output_dir` as a global "downloads folder."
- Clip-derived operations (extend/cover/repaint/mashup/sample/add-vocal/replace) put the new clip in *the same workspace as its parent*, since they're logically lineage-related. A user explicitly opting out via `--output` is the documented escape hatch.

Changing this in one command would diverge from 5 others, so I'm keeping it consistent. If we want `config.output_dir` to apply across the board, that's a worthwhile follow-up that should touch all six commands at once.

**Don't copy source `model`/`inference_steps` (L2529)** — Valid critique, but matching the established pattern. `extend` (L1949–1951), `repaint` (L2343–2345), and `mashup` all copy these from the source. The argument for keeping them: they describe *the upstream lineage*, useful when tracing how a clip was originally generated (the parent_clip_id chain ends at a `generate`-mode clip whose model/inference_steps were the real generation knobs). The argument against: a `clips search --model turbo` will surface a `replace`-mode child that was actually generated by whatever the ACE-Step server picked.

Filing as a follow-up since fixing this in one command and not the others would make `clips search` results *more* confusing, not less.

frankbria:
**Reply to follow-up review:**

**Bug #2 (tolerance gap) — Fixed in 859a68b.** Aligned both guards to a 1ms tolerance:
```python
if len(replace_full) < end_ms - 1:   # was: end_ms - 10
    ...
if len(middle) < expected_middle_ms - 1:
    ...
```
Now anything that passes the first guard cannot trip the second, so the user only sees one consistent error message for truncation.

**Design #3 (poll loop duplication)** — Acknowledged. Filing as a follow-up issue; ≥5 commands now share this block and refactoring all of them belongs in its own PR rather than mixed in here.

**Design #5 (lazy pydub import)** — Re-checked: pydub is *not* imported at module level in `cli.py`. `grep -n "from pydub" src/acemusic/cli.py` shows two hits, both inside function bodies (L2292 inside `repaint`, L2704 inside `replace`). The lazy pattern was established by US-6.3 (#61); my code matches it. If we want pydub at module level, that's a 2-line follow-up that should also remove repaint's lazy import — happy to file that separately but not changing one without the other.

**New observation (no `--prompt` on `add-vocal`)** — Intentional. The user-facing contract is "give me a vocal layered over this instrumental." The backdrop prompt is just the model's hint that there's an instrumental there — not something the user should need to tune. `--style` already covers the creative knob. If real users find the auto-derived prompt limiting in practice, adding `--prompt` later is non-breaking.

Otherwise: thanks for the careful re-read.

frankbria:
Pydub import hoisted to module level in cd9336d, removing the lazy imports in both repaint and replace. Now consistent and addresses the reviewer's concern. PR should be merge-ready once CI passes on this commit.

PR #65, opened 2026-05-22
feat: US-6.7 full-song auto-extend (auto-assemble from short seed)

frankbria:
## Summary

Closes #39 — US-6.7: Get Full Song (Auto-Extend).

Adds `acemusic full-song <clip_id>` which grows a short seed clip into a complete song by planning seven canonical sections (`intro → verse → chorus → verse → chorus → bridge → outro`) and chaining seven `extend` (repaint-task) generations, each with a section-specific style hint.

## Approach

The CodeRabbit plan on the issue (from April, when the codebase was at Stage 2) proposed building a clip-store, a `concatenate_wav` utility, and an `extend` command as prerequisites. Stages 4–6 since then already shipped all of those — `extend` (US-6.1) in particular uses ACE-Step's `task_type="repaint"`, which returns the full extended audio (source + new section) in one call. So `full-song` reduces to a thin orchestrator that chains extends and stores each section as its own clip with lineage.

**New files**
- `src/acemusic/song_structure.py` (~80 lines) — pure planning module: `SONG_STRUCTURE`, `SECTION_CONFIG`, `plan_sections()`.
- `tests/test_song_structure.py` — 11 tests for the planner.
- `tests/test_full_song.py` — 14 tests for the CLI command.

**Modified**
- `src/acemusic/cli.py` — adds `_render_section_plan`, `_generate_section`, and the `@app.command("full-song")` handler. No existing commands were touched (one import added).

## Acceptance criteria

- [x] Produces a clip of approximately the target duration — `test_final_clip_duration_approximates_target` (≥ 0.8 × target) + `test_plan_total_never_overshoots_target` (never overshoots).
- [x] Song has audible structural variety (not just repetition) — `test_each_section_passes_distinct_style_hint` + `test_style_anchors_to_seed_not_previous_section`.
- [x] Interactive mode pauses for confirmation between sections — `test_interactive_mode_prompts_between_sections` (6 prompts for 7 sections) + `test_interactive_n_exits_gracefully_after_partial_chain`.
- [x] `--auto` mode completes without user intervention — `test_auto_mode_skips_all_confirmations` + `test_auto_mode_generates_seven_sections`.

## Codex review fixes (caught before push)

1. **Plan overshoot at tight margins**: With seed=58s and target=60s, the `MIN_SECTION_SECONDS=4.0` floor was unconditional, producing 7 × 4s = 28s of sections (→ 86s song). Fixed by only enforcing the floor when there's enough headroom for all sections; otherwise fall back to raw proportional distribution. Regression test: `test_plan_total_never_overshoots_target`.
2. **Style accumulation across sections**: Previously the style was built from `source.style_tags` (the previously extended clip), so each section compounded earlier hints (`"ambient, intro..., verse..., chorus..."` by section 4). Fixed by passing the seed's original `style_tags` as `base_style_tags` and using it as the stable anchor. Regression test: `test_style_anchors_to_seed_not_previous_section`.

## Usage

```bash
# Auto mode (no prompts)
acemusic full-song 42 --auto

# Interactive mode (default) — pauses after each section
acemusic full-song 42

# Custom target length and base style
acemusic full-song 42 --target-duration 180 --style "indie folk" --auto
```

Each generated section is registered as its own clip (`generation_mode="full-song"`, `parent_clip_id` chained to the previous one), so users can audit the lineage or rewind to any earlier section. The final clip's title is `"<seed> (full song)"`.

## Known limitations

- ACE-Step must run with shared filesystem access to read `src_audio_path` (same constraint as `extend`). Remote ACE-Step deployments aren't supported yet.
- Each section is a separate ACE-Step job; a 7-section song takes ~7× a single generation. No parallelism.
- Per-section style conditioning relies entirely on prompt hints; the model decides whether to honor them.

## Test plan

- [x] `uv run pytest --no-cov` → 590 passed, 1 skipped, 7 deselected (no regressions; baseline was 588).
- [x] `uv run ruff check src tests` → all checks passed.
- [x] `uv run black --check src tests` → all clean.
- [x] Cross-family review via `codex review --uncommitted` → 2 findings fixed (above).
- [ ] Manual run against live ACE-Step server (requires GPU; demo will follow once a server is available).

## Summary by CodeRabbit

* **New Features**
  * Generate full-length songs from a seed clip with a new full-song command that plans sections, renders a plan, and sequentially extends the clip into chained sections.
  * Optionally confirm between sections (interactive) or run continuously (--auto); final clip is marked as a “full song”.
  * Preserve seed style tags, with options to override style or provide lyrics for every section.
  * Automatic section planning to meet a target duration with per-section style hints and validations.

* **Tests**
  * Added comprehensive tests for generation, section planning, validations, interactive prompts, --auto behavior, style handling, lineage chaining, and failure recovery.

[Review Change Stack](https://app.coderabbit.ai/change-stack/frankbria/auto-music-studio/pull/65?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,7 +9,7 @@ AI-powered music generation platform built on **ACE-Step-1.5** (fork: `github.co
 3. **Layer 3 — Web UI** (Stages 15–21): Next.js frontend
 4. **Layer 4 — Advanced Integrations** (Stages 22–28): VST3 plugin, music video, voice models, credits, moderation
 
-**Current progress:** Stages 1–4 complete; Stage 6 in progress (US-6.1 extend, US-6.2 cover, US-6.3 repaint, US-6.4 mashup, US-6.5 sample, and US-6.6 add-vocal/replace implemented). CLI entry point, health check, basic generation, musical parameters, quality/speed tradeoffs, model selection, sound modes, workspace management, clip extension, cover-mode restyle, section repaint with crossfade blending, two-clip mashup with BPM alignment, loop sampling, vocal layering, and section replacement all implemented.
+**Current progress:** Stages 1–4 complete; Stage 6 in progress (US-6.1 extend, US-6.2 cover, US-6.3 repaint, US-6.4 mashup, US-6.5 sample, US-6.6 add-vocal/replace, and US-6.7 full-song implemented). CLI entry point, health check, basic generation, musical parameters, quality/speed tradeoffs, model selection, sound modes, workspace management, clip extension, cover-mode restyle, section repaint with crossfade blending, two-clip mashup with BPM alignment, loop sampling, vocal layering, section replacement, and full-song auto-extend all implemented.
 
 ## Commands
 
@@ -72,6 +72,11 @@ uv run acemusic add-vocal 42 --lyrics "[Chorus]\nLouder now" --voice soulful --s
 uv run acemusic replace 42 --start 30s --end 45s --prompt "make this section more energetic"      # locks context (default)
 uv run acemusic replace 42 --start 1m --end 1m30s --prompt "soft strings" --no-lock-context        # use model output as-is
 
+# Full song auto-extend (US-6.7) — grow a short seed clip into a complete song
+uv run acemusic full-song 42 --auto                                       # build a ~3.5min song from clip 42 without prompts
+uv run acemusic full-song 42                                              # interactive: confirm after each of 7 sections
+uv run acemusic full-song 42 --target-duration 180 --style "indie folk"   # custom length + base style override
+
 # Workspace management (US-4.1)
 uv run acemusic workspace list                        # list all workspaces (auto-creates Default)
 uv run acemusic workspace create "My Album"           # create a new workspace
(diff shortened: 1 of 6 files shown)


frankbria:
## Review triage

### Accepted (in commit e14ea3b)

- **CodeRabbit nit (test coverage)**: Added `(30.0, 60)` and `(60.0, 100)` to the regression test. While verifying I discovered my P1 fix was incomplete — at moderate margins like `(30, 60)` the floor still fires (since 30 ≥ 7×4=28) but bumps small sections enough to overshoot by ~2.88s. Tightened the planner to check the actual floored sum against `remaining`, so the planner never overshoots in any configuration. Good catch, thank you.
- **Claude #4 (--style help)**: Updated to clarify that `--style` replaces the seed's style tags as the per-section anchor (does not layer).
- **Claude #6 (--style test coverage)**: Added `test_style_flag_overrides_seed_style_tags` to lock the contract.

### Skipped (with reason)

- **CodeRabbit inline (rewrite planner to always exactly equal `remaining`)**: The current contract is "no overshoot; apply the floor when it fits, otherwise stay raw proportional." With the tightened condition, planned totals now match `remaining` exactly in every case I tested (30/210, 30/60, 58/60, 60/100, 30/33). A more complex iterative redistribution algorithm isn't needed.
- **Claude #1 (reverse `prompt` priority)**: The existing `extend` command at `cli.py:1862` uses the same `style_tags → title → fallback` priority; diverging here would be inconsistent with the established pattern. If the prompt-priority discussion is worth having, it should be a separate PR that updates both commands together.
- **Claude #2 (configurable poll interval)**: YAGNI — `extend` has the same hardcoded 2s interval. Worth doing as a one-line change across all commands in a follow-up if it becomes a real friction point.
- **Claude #3 (assert → graceful error)**: `plan` is hardcoded to 7 sections via `SONG_STRUCTURE`, so the assertion is unreachable today. Per "no validation for impossible scenarios."
- **Claude #5 (dataclass for 12 params)**: Premature abstraction — `_generate_section` has one caller. If we add a second consumer the dataclass becomes a clear win.
- **Minor nits (type widening, cleanup comment)**: No behavior impact; current code is consistent with the rest of the file.

CI re-running on e14ea3b.

frankbria:
## Follow-up triage (commit 042f09c)

### Accepted
- **Point 3 — assert → graceful error**: One-line change applied. CLI tool should never surface a bare traceback even on unreachable code paths.

### Skipped (with reason)
- **Point 1 — prompt/style priority**: The existing `extend` command at `cli.py:1862` uses the identical pattern (`source.style_tags or source.title or "continue the song"`). Changing only `full-song` would make the two commands behave inconsistently for the same seed. The semantic concern is legit but should be addressed as a single PR that updates both commands together. I've made a mental note; happy to file a follow-up issue if useful.
- **Docstring coverage gate**: I don't see this in any check on the PR (CodeRabbit status is SUCCESS), and all the new public functions (`plan_sections`, `_render_section_plan`, `_generate_section`, `full_song`, `Section`) have docstrings. No action needed unless CodeRabbit re-flags after this push.

Ready to merge once CI is green.

frankbria:
## Round 3 triage (commit eb2ab65)

### Accepted
- **Comment bug in `test_style_anchors_to_seed_not_previous_section`**: Real find — `SONG_STRUCTURE[3]` is the second *verse*, not the second chorus. Logic was correct, but the comment was actively misleading. Fixed.

### Skipped (with reason)
- **Design: prompt overloading**: Re-raised from prior round; same skip rationale (matches `extend`'s established pattern; should be addressed for both commands at once).
- **Minor: unreachable guard**: This is the exact change you asked for in the previous review round (Point 3). The guard is intentional for forward-compat with future SONG_STRUCTURE changes — leaving as-is.
- **Minor: compound style on intermediate clips**: True edge case if a user chains `extend` or `full-song` off a `full-song`-mode intermediate. The right fix is for `extend` to also anchor to a stable seed reference (parallel to what we did here). Out of scope for this PR.
- **Test gap: mid-pipeline `duration=None`**: Defensive test for a scenario the guard handles cleanly. Low value vs cost.

Ready to merge.

PR #66, opened 2026-05-25
feat: US-7.1 single clip export (WAV/FLAC/MP3/WAV32)

frankbria:
## Summary

Adds `acemusic export <clip_id> --format wav|wav32|flac|mp3 [--output PATH]` to export a stored clip into the chosen audio format. Default output is `./<clip-title-slug>.<ext>`.

| Format | Spec |
|--------|------|
| `wav`  | 48 kHz / 24-bit PCM (pcm_s24le) |
| `wav32`| 48 kHz / 32-bit float PCM (pcm_f32le) |
| `flac` | Lossless FLAC at source sample rate |
| `mp3`  | 320 kbps CBR |

## Acceptance Criteria — Demo Mapping

- [x] Each format produces a valid, playable file → `test_export_*_invokes_pydub_with_*` (mocked encode) + real WAV roundtrip + manual demo (Phase 11)
- [x] WAV output is 48 kHz / 24-bit → `test_export_wav_real_roundtrip_is_48k_24bit` asserts via stdlib `wave` module (`getframerate()==48000`, `getsampwidth()==3`)
- [x] File is named after the clip title by default → `test_default_output_path_uses_slugified_clip_title`

## Adapted from CodeRabbit Plan

The CodeRabbit-generated plan (issue body) assumed pydub was not installed and the clip DB didn't yet exist. Reality after Stage 4–6 merged:
- `pydub>=0.25.1` already a dep — skipped that task
- `db.get_clip(clip_id: int)` already exists; `clip_id` is **int**, not string
- `get_duration()` already handles MP3/FLAC via generic `mutagen.File()` — no extension needed
- Used `EXPORT_FORMATS` (distinct constant) to avoid clashing with existing input-side `SUPPORTED_FORMATS`
- Pydub `from_file(..., format=...)` hint passed explicitly (project convention to avoid ffprobe)
- Export does **not** create a new Clip DB record (issue is export-to-disk only)

## Known Limitations

- The real-pydub WAV roundtrip test (`test_export_wav_real_roundtrip_is_48k_24bit`) is skipped when ffmpeg is not on PATH. CI ubuntu-latest does not install ffmpeg, so it'll be skipped there. FLAC/MP3/WAV32 codec verification is covered by mock-based tests that don't need ffmpeg.
- The `export` command does not insert a new Clip record into the DB (intentional — issue asks only for file output). If we later want to track exports as derived clips, a follow-up issue can add that.
- Pydub's `--output` path is taken verbatim; no extension normalization (e.g., `--format mp3 --output foo.wav` will produce an MP3 file named `foo.wav`).

## Test plan

- [x] `uv run pytest tests/test_audio.py::TestExportAudio tests/test_export_cli.py` — 20 tests pass
- [x] `uv run pytest --no-cov` — full suite 611 passed, 1 skipped (was skipped before this change)
- [x] `uv run ruff check .` — passes
- [x] `uv run black --check .` — passes
- [ ] Demo (Phase 11): run `acemusic export` against a real clip, play each output format
- [ ] CI green on push

Closes #40

## Summary by CodeRabbit

* **New Features**
  * Export audio clips to WAV (48 kHz/24-bit), WAV32 (32-bit float), FLAC, and MP3 (320 kbps); format validation and clear error reporting. New CLI export command with sensible default naming, extension mapping, and automatic parent-directory creation; prints success and file size.

* **Tests**
  * Added unit and CLI tests covering format validation, export behaviors, filename/output handling, and error cases.

* **Documentation**
  * Added CLI export usage examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/frankbria/auto-music-studio/pull/66?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -84,6 +84,13 @@ uv run acemusic workspace switch "My Album"           # set active workspace
 uv run acemusic workspace rename "My Album" "Debut"   # rename a workspace
 uv run acemusic workspace delete "Debut"              # delete (prompts if clips exist)
 uv run acemusic workspace delete "Debut" --force      # delete without confirmation
+
+# Export (US-7.1) — export a clip to disk in the chosen audio format
+uv run acemusic export 42                                              # default: WAV 48kHz/24-bit → ./<slug>.wav
+uv run acemusic export 42 --format wav32                               # 48kHz / 32-bit float WAV
+uv run acemusic export 42 --format flac                                # lossless FLAC
+uv run acemusic export 42 --format mp3                                 # 320 kbps MP3
+uv run acemusic export 42 --format flac --output /path/to/out.flac     # explicit destination path
 ```
 
 ## Architecture
--- a/src/acemusic/audio.py
+++ b/src/acemusic/audio.py
@@ -476,7 +476,6 @@ def export_audio(source_path: str | Path, dest_path: str | Path, fmt: str) -> Pa
     elif fmt == "mp3":
         audio.export(str(dest_path), format="mp3", bitrate="320k")
     else:
-        # Guard against EXPORT_FORMATS being extended without a matching branch.
         raise AssertionError(f"Unhandled export format: {fmt!r}")
 
     return dest_path
--- a/src/acemusic/audio.py
+++ b/src/acemusic/audio.py
@@ -12,6 +12,10 @@ if TYPE_CHECKING:
 
 SUPPORTED_FORMATS: set[str] = {".wav", ".flac", ".mp3", ".ogg", ".aac", ".aiff"}
 
+# Output formats accepted by export_audio() (US-7.1). Distinct from SUPPORTED_FORMATS
+# above, which lists *input* file extensions accepted by `import`.
+EXPORT_FORMATS: tuple[str, ...] = ("wav", "wav32", "flac", "mp3")
+
 _PITCH_CLASSES: list[str] = ["C", "C#", "D", "D#", "E", "F", "F#", "G", "G#", "A", "A#", "B"]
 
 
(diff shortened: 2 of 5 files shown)


frankbria:
## Phase 11 Demo — Acceptance Criteria with Outcome Evidence

Demoed against a real 1-second 440Hz sine WAV (44.1kHz/float source) in an isolated `HOME` so no live workspace was touched. Each acceptance criterion is mapped to outcome evidence, not just exit codes.

### AC1: Each format produces a valid, playable file

Probed with `soundfile.info()` / `mutagen.File()` after export:

| Format | Output | Codec / Rate / Channels / Duration |
|--------|--------|------------------------------------|
| `wav`   | acceptance-demo-track.wav (288 KB)  | PCM_24, 48000 Hz, 2 ch, 1.000 s |
| `wav32` | acceptance-demo-track.wav (384 KB)  | FLOAT, 48000 Hz, 2 ch, 1.000 s |
| `flac`  | acceptance-demo-track.flac (21 KB)  | FLAC (lossless), 44100 Hz, 2 ch, 1.000 s |
| `mp3`   | acceptance-demo-track.mp3 (43 KB)   | MP3, 44100 Hz, 320 kbps CBR, 1.045 s |

### AC2: WAV output is 48 kHz / 24-bit

Verified via stdlib `wave`:
```
framerate = 48000 Hz   (expect 48000)
sampwidth = 3 bytes (24-bit)  (expect 3 bytes / 24-bit)
channels  = 2
nframes   = 48000
```

### AC3: File is named after the clip title by default

Clip title `"Acceptance Demo Track"` → output file `acceptance-demo-track.wav` (via `make_slug()`).

### Extras demoed

- `--output /tmp/us71-output/custom-name.flac` → wrote to the explicit path
- `export 9999` → `Error: clip 9999 not found.` exit 1
- `export 1 --format ogg` → `Error: invalid --format 'ogg'. Allowed values: wav, wav32, flac, mp3` exit 1
- `export --help` shows the typer help with all options and the four-format list

All three acceptance criteria are demonstrated with outcome evidence (sample rate, bit depth, codec verified by independent probe), not just successful exit codes.

frankbria:
## Review responses

**Items addressed in 543d2ec:**
- Dead `mock_seg.set_frame_rate.return_value = mock_seg` stub removed from `test_export_wav32_uses_float_codec_parameters`.
- Dead `else 0` branch simplified to `size = dest.stat().st_size` (export_audio either succeeds or raises, so the file is always present here).

**Item declined (with reason):** `format` parameter name in `export_cmd`.

This is the consistent convention across every other typer command in `cli.py` (`generate`, `sounds`, etc. all use `format: str = typer.Option(...)`). Renaming just this one would be inconsistent. Either we keep the convention or change all of them in a follow-up — single-command rename is worse than leaving it. Marking as won't-fix for this PR; happy to tackle a global rename in a separate PR if desired.

frankbria:
@coderabbitai — addressed in 24a3018 and prior commits:

**Addressed:**
- ✅ Parent-dir creation (`dest.parent.mkdir(parents=True, exist_ok=True)`) added in 874d1ea, covered by `test_output_creates_missing_parent_directories`
- ✅ Two inline `#` comments removed in 24a3018 per repo no-comments policy (the mkdir comment in cli.py and the EXPORT_FORMATS guard comment in audio.py)

**Declined:**
- ❌ Adding `@pytest.mark.integration` to `test_export_wav_real_roundtrip_is_48k_24bit`. My `skipif(shutil.which("ffmpeg") is None)` is preferable here — it runs the test in every environment that has ffmpeg (including local dev) and only skips where ffmpeg is unavailable. The integration marker would exclude it from default runs even where ffmpeg is present, losing local-dev validation of the codec parameters.

The test docstring on `test_export_unhandled_branch_raises_assertion` is kept — pytest test docstrings document intent and aren't subject to the inline-comment policy.

frankbria:
Quick triage on the latest review pass:

**Applied (4cedfd0):**
- ✅ Trimmed `# US-7.1 export targets;` → `# Export-side formats;` — drops the story-ref prefix that would rot over time, keeping just the useful "why two constants" disambiguation.

**Declined:**

- ❌ `Optional[Path]` → `Path | None` (Claude). The data: cli.py uses `Optional[Path]` **14 times** and `Path | None` only **1 time**; `Optional[...]` overall **99 times**. The PEP 604 union syntax is the minority style in this file. Renaming one parameter would *introduce* inconsistency rather than remove it. Global migration is a separate PR.

- ❌ Remove the EXPORT_FORMATS comment entirely (CodeRabbit). The comment documents a non-obvious WHY (two similar constants in the same file with subtly different purposes — input vs. output formats). Per the broader policy `Only add a comment when the WHY is non-obvious`, this qualifies. The file already has multi-line comments in `crop_audio`, `combine_sample`, etc.; the no-comments rule is enforced via judgment, not absolutely.

PR #92, opened 2026-06-01
fix: retry transient errors while polling generation status

frankbria:
## Summary
The follow-up to the ACE-Step update work. The CLI poll loops aborted on a **single** `query_result` failure, so a cold-start blip (server briefly blocking while it loads models) killed the whole job with `Error polling status: ... timed out` — even though generation then succeeded server-side. This hardens polling.

## Changes
- **New shared helper `_poll_until_complete()`** — tolerates transient connection/timeout errors and **retries within the `ACEMUSIC_POLL_TIMEOUT` budget**; aborts immediately on a genuine API error; gives up after **5 consecutive** connection failures so a truly-down server fails fast (doesn't wait out the whole budget).
- **Replaced all 10 duplicated poll loops** (generate, sounds, extend, full-song, cover, repaint, add-vocal, replace, mashup, sample) with the helper — fixes the fragility everywhere and removes ~200 lines of duplication.
- **Fixed `_is_connection_error`** to match `"timed out"` (httpx renders read/connect timeouts that way; the previous `"timeout"` keyword never matched it — so timeouts weren't even classified as connection errors).

## Test Plan
- [x] New unit tests: transient error → retried & succeeds; persistent errors → bounded abort; genuine API error → immediate abort (no retry)
- [x] **697 unit tests pass** (3 new), 1 skipped, 15 integration deselected
- [x] Mutation check: forcing `_is_connection_error→False` and removing the consecutive cap each fail the right test
- [x] `ruff` + `black` clean

## Notes
- Minor UX change: a few command-specific failure prefixes (e.g. "Cover failed") are now the generic "Generation failed" from the helper; the underlying error message is preserved and no test asserts the old wording.
- Real-world origin: confirmed during the ACE-Step server update — `acemusic generate` works, but a cold first request could trip the old single-shot 10s poll timeout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Polling now distinguishes transient connection/timeouts from non-retryable errors, preventing undue fast-failures and improving reliability.

* **Improvements**
  * Unified polling and retry behavior across multiple CLI commands, with clearer timeout semantics and visible progress status during long-running tasks.

* **Tests**
  * Added CLI tests covering retry-on-transient-errors, tolerating repeated timeouts, bounded fast-fail on persistent connection failures, and immediate abort on non-retryable API errors.

### DIFF
--- a/src/acemusic/cli.py
+++ b/src/acemusic/cli.py
@@ -249,8 +249,75 @@ def _is_connection_error(exc: AceStepError) -> bool:
     """Return True if the AceStepError is a connection/timeout failure (not an API error)."""
     msg = str(exc).lower()
     return any(
-        keyword in msg for keyword in ("connection refused", "connect", "timeout", "unreachable", "name or service")
+        keyword in msg
+        # "timed out" and "timeout" are distinct substrings: httpx renders read/
+        # connect timeouts as "timed out", so both must be matched.
+        for keyword in ("connection refused", "connect", "timeout", "timed out", "unreachable", "name or service")
     )
+
+
+_MAX_CONSECUTIVE_POLL_ERRORS = 5
+
+
+def _poll_until_complete(
+    ace_client: AceStepClient,
+    task_id: str,
+    status_bar,
+    label: str,
+    poll_timeout: float,
+    poll_interval: float = 2.0,
+) -> dict:
+    """Poll ``query_result`` until the task completes; return the result dict.
+
+    Transient connection/timeout errors (e.g. the server briefly blocking while it
+    loads models on a cold start) are tolerated and retried within the overall
+    ``poll_timeout`` budget, rather than aborting the whole job on a single blip.
+    A genuine API error aborts immediately, and a run of consecutive connection
+    failures (``_MAX_CONSECUTIVE_POLL_ERRORS``) aborts so a truly-down server fails
+    fast instead of waiting out the full budget.
+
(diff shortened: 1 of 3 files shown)


frankbria:
## Review triage (codex, 3 rounds)

**P1 (round 1) — consecutive cap could abort a long cold start:** fixed — timeouts now retry for the full `ACEMUSIC_POLL_TIMEOUT` budget; only hard connection failures count toward the fast-fail cap.

**P2 (round 2) — API errors misclassified as transient via message text:** fixed — added typed `AceStepConnectionError` (raised for `httpx.RequestError`); the poll helper branches on `isinstance` + `exc.is_timeout`, so a 500 whose body mentions "connection" surfaces immediately instead of being retried/masked.

**P2 (round 3) — connect timeouts treated as "reachable":** fixed — `is_timeout` is now `isinstance(exc, httpx.ReadTimeout)` only; connect timeouts (blackholed host) leave it False and fast-fail after the cap.

**P1 (round 3) — "ElevenLabs fallback regressed on poll-time outages": rebutted (not a regression).** The original poll loops (`0e30e90`) already raised `typer.Exit` on poll errors, never `AceStepError`, so the fallback at `cli.py:563-575` only ever applied to **submit-time** failures — never poll-time. My helper preserves that exactly. Verified via `git show 0e30e90:src/acemusic/cli.py`. (Adding poll-time fallback would be a separate enhancement, out of scope here.)

Validation: 698 unit tests pass; CI green on 3.11/3.12; mutation checks on the retry policy pass.

Issue #95, opened 2026-06-02
ElevenLabs: per-operation backend selector (--backend)


frankbria:
Part of #94 (ElevenLabs first-class backend).

## Summary
Add a per-operation backend selector so ElevenLabs can be chosen as a primary engine, not just a fallback. Foundation for the rest of the epic.

## Scope
- `--backend auto|ace-step|elevenlabs` flag on backend-capable commands (and/or a config default `ACEMUSIC_BACKEND`).
- `auto` = current behavior (ACE-Step, fall back to ElevenLabs on transport failure — see also #93).
- Explicit `elevenlabs` routes the operation to ElevenLabs when supported; if an operation is unsupported by the chosen backend (e.g. MIDI), exit with a clear message rather than silently switching.
- Centralize backend resolution (one helper) so all commands share consistent routing/validation.

## Acceptance criteria
- [ ] `--backend` (and config default) selects the engine for supported operations
- [ ] Choosing a backend that can't perform an operation fails with an actionable message (no silent fallback for explicit choices)
- [ ] `auto` preserves today's ACE-Step-first behavior
- [ ] Unit tests cover routing for each backend value + the unsupported-operation path

## Notes
Extends **US-2.5 (backend selection)**; aligns with **Layer 2 / Stage 11 (Compute Routing)**.

PR #101, opened 2026-06-02
fix: ElevenLabs fallback on poll-time ACE-Step outages (#93)

### LINKED ISSUE #93: ElevenLabs fallback on poll-time ACE-Step outages (not just submit-time)
## Summary

The ElevenLabs auto-fallback only triggers on **submit-time** ACE-Step failures. If `submit_task()` succeeds but ACE-Step then drops out **during polling** (`/query_result`), the command fails with exit 1 instead of falling back to ElevenLabs — even when `ELEVENLABS_API_KEY` is configured.

This is longstanding behavior (it predates the poll-retry refactor in #92), surfaced during the codex review of #92. The retry hardening in #92 means poll-time outages are now retried within `ACEMUSIC_POLL_TIMEOUT` and only a *persistent* failure exits — but on that persistent failure it still does not fall back.

## Current behavior

- `generate()` wraps the ACE-Step path in `try/except AceStepError` and falls back to ElevenLabs only when `_is_connection_error(exc)` is true (`src/acemusic/cli.py:563-575`).
- The shared poll helper `_poll_until_complete()` raises `typer.Exit(code=1)` on timeout / persistent connection failure (`src/acemusic/cli.py:262-318`), which is **not** an `AceStepError`, so it bypasses the fallback `except`.
- Net: submit-time connection failure → fallback ✅; poll-time connection failure → hard exit ❌.

## Desired behavior

When ACE-Step is reachable at submit but the job cannot be completed due to a **transport/connection** failure during polling (after retries are exhausted), and `ELEVENLABS_API_KEY` is set:
- Fall back to ElevenLabs (consistent with submit-time behavior), printing the same "ACE-Step unavailable — falling back to ElevenLabs" notice.
- A **genuine API error** or a reported job **failure** (status `failed`) must still exit immediately — do NOT fall back (the server worked; the request was bad). Only transport/connection failures should fall back.

## Suggested approach

- Have `_poll_until_complete()` raise a typed transport error (e.g. re-raise `AceStepConnectionError`) on persistent connection failure / timeout-budget-exhaustion, instead of `typer.Exit`, so callers can decide whether to fall back.
- Keep raising `typer.Exit` (or a non-connection `AceStepError`) for reported `failed` status and genuine API errors so those do not trigger fallback.
- Update `generate()`'s fallback `except` to catch the poll-time transport error and route to ElevenLabs (gated on `ELEVENLABS_API_KEY`).
- Consider whether the other 9 commands that use the helper (cover, extend, mashup, …) should fall back too, or whether fallback remains generate/sounds-only (ElevenLabs has no cover/extend/etc. equivalent — likely generate/sounds only).

## Acceptance criteria

- [ ] With `ELEVENLABS_API_KEY` set, a persistent **connection/timeout** failure during polling falls back to ElevenLabs and produces output (exit 0)
- [ ] A reported job `failed` status still exits 1 (no fallback)
- [ ] A genuine API/HTTP error during polling still exits 1 (no fallback)
- [ ] Without `ELEVENLABS_API_KEY`, poll-time connection failure exits 1 with the existing actionable message
- [ ] Unit tests cover all four paths (mocked `query_result` raising `AceStepConnectionError` vs returning `failed` vs `AceStepError`)

## References

- Related story: US-2.5 (Backend selection and ElevenLabs cloud fallback)
- Surfaced by: codex review on #92; current fallback at `src/acemusic/cli.py:563-575`, poll helper at `src/acemusic/cli.py:262-318`

frankbria:
## Summary
Closes #93. `generate` now falls back to ElevenLabs when ACE-Step drops **during polling**, not only at submit — closing the gap surfaced by the codex review of #92.

- **`_poll_until_complete(..., raise_on_transport_error=False)`**: when set, a **persistent connection failure** (consecutive-failure cap — ACE-Step unreachable) re-raises `AceStepConnectionError` so the caller can fall back. Default `False` ⇒ the **other 9 poll-helper callers are unchanged**.
- **`_generate_via_ace_step`** forwards the flag; **`generate`** passes `True`, and its existing `except` (from #95) routes the transport error: `auto` → ElevenLabs fallback; explicit `ace-step` → exit with "use --backend auto"; no key → "Set ELEVENLABS_API_KEY". Connection check also matches `AceStepConnectionError` by type.
- **Timeouts do NOT fall back** (codex P2): a timeout means ACE-Step is slow (cold start/long job), not down — it exits with the existing "Timed out" message, preserving the in-flight job and ACE-Step-only flags. Only connection failures fall back.

## Acceptance Criteria
- [x] Poll-time connection failure + `ELEVENLABS_API_KEY` (auto) → fall back, exit 0
- [x] Reported `failed` status → exit 1, no fallback
- [x] Genuine API error during polling → exit 1, no fallback
- [x] No key + poll-time connection failure → exit 1, actionable message
- [x] Unit tests cover all four paths (+ explicit ace-step poll failure)

## Test Plan
- [x] 719 unit tests pass; 5 new (`TestPollTimeFallback`); lint + format clean
- [x] Mutation checks: disabling the raise-on-cap and making `failed` status raise each fail the right test
- [x] Cross-family review: **codex** — P2 (timeout-no-fallback) + P3 (host-independent tests) addressed

- **Generate-only**, per the issue's open question: the other backend-capable commands (cover/extend/mashup/…) keep exit-on-failure — ElevenLabs has no equivalent operation, so fallback there is meaningless.
- Fallback triggers on **connection** failures (server unreachable), not timeouts (slow-but-up) — deliberate (codex P2).

Closes #93

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fallback to ElevenLabs when ACE-Step experiences connection failures during the polling phase
  * Enhanced error messages to specify missing API key requirements
  * Refined handling to distinguish between connection issues, timeouts, and API errors

### DIFF
--- a/src/acemusic/cli.py
+++ b/src/acemusic/cli.py
@@ -267,6 +267,7 @@ def _poll_until_complete(
     label: str,
     poll_timeout: float,
     poll_interval: float = 2.0,
+    raise_on_transport_error: bool = False,
 ) -> dict:
     """Poll ``query_result`` until the task completes; return the result dict.
 
@@ -281,13 +282,27 @@ def _poll_until_complete(
       (or a timeout) resets the counter.
     - A **genuine API error** aborts immediately.
 
-    Raises ``typer.Exit(code=1)`` on timeout, reported failure, or abort.
+    Transport outages (the consecutive-failure abort and timeout-budget
+    exhaustion) normally print and ``typer.Exit(code=1)``. When
+    ``raise_on_transport_error`` is True, they instead re-raise
+    ``AceStepConnectionError`` so the caller can decide whether to fall back to
+    another backend (used by ``generate``). Reported ``failed`` status and
+    genuine API errors always exit — they are never a transport problem and must
+    not trigger a fallback.
+
+    Raises:
+        AceStepConnectionError: transport failure when ``raise_on_transport_error``.
+        typer.Exit: otherwise on timeout/failure/abort.
     """
     start = time.monotonic()
     consecutive_conn_failures = 0
     while True:
         elapsed = time.monotonic() - start
         if elapsed >= poll_timeout:
+            if raise_on_transport_error:
+                raise AceStepConnectionError(
+                    f"ACE-Step timed out after {poll_timeout:.0f}s while polling", is_timeout=True
+                )
             console.print(f"[red]Timed out after {poll_timeout:.0f} seconds.[/red]")
             raise typer.Exit(code=1)
 
@@ -303,6 +318,8 @@ def _poll_until_complete(
             else:
                 consecutive_conn_failures += 1
                 if consecutive_conn_failures >= _MAX_CONSECUTIVE_POLL_ERRORS:
+                    if raise_on_transport_error:
+                        raise
                     console.print(
                         f"[red]Error polling status after {consecutive_conn_failures} consecutive "
                         f"connection failures: {exc}[/red]"
@@ -312,7 +329,7 @@ def _poll_until_complete(
             time.sleep(poll_interval)
             continue
         except AceStepError as exc:
-            # Genuine API/HTTP error — surface it immediately.
+            # Genuine API/HTTP error — surface it immediately (never a fallback case).
             console.print(f"[red]Error polling status: {exc}[/red]")
             raise typer.Exit(code=1)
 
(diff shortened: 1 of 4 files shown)


frankbria:
## Final triage

- **CodeRabbit Major — "Don't route submit-time timeouts into fallback":** **resolved.** Generate's fallback decision no longer uses `_is_connection_error` directly; it uses `_should_fall_back(exc)`, which falls back only on a *hard* connection failure and explicitly excludes timeouts (`AceStepConnectionError` with `is_timeout=False`, or a non-timeout connection message). Timeouts (submit- or poll-time) now exit without switching backends. Added `test_submit_timeout_does_not_fall_back`. (The remaining inline comment is anchored to commit `1d5662c`, before the fix in `ec593e9`.)
- **codex P2/P3:** addressed earlier — timeout-budget exhaustion exits (no fallback); poll-failure tests patch `load_config` (host-independent).
- **claude-review:** dead import removed; `AceStepConnectionError ⊂ AceStepError` invariant pinned by a test.

Demo: poll-time fallback verified **live** (killed the server mid-poll → real ElevenLabs MP3 with a key; "Set ELEVENLABS_API_KEY" exit without). failed-status / API-error paths covered by deterministic unit tests. 721 unit tests pass; CI green 3.11/3.12; GitGuardian green.

Issue #97, opened 2026-06-02
ElevenLabs: stem separation backend (alt to demucs)


frankbria:
Part of #94 (ElevenLabs first-class backend).

## Summary
Add ElevenLabs stem separation as an alternative backend for `acemusic stems`, via `POST /v1/music/stem-separation` (accepts an arbitrary uploaded audio file). Useful when no local GPU/demucs is available.

## Scope
- New client method: upload a clip (multipart `file`) to `/v1/music/stem-separation`, default `stem_variation_id=six_stems_v1`, with `output_format`.
- Wire into the stems command behind the backend selector (`--backend elevenlabs`).
- Persist the returned stems as child clips with lineage, matching the existing demucs-based stems output shape (US-5.3) so downstream (DAW export, etc.) is unaffected.
- Note: ElevenLabs offers a six-stem variation; map/label stems consistently with the existing four-stem convention (vocals/drums/bass/other) or document the additional stems.

## Acceptance criteria
- [ ] `stems <clip> --backend elevenlabs` uploads the clip and produces playable stem files linked to the parent
- [ ] Stem outputs integrate with existing clip/lineage storage and naming
- [ ] Six-stem vs four-stem labeling is handled and documented
- [ ] Errors (auth, file too large, latency/timeout) surface clearly
- [ ] Unit tests (mocked client) cover the upload + persistence path

## Notes
Alternative to local demucs (US-5.3); cloud, no CUDA. ElevenLabs output is MP3 — stems will be MP3 quality, not lossless.

PR #104, opened 2026-06-04
test: real-API integration test for ElevenLabs separate_stems (#97)

frankbria:
## Summary
Follow-up to #103: now that a valid `ELEVENLABS_API_KEY` is configured locally, add an integration test that exercises `separate_stems()` against the live API.

- Verifies the six documented stem labels come back from a real separation
- Pins the undocumented ZIP-entry-naming assumption from #97 — if ElevenLabs changes archive filenames, this test catches it
- Each stem must decode as MP3 with >1s of audio
- Skips when `ELEVENLABS_API_KEY` is absent (CI), same as the existing `TestElevenLabsIntegration` tests

## Validation
- Ran against the live API: 6 stems (vocals, drums, bass, guitar, piano, other), all playable 4s MP3s at 44.1 kHz
- Unit suite unaffected (54 passed); ruff + black clean

- Consumes API credits per run (consistent with the existing integration tests)
- None otherwise

## Summary by CodeRabbit

* **Tests**
  * New integration test added to validate audio stem separation functionality. Verifies that separated audio stems are correctly generated in playable format, meet quality duration thresholds, and conform to expected stem categorization structure and naming conventions.

### DIFF
--- a/tests/test_elevenlabs_client.py
+++ b/tests/test_elevenlabs_client.py
@@ -657,3 +657,31 @@ class TestElevenLabsIntegration:
 
         segment = AudioSegment.from_file(io.BytesIO(audio), format="mp3")
         assert segment.duration_seconds > 1.0
+
+    def test_separate_stems_returns_six_playable_stems(self, tmp_path):
+        """Integration: separate_stems() on a real clip yields the six documented labels as decodable MP3s.
+
+        Also pins the undocumented ZIP-entry-naming assumption (#97): the real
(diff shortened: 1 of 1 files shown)


frankbria:
Re: hardcoded `output_format` — intentional: the test decodes stems with `format="mp3"`, so the request format must be pinned to an MP3 variant for the assertion to be deterministic (an env override like `pcm_44100` would break the decode step, not the contract under test). Same pattern as `test_compose_plan_roundtrip_returns_playable_audio`.

PR #151, opened 2026-06-16
feat(api): compute status endpoint (US-11.4)

### LINKED ISSUE #125: US-11.4: Compute status endpoint
## User Story

**As a** musician, **I want to** check the status of available compute targets, **so that** I know whether my local GPU is busy and whether remote is available before generating.

## Description

A status endpoint that aggregates health information from both local and remote compute targets into a single response.

## Functional Requirements

- `GET /api/v1/compute/status` returns combined status of local and remote targets
- Local status: available (bool), GPU name, VRAM total/used, active jobs, loaded models
- Remote status: available (bool), provider, endpoint_id, active workers, max workers, scaling status
- Current routing preference included in response
- Timeout: 3 seconds per target check (non-blocking, parallel)

## Acceptance Criteria

- [ ] With local GPU running: local status shows available with GPU details
- [ ] With local GPU stopped: local status shows unavailable
- [ ] With RunPod configured: remote status shows endpoint details
- [ ] With RunPod not configured: remote status shows unavailable (not an error)
- [ ] Response returns within 5 seconds even if one target is unreachable

## Dependencies

**Blocked by:** US-11.1 (compute routing engine), US-11.2 (RunPod serverless integration).

## References

- User stories: `user-stories/02-layer-2-platform-api.md` — Stage 11, US-11.4

frankbria:
## US-11.4: Compute status endpoint

Closes #125.

`GET /api/v1/compute/status` aggregates the health of both compute targets into one response so a client can check, **before** generating, whether the local GPU is busy and whether remote is available.

### What it does
- Probes the **local** ACE-Step server (`/v1/stats`) and the **remote** RunPod endpoint (`/health`) **in parallel**, each bounded by `compute_status_timeout` (default 3s) — so the aggregate response stays well under the issue's 5-second ceiling even when a target hangs.
- Reports each target's `available` flag plus best-effort detail (GPU name / VRAM / active jobs / loaded models for local; provider / endpoint_id / active+max workers / scaling status for remote) and the configured `routing_preference`.
- A down or unconfigured target degrades to `available=False` — never an error/500.
- **Auth-gated** like the rest of the API (only `/health` and `/auth` are public): worker counts and the RunPod endpoint id are operational detail for signed-in users.

### Plan adaptation (the issue's CodeRabbit plan was partially stale)
US-11.1/11.2 already landed, so this reuses existing infrastructure instead of building the parallel structures the plan proposed:
- Settings the plan wanted to add already exist as `local_url`, `runpod_api_key`, `runpod_endpoint_id`, `compute_preference` — only `compute_status_timeout` is new.
- Status logic lives in `services/compute_status.py` next to the existing `services/routing.py` (no new `compute/` package); response models live with it (the `models/` package is for Beanie documents).
- Added `RunPodClient.health_details()` (parsed `/health` body) since the existing `health()` returns only a bool.

### Acceptance criteria → evidence
| Criterion | Covered by |
|---|---|
| Local GPU running → available with GPU details | `test_available_with_details_when_server_responds`, `…_flat_gpu_string…` |
| Local GPU stopped → unavailable | `test_unavailable_on_connection_error` / `…_on_timeout` / `…_non_2xx…` |
| RunPod configured → remote shows endpoint details | `test_available_with_worker_detail_when_reachable` |
| RunPod not configured → unavailable (not an error) | `test_unavailable_and_no_http_when_not_configured` |
| Returns within 5s even if a target is unreachable | `test_responds_quickly_when_targets_unreachable`, `test_probes_run_concurrently` |

### Tests
20 new tests in `tests/test_compute_status.py` (respx + monkeypatch, no DB). Service coverage 97%, router 100%. Full non-integration suite: 1285 passed.

### Review
Cross-family review by `codex review` found one P2 (zero-valued VRAM dropped by an `or`-chain) — fixed in a follow-up commit with a regression test.

### Known limitations
- `max_workers` is not exposed by RunPod's `/health`; it is left `None` (would require RunPod's GraphQL API).
- Local GPU name / VRAM are populated only when the ACE-Step `/v1/stats` payload exposes them; otherwise `None` (graceful) — the field names are matched best-effort across common spellings.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -95,6 +95,11 @@ ACEMUSIC_API_COMPUTE_PREFERENCE=local_first
 # This is a routing probe URL (independent of the CLI's ACEMUSIC_BASE_URL).
 ACEMUSIC_API_LOCAL_URL=http://localhost:8001
 
+# Compute status endpoint (US-11.4). Per-target health-probe budget for
+# GET /api/v1/compute/status; local and remote are probed in parallel, each
+# bounded by this timeout, so the response stays under the 5s ceiling.
+ACEMUSIC_API_COMPUTE_STATUS_TIMEOUT=3
+
 # RunPod serverless remote routing (US-11.2). Enables remote GPU generation: when
 # BOTH the API key and endpoint ID are set, the routing engine probes the endpoint's
 # /health and may route generations to it. Leave either blank to run local-only
--- a/README.md
+++ b/README.md
@@ -111,6 +111,23 @@ default 5s) up to `RUNPOD_TIMEOUT` seconds (default 300), retries transient 5xx
 responses, and stores the returned clips exactly like a local job. `RUNPOD_BASE_URL`
 (default `https://api.runpod.ai/v2`) can point at a staging endpoint or proxy.
 
+| Method & path | Purpose |
+| --- | --- |
+| `GET /api/v1/compute/status` | Combined health of both compute targets, for a pre-flight check before generating (US-11.4) |
+
+The response carries `local`, `remote`, and `routing_preference`. `local`
+reports `available` plus best-effort `gpu_name`, `vram_total_mb`, `vram_used_mb`,
+`active_jobs`, and `loaded_models` (detail fields are `null` when the ACE-Step
+`/v1/stats` payload omits them). `remote` reports `available`, `provider`,
+`endpoint_id`, `active_workers`, `max_workers`, and `scaling_status`; an
+unconfigured RunPod returns `available: false` with `provider: null` (and makes
+no HTTP call), while a configured-but-unreachable endpoint keeps
+`provider: "runpod"` and `endpoint_id`. Both targets are probed in parallel,
+each bounded by `COMPUTE_STATUS_TIMEOUT` (default 3s), so the endpoint answers
+well within 5 seconds even when a target hangs — a down or unconfigured target is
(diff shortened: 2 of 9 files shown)


frankbria:
Thanks — addressed all three actionable items in 81f3723:

- **P2 (`or 0` / `active_workers` asymmetry)** — added a comment by `running` documenting that it is intentionally left `None`-vs-`0` distinguishable (0 active workers is a real reading), so no one re-adds `or 0` there by reflex.
- **P2 (timing test gap)** — added `test_responds_quickly_when_runpod_configured_and_hanging`: RunPod is configured and its `/health` blocks 1s, with `compute_status_timeout=0.2`; asserts the aggregate returns in <1s and `remote.available is False`. This exercises the bounded `asyncio.wait_for` path that the unconfigured case short-circuited past, so AC5 is now fully covered.
- **P3 (unbounded timeout)** — `compute_status_timeout` now `Field(default=3.0, gt=0, lt=5.0)`, so a config that would break the 5s ceiling fails at startup.

The two self-resolved items I left as-is by design: the broad `except Exception` in `get_remote_status` is the intended best-effort-probe contract, and a leaked non-`RequestError` from `health_details` is already caught by that same `except Exception` at the call site (so the `health()`-consistent narrow catch is safe here too).

Note: CodeRabbit hit the org review rate limit on this PR, so the cross-family review pass was `codex review` (which caught the original zero-VRAM P2, already fixed in a14-style follow-up).

PR #152, opened 2026-06-17
feat(api): RunPod Network Volume setup (US-11.5)

### LINKED ISSUE #126: US-11.5: RunPod Network Volume setup
## User Story

**As a** developer, **I want** a scripted setup process for the RunPod Network Volume with model weights, **so that** remote deployment can be reproduced or recovered quickly.

## Description

Provide a setup script and management commands for initializing the RunPod Network Volume with ACE-Step-1.5 model weights. The volume persists model weights so they only need to be downloaded once.

## Functional Requirements

- Setup script: `scripts/runpod-setup.py` that creates a Network Volume, spins up a temporary pod, downloads model weights, and stops the pod
- `GET /api/v1/compute/remote/volume` returns volume info (size, used, region)
- Script configurable for region and volume size
- Documentation of the one-time setup process in code comments
- Cost estimate displayed before creating resources

## Acceptance Criteria

- [ ] Setup script creates a Network Volume and downloads model weights
- [ ] Volume persists after the temporary pod is stopped
- [ ] Future serverless workers or pods can mount the volume and access weights
- [ ] Script is idempotent (re-running does not create duplicate volumes)

## Dependencies

**Blocked by:** US-11.3 (Docker image for ACE-Step — temporary pod uses it to download weights).

## References

- User stories: `user-stories/02-layer-2-platform-api.md` — Stage 11, US-11.5

frankbria:
## Summary
Implements #126 (US-11.5): one-time RunPod Network Volume setup for ACE-Step-1.5 model weights, plus an API endpoint to inspect the configured volume.

- **Settings**: added `runpod_network_volume_id` and `runpod_rest_base_url` to `ApiSettings` (env: `ACEMUSIC_API_RUNPOD_NETWORK_VOLUME_ID`, `ACEMUSIC_API_RUNPOD_REST_BASE_URL`); documented in `.env.example`.
- **Volume service** (`api/services/volume.py`): transport-agnostic `get_remote_volume()` querying RunPod's management REST API (`rest.runpod.io/v1/networkvolumes`), raising typed domain errors. Mirrors the `compute_status` service pattern.
- **Endpoint**: `GET /api/v1/compute/remote/volume` added to the existing auth-gated compute router; maps to **503** (not configured), **404** (volume absent), **502** (RunPod unreachable), **200** (volume metadata).
- **Setup script** (`scripts/runpod-setup.py`): `argparse` CLI that, idempotently, finds-or-creates the network volume, spins up a temporary GPU pod from the US-11.3 image to download weights, polls to completion, then stops + terminates the pod (volume persists). Shows a cost estimate and confirmation prompt; supports `--dry-run`.

## Adapted from the issue's CodeRabbit plan (which was stale)
The plan predated US-11.1–11.4 and assumed no compute router / RunPod settings existed. Adaptations:
- Reused the **existing** compute router (kept its `get_current_user` auth gate) instead of creating a new unauthenticated one.
- Only **added** `runpod_network_volume_id`/`runpod_rest_base_url`; the api-key/endpoint-id fields already exist.
- **No `runpod` SDK dependency**: both volume and pod operations use direct `httpx` REST, consistent with `runpod_client.py`, keeping the script unit-testable and the runtime lean.

## Acceptance Criteria
- [x] Setup script creates a Network Volume and downloads model weights (idempotent create + temp-pod download; `--dry-run` + mocked tests; live run is integration-only — see Known Limitations)
- [x] Volume persists after the temporary pod is stopped (script terminates the pod, never deletes the volume; asserted by `test_provisions_volume_then_terminates_pod_without_deleting_volume`)
- [x] Future serverless workers/pods can mount the volume (id printed for `ACEMUSIC_API_RUNPOD_NETWORK_VOLUME_ID`; `GET /compute/remote/volume` confirms it)
- [x] Script is idempotent (re-run reuses the existing volume by name, no duplicate POST; asserted by `test_reuses_existing_volume_without_creating`)

## Test Plan
- [x] Unit tests written first (TDD: RED→GREEN per step)
- [x] All tests passing (1318 passed, 0 failed)
- [x] Diff coverage ≥85% on new code (`volume.py` 98%, `compute.py` 95%)
- [x] Linting clean (`ruff`) and formatting clean (`black --check`)
- [x] Internal review (advisory) — superseded by cross-family pass
- [x] Cross-family review pass: **codex** (one P2 found and fixed — see below)

- **Live RunPod provisioning is not exercised in CI.** Creating a real volume + GPU pod requires a paid RunPod account and GPU; all tests mock RunPod's REST API via `respx`. The `--dry-run` mode prints the full plan + cost estimate without spending.
- **Weight-download command is best-effort and configurable.** The default `--download-cmd` warms the HF cache by starting the ACE-Step server until healthy; depending on how the published image fetches weights, operators may need to override `--download-cmd`. It now fails loudly (non-zero, no success marker) if the server never becomes healthy.
- **`used_gb` is always `None`.** RunPod's REST API reports allocated `size` but not live usage, so the field is present (forward-compatible) but unpopulated.
- **Completion is detected from the pod's terminal state**, not its container exit code (RunPod REST does not reliably expose it); the success message directs operators to confirm `ACE_STEP_WEIGHTS_READY` in the pod log.

## Implementation Notes
- codex review flagged a P2 silent-success bug: the original download loop exited 0 and printed the success marker even if the server never became healthy. Fixed in a dedicated commit — the loop now tracks readiness and exits non-zero without the marker on failure, and `run_setup` no longer over-claims completion.

Closes #126

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -115,6 +115,15 @@ ACEMUSIC_API_RUNPOD_POLL_INTERVAL=5
 # Override only to point at a staging endpoint or a proxy (default shown).
 # ACEMUSIC_API_RUNPOD_BASE_URL=https://api.runpod.ai/v2
 
+# RunPod Network Volume (US-11.5). The persisted volume holding ACE-Step model
+# weights, provisioned once by `scripts/runpod-setup.py` (which prints the id to
+# paste here). Workers mount it so weights download only once. Leave blank until
+# the setup script has run — GET /api/v1/compute/remote/volume returns 503 until
+# it is set. RUNPOD_REST_BASE_URL is RunPod's management REST API (volumes/pods),
+# distinct from the serverless RUNPOD_BASE_URL above; override only for staging.
+ACEMUSIC_API_RUNPOD_NETWORK_VOLUME_ID=
+# ACEMUSIC_API_RUNPOD_REST_BASE_URL=https://rest.runpod.io/v1
+
 # Async job processor (US-9.2). A background task in the API process polls MongoDB
 # for queued generation jobs and runs them. JOB_CONCURRENCY bounds how many run at
 # once; JOB_POLL_INTERVAL is the idle sleep (seconds) between polls when the queue
--- a/README.md
+++ b/README.md
@@ -102,18 +102,25 @@ default `local_first`), and `LOCAL_URL` (local ACE-Step probe URL, default
 `http://localhost:8001`).
 
 Remote generation runs on a RunPod serverless endpoint (US-11.2). Set
-`RUNPOD_API_KEY` and `RUNPOD_ENDPOINT_ID` (both required) to enable it; remote
+`ACEMUSIC_API_RUNPOD_API_KEY` and `ACEMUSIC_API_RUNPOD_ENDPOINT_ID` (both required) to enable it; remote
 routing is reported available only when both are set and the endpoint's `/health`
 answers, so an unconfigured or down endpoint degrades safely (`*_first` falls back
 to local, `*_only` returns 503) rather than crashing. A remote-routed job submits to
-the endpoint, polls with a cold-start-tolerant cadence (`RUNPOD_POLL_INTERVAL`,
-default 5s) up to `RUNPOD_TIMEOUT` seconds (default 300), retries transient 5xx
-responses, and stores the returned clips exactly like a local job. `RUNPOD_BASE_URL`
+the endpoint, polls with a cold-start-tolerant cadence (`ACEMUSIC_API_RUNPOD_POLL_INTERVAL`,
+default 5s) up to `ACEMUSIC_API_RUNPOD_TIMEOUT` seconds (default 300), retries transient 5xx
+responses, and stores the returned clips exactly like a local job. `ACEMUSIC_API_RUNPOD_BASE_URL`
 (default `https://api.runpod.ai/v2`) can point at a staging endpoint or proxy.
+
+The RunPod Network Volume — which holds model weights so each worker skips the
+download — is provisioned once by running `scripts/runpod-setup.py` (US-11.5).
+After the script runs, paste the printed volume id into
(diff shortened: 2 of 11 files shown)


frankbria:
Thanks for the thorough passes. Triage of the flagged items, addressed in f83d6be:

**Fixed**
- **`raise_for_status()`/`json()` inside `async with`** — moved both inside the context manager for unambiguous lifetime (it was already correct in practice, but this matches intent).
- **Misleading "not configured" message** — now names only the missing field(s), so an operator who set the key but not the volume id gets the right remediation step.
- **`available` always `True`** — documented in the model that it means "found on RunPod" (RunPod's REST API exposes no per-volume health state).
- **`--container-disk` flag** — added for heavier custom images.
- **No progress during poll** — `wait_for_pod_finish` now prints elapsed time each poll.
- **`:latest` default image** — `--image` help now recommends pinning for reproducibility.
- **Stale pricing constant** — annotated with `# Last verified: 2026-06-16`.
- **`import json` in a test** — hoisted to module top.
- **Bonus (surfaced while verifying #4):** the download pod is now pinned to the volume's data center (`dataCenterIds`), since a pod can only attach a network volume in the same DC.

**Verified as non-issues (no change needed)**
- **Async tests "may pass silently"** — `pyproject.toml` sets `asyncio_mode = "auto"`, so `pytest-asyncio` runs unmarked `async def` tests automatically (same convention as `test_compute_status.py`). Confirmed with `pytest -v`: all 9 `TestGetRemoteVolumeService` tests report **PASSED**, not skipped.
- **`dockerStartCmd` field name/type** — confirmed against RunPod's current create-pod REST schema (`rest.runpod.io/v1/pods`): the request body documents `dockerStartCmd` as a **list of strings** (`"dockerStartCmd": []`), exactly as used here. No `dockerArgs` string field in the v1 REST API.

PR #162, opened 2026-06-18
refactor(api): collapse create_*_job clones into shared factory (US-0.1)

### LINKED ISSUE #155: US-0.1: Collapse service-layer create_*_job clones into one factory
## US-0.1: Collapse service-layer `create_*_job` clones into one factory

**Source:** ponytail over-engineering audit (whole-repo, 2026-06-17).
**Phase priority:** Stage 0 (do **before** resuming stage 13/14 feature work) — 1 of 2 foundation refactors. Top audit finding.

> **Why stage 0, not 15:** every new job-type story clones this pattern. Stage 12 mastering already copied it; stages 13 (distribution) and 14 (DAW/playback) will copy it again. Consolidating first means the next stories build on the cleaned-up factory instead of multiplying the clone.

### Problem
The service layer adds **no logic** over the task layer — it is a pass-through. Four
`create_*_job` functions are ~50-line near-clones that differ only by the job-type
validation tuple:

- `src/acemusic/api/services/editing.py` → `create_edit_job` (`EDIT_JOB_TYPES`)
- `src/acemusic/api/services/extraction.py` → `create_extraction_job`
- `src/acemusic/api/services/iterative.py` → `create_iterative_job`
- `src/acemusic/api/services/mastering.py` → `create_mastering_job`
- `src/acemusic/api/services/generation.py` → `create_generation_job` (the original this set was copied from)

Each does the identical sequence: validate `job_type` against a module constant →
build `Job(... status=QUEUED ...)` → `await job.insert()` → `await dispatch_job(str(job.id))`
→ `except BaseException: await job.delete(); raise`.

### Cut
Introduce one shared factory and have each module delegate (keeping its job-type
constant tuple and public function name for caller compatibility):

```python
# api/services/jobs.py
async def create_job(*, user_id, workspace_id, job_type, valid_types, params) -> Job:
    if job_type not in valid_types:
        raise ValueError(f"Unknown job type: {job_type!r}")
    job = Job(user_id=user_id, workspace_id=workspace_id, job_type=job_type,
              status=JobStatus.QUEUED, input_params=params)
    await job.insert()
    try:
        await dispatch_job(str(job.id))
    except BaseException:
        await job.delete()
        raise
    return job
```

Thin per-domain wrappers (`create_edit_job`, …) just forward with their own
`valid_types`, or callers call `create_job` directly with the constant.

### Acceptance criteria
- [ ] Single `create_job` factory; the per-domain functions delegate (or are removed and callers updated).
- [ ] `BaseException` orphan-cleanup behaviour preserved (the `CancelledError` reason in the existing comments still holds).
- [ ] Existing service/router tests pass unchanged; no API contract change.
- [ ] ~150 lines net removed.

### Dependencies
None. Sequence **before US-0.2** only to serialize churn in the `api/` layer (different files, same review window).

frankbria:
## US-0.1: Collapse service-layer `create_*_job` clones into one factory

Closes #155. Stage-0 foundation refactor (1 of 2).

### What changed
New `src/acemusic/api/services/jobs.py` with a single `create_job(...)` that owns the shared sequence every job-submission service repeated:

```
validate job_type (optional) → build queued Job → insert → dispatch → on BaseException: delete orphan + re-raise
```

The five per-domain functions now **delegate** to it, keeping their public names and job-type constants so every caller (routers, `batch.py`, `export.py`) is untouched:

| Module | Wrapper | Delegation detail |
|--------|---------|-------------------|
| `editing` | `create_edit_job` | `valid_types=EDIT_JOB_TYPES` |
| `iterative` | `create_iterative_job` | `valid_types=ITERATIVE_JOB_TYPES` |
| `mastering` | `create_mastering_job` | no `valid_types` (single fixed type, as before) |
| `extraction` | `create_extraction_job` | keeps its own validation + idempotency lookup, delegates the persist |
| `generation` | `create_generation_job` | keeps workspace resolution, passes `compute_target` |

### Acceptance criteria
- [x] Single `create_job` factory; per-domain functions delegate.
- [x] `BaseException` orphan-cleanup behaviour preserved — verified by the three integration rollback tests (job count returns to baseline after a failed dispatch).
- [x] Existing service/router tests pass unchanged — full suite green in both default and `-m integration` modes.
- [x] Net lines removed from the duplicated clones (see Known Limitations on the "~150" estimate).

### Verification
- Default suite: **1361 passed**, 0 failed.
- Affected files under `-m integration`: **244 passed**, 0 failed.
- `ruff` + `black --check` clean on all changed files.
- Cross-family review: `codex review` (its P2 finding is addressed — see below).

### Known limitations
- **Net line count is ~ -62 from the wrappers, not the issue's aspirational "~150".** Once the new factory file (with its docstring) is counted, the repo-wide raw delta is roughly neutral. The real win is the dedup: four copies of the insert/dispatch/cleanup sequence collapsed to one source of truth — future job types (stages 13/14) build on it instead of re-cloning.
- **Three rollback tests had their monkeypatch target repointed** from `<service_module>.dispatch_job` to `acemusic.api.services.jobs.dispatch_job`, because `dispatch_job` is now imported only in the shared factory. The seam genuinely moved; the assertions (orphan cleanup, no refund leak) are unchanged and still fully exercised. This was the codex P2 finding — caught because these tests are `@pytest.mark.integration` and skipped in the default run.

### DIFF
--- a/docs/demos/us-0.1-job-factory.md
+++ b/docs/demos/us-0.1-job-factory.md
@@ -1,0 +1,72 @@
+# US-0.1: Shared create_job factory — acceptance demo
+
+*2026-06-18T00:13:21Z*
+
+Refactor goal: collapse five ~50-line `create_*_job` clones into one shared factory (`api/services/jobs.create_job`) while preserving every public function name, the BaseException orphan-cleanup, and all existing tests. Each criterion below is proven with live outcome evidence.
+
+### Criterion 1 — Single factory; per-domain functions delegate
+
+```bash
+echo "Factory definition:"; grep -n "async def create_job" src/acemusic/api/services/jobs.py; echo; echo "Each wrapper delegates to create_job:"; for f in editing extraction iterative mastering generation; do printf "  %-12s -> " "$f"; grep -c "create_job(" src/acemusic/api/services/$f.py; done
+```
+
+```output
+Factory definition:
+18:async def create_job(
+
+Each wrapper delegates to create_job:
+  editing      -> 1
+  extraction   -> 1
+  iterative    -> 1
+  mastering    -> 1
+  generation   -> 1
+```
+
+Outcome: one factory, and all five wrappers reference `create_job` — no module still hand-rolls the insert/dispatch/cleanup.
+
+### Criterion 2 — BaseException orphan-cleanup preserved
+
+The three integration rollback tests make dispatch raise *after* the job is inserted, then assert the job count returns to baseline (no orphaned QUEUED doc the processor would run for free).
+
+```bash
+ulimit -n 64000; uv run pytest tests/test_mastering_service.py tests/test_clips_iterative_tasks.py tests/test_credits_api.py -m integration -k "dispatch" -v 2>&1 | grep -E "PASSED|FAILED|passed|failed"
+```
+
+```output
+tests/test_mastering_service.py::TestCreateMasteringJob::test_dispatch_failure_rolls_back_job PASSED [ 33%]
+tests/test_clips_iterative_tasks.py::TestService::test_dispatch_failure_rolls_back_job PASSED [ 66%]
+tests/test_credits_api.py::TestRefundOnJobCreationFailure::test_no_orphaned_job_when_dispatch_fails_after_insert PASSED [100%]
+================= 3 passed, 55 deselected, 1 warning in 4.58s ==================
+```
+
+### Criterion 3 — Validation behaviour kept (delegated via valid_types)
+
+```bash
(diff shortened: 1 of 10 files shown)


frankbria:
Thanks for the review. Applied the two polish items: the validation error now reads `Unknown job type {x!r}; expected one of {valid_types}`, and `params` is now `dict[str, Any]`. Declined #4 (dedicated mocked unit test for `jobs.py`): the factory's guard, happy path, and BaseException orphan-cleanup are all exercised by real-service integration tests (the three `test_dispatch_failure_rolls_back_job` + `test_unknown_job_type_raises`), and this project's policy is no mocking — integration/E2E hit real Mongo. #3 (residual `Job` imports) is intentional — used by the `-> Job` return annotation; ruff is clean.

PR #163, opened 2026-06-18
refactor(api): shared task-layer helpers + unify JobProcessingError (US-0.2)

### LINKED ISSUE #156: US-0.2: Extract shared API task-layer helpers + unify JobProcessingError
## US-0.2: Extract shared API task-layer helpers + unify `JobProcessingError`

**Source:** ponytail over-engineering audit (whole-repo, 2026-06-17).
**Phase priority:** Stage 0 (do **before** resuming stage 13/14 feature work) — 2 of 2 foundation refactors.

> **Why stage 0, not 15:** the task modules that stages 13/14 will extend copy these helpers and exception classes verbatim. Consolidate first so new job types reuse the shared helpers instead of cloning them.

### Problem
The task modules copy-paste the same clip lifecycle helpers and each redefine their
own near-identical exception class.

**Duplicated helpers** (`load source clip → download → store derived clip → rollback`):
- `_load_source_clip(job)` — verbatim in `editing.py:40`, `extraction.py:57`, `mastering.py:84`, `export.py:45`, `iterative.py:97`
- `_download_source(...)` — 4 copies with trivial signature differences (`editing.py:107`, `extraction.py:75`, `mastering.py:101`, `export.py:63`)
- `_store_*_clip` + `_rollback_clips` — identical download→upload→insert→rollback loop in `editing.py`, `iterative.py`, `mastering.py`, `extraction.py`

**Duplicated exceptions:** six per-module classes that are all empty `Exception` subclasses —
`EditProcessingError` (`editing.py:36`), `ExtractionProcessingError` (`extraction.py:53`),
`MasteringProcessingError` (`mastering.py:48`), `ExportProcessingError` (`export.py:41`),
`IterativeProcessingError` (`iterative.py:88`). **`JobProcessingError` already exists** at
`api/tasks/processor.py:72`.

### Cut
- New `api/tasks/common.py` (or extend `api/tasks/__init__.py`) with shared
  `load_source_clip`, `download_source`, `store_derived_clip`, `rollback_clips`.
  Parameterize the small differences (dest `Path` vs return `bytes` vs workspace wrapper)
  rather than keeping four copies.
- Replace the six `*ProcessingError` classes with the existing `JobProcessingError`.

### Acceptance criteria
- [ ] One implementation each of source-clip load / download / store / rollback.
- [ ] All raises use `JobProcessingError`; the five subclasses deleted.
- [ ] Rollback-on-failure semantics unchanged (orphan-clip cleanup still runs).
- [ ] Task tests pass unchanged.
- [ ] ~200 lines net removed.

### Dependencies
Touches the same five `api/tasks/*.py` files that US-0.1's downstream does. Do **after US-0.1** to avoid concurrent edits to the editing/extraction/iterative/mastering modules.

frankbria:
## US-0.2: Extract shared API task-layer helpers + unify `JobProcessingError`

Closes #156.

Stage-0 foundation refactor (from the ponytail over-engineering audit). The job task modules copy-pasted the same clip lifecycle helpers and each redefined a near-identical empty `Exception` subclass. This consolidates them so future job types reuse the shared helpers instead of cloning.

### What changed
- **New `api/tasks/common.py`** with one implementation each of:
  - `load_clip(clip_id)` / `load_source_clip(job)` — resolve a source clip, fail the job cleanly if it vanished
  - `download_clip(storage, clip) -> bytes` — download audio, fail the job if the object is missing
  - `store_clip(storage, clip, data)` — upload + insert, rolling the object back if the insert fails
  - `rollback_clips(storage, clip_ids)` — best-effort batch cleanup of already-stored children
- **`JobProcessingError` moved into `common.py`** (was in `processor.py`) and re-imported there, so the job modules import it without a cycle. The five per-module subclasses (`EditProcessingError`, `ExtractionProcessingError`, `MasteringProcessingError`, `ExportProcessingError`, `IterativeProcessingError`) are **deleted** — all raises now use `JobProcessingError`.
- `editing`, `export`, `extraction`, `mastering`, `iterative` refactored to call the shared helpers; each module still builds its own (module-specific) `Clip`.

### Acceptance criteria
- [x] One implementation each of source-clip load / download / store / rollback.
- [x] All raises use `JobProcessingError`; the five subclasses deleted.
- [x] Rollback-on-failure semantics unchanged (`store_clip` rolls back a just-uploaded object on insert failure; `rollback_clips` cleans up earlier children on a mid-batch failure).
- [x] Task tests pass unchanged (26 unit + 196 integration green).
- [x] ~140 lines net removed from production code.

### Testing
- `pytest` task-layer suites (editing/extraction/mastering/export/iterative/full-song): **26 passed**
- `pytest -m integration` same suites: **196 passed**
- `ruff check` + `black --check`: clean

### Known limitations
- `processor._rollback_clips` (the `generate`-path rollback over `(Clip, path)` tuples) is intentionally left as-is — different shape/semantics, out of scope for this issue.
- `editing._download_source` keeps its thin `_EditWorkspace`-aware wrapper (the "workspace wrapper" small-difference the issue called out); it now delegates to `download_clip`.

### DIFF
--- a/src/acemusic/api/tasks/common.py
+++ b/src/acemusic/api/tasks/common.py
@@ -1,0 +1,95 @@
+"""Shared task-layer helpers for the job handlers (US-0.2).
+
+Every job-type module (editing, extraction, mastering, export, iterative) runs
+the same clip lifecycle: resolve the source clip, download its audio, store a
+derived clip with rollback-on-failure. These helpers are that lifecycle, factored
+out of the per-module copies so a new job type reuses them instead of cloning.
+
+:class:`JobProcessingError` is the single failure type every handler raises; the
+:class:`~acemusic.api.tasks.processor.JobProcessor` records its message as the
+job's failure (it lived in ``processor`` before US-0.2 and is re-exported there).
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+
+from beanie import PydanticObjectId
+
+from acemusic.storage import StorageBackend
+
+from ..models import Clip, Job
+
+logger = logging.getLogger(__name__)
+
+
+class JobProcessingError(Exception):
+    """A job could not be processed (missing source, audio/ML/service failure)."""
+
+
+async def load_clip(clip_id: object) -> Clip:
+    """Resolve a source clip by id, or fail the job with a clear error.
+
+    The clip may legitimately vanish between enqueue and processing (the owner
+    can DELETE it while the job is queued), so a miss is a job failure, not a
+    crash.
+    """
+    try:
+        oid = PydanticObjectId(clip_id)
+    except Exception as exc:
+        raise JobProcessingError(f"Job has an invalid source clip id: {clip_id!r}") from exc
+    clip = await Clip.get(oid)
+    if clip is None:
+        raise JobProcessingError(f"Source clip {clip_id} no longer exists")
+    return clip
+
+
+async def load_source_clip(job: Job) -> Clip:
+    """Resolve the clip referenced by ``job.input_params['clip_id']``."""
+    return await load_clip((job.input_params or {}).get("clip_id"))
+
+
+async def download_clip(storage: StorageBackend, clip: Clip) -> bytes:
+    """Download a clip's audio bytes, failing the job if the object is missing."""
+    try:
+        return await asyncio.to_thread(storage.download, clip.file_path)
+    except FileNotFoundError as exc:
+        raise JobProcessingError(f"Source clip {clip.id} audio object {clip.file_path!r} is missing") from exc
+
+
+async def store_clip(storage: StorageBackend, clip: Clip, data: bytes) -> Clip:
+    """Upload ``data`` to ``clip.file_path`` and insert ``clip``.
+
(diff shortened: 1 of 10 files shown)


frankbria:
Thanks for the careful review.

**Fixed (e799221):** `rollback_clips` now wraps `await clip.delete()` in its own best-effort `try/except`, so a transient Mongo error during rollback can no longer mask the original job-failure exception that callers re-raise. Verified this gap pre-existed identically in the per-module `_rollback_clips`/`_rollback_children` this PR consolidated — so it was carried over faithfully and is now fixed once at the single chokepoint. This also makes the code match the "best-effort" wording already in the docstring.

**Minor observations — no change, by design:**
- `PydanticObjectId(clip_id)` conversion: callers only ever pass ids minted by `store_clip`, so a malformed id is unreachable; left as-is to avoid speculative guarding.
- `_store_child_clip` ignoring `store_clip`'s return and returning the pre-built `clip_id`: intentional — the iterative path already holds the id; both are equivalent.
- `_store_master_clip` docstring trim: intentional, the rollback contract now lives in `store_clip` where the behavior is.

PR #168, opened 2026-06-20
feat(runpod): remote audio delivery via S3 presigned URLs (US-11.x)

### LINKED ISSUE #150: US-11.x: Remote audio delivery for RunPod serverless worker (S3/presigned URLs)
## Background

The US-11.3 RunPod serverless handler (`docker/handler.py`, PR #149) returns ACE-Step's
**worker-local** audio URLs (`http://localhost:8001/v1/audio?path=...`). When the worker
runs remotely on RunPod, the platform's `RunPodClient.download_audio` (US-11.2) cannot
reach those URLs — `localhost` resolves on the platform host, not the worker — so a
completed remote job would fail during clip storage.

This was a deliberate scope decision for US-11.3 (build the image); remote audio delivery
is follow-up infrastructure.

## Task

Make generated audio reachable by the platform from a remote RunPod worker. Options:
- Handler uploads clips to S3 (the `s3` extra already exists) and returns **presigned URLs**.
- Or return audio **bytes** inline (base64) and extend `RunPodClient`/`_extract_audio_urls`
  + the download path to accept them.

## Acceptance criteria
- [ ] A remote (RunPod) generation returns audio the platform can actually download & store
- [ ] No worker-local `localhost` URLs escape to the platform
- [ ] Covered by tests (handler output shape + platform consumption)

## References
- PR #149 (US-11.3), `docker/handler.py`
- `src/acemusic/runpod_client.py` (`_extract_audio_urls`, `download_audio`)
- `model-deployment.md` §3.4

frankbria:
## Summary

Makes generated audio reachable by the platform from a **remote** RunPod serverless worker. Previously the handler returned worker-local URLs (`http://localhost:8001/v1/audio?...`) that a remote platform host cannot reach, so any completed remote job would fail during clip storage (US-11.3 scope deferral, issue #150).

### Handler (`docker/handler.py`, `docker/Dockerfile`)
- When `ACEMUSIC_S3_BUCKET` is configured, each generated clip is **downloaded on-worker** from the ACE-Step server and **re-uploaded to S3**, and a **presigned GET URL** is returned in its place (key `{<ACEMUSIC_S3_PREFIX>/}runpod/<task_id>/<i><ext>`).
- Reuses the platform's existing `ACEMUSIC_S3_*` env vars (`_BUCKET`, `_PREFIX`, `_REGION`, `_ENDPOINT_URL`, `_URL_EXPIRY`). `boto3` added to the worker image.
- With **no bucket** configured, worker-local URLs pass through unchanged (local / single-host use). On an actual serverless worker (`RUNPOD_ENDPOINT_ID` present) a missing bucket **fails loudly** instead of returning unreachable URLs.
- The ACE-Step bearer token is attached **only** for the worker-local origin (compared on parsed scheme/host/port, not a string prefix), so it is never disclosed to an external result host.

### Platform client (`src/acemusic/runpod_client.py`)
- `_extract_audio_urls` now drops any `localhost` / `127.0.0.1` / `[::1]` URLs (with a warning) as a safety net, so stale worker images can never hand the platform URLs it cannot fetch.

## Acceptance criteria
- [x] A remote (RunPod) generation returns audio the platform can actually download & store — handler re-hosts clips to S3 and returns presigned URLs
- [x] No worker-local `localhost` URLs escape to the platform — handler returns presigned URLs; client filters localhost defensively
- [x] Covered by tests — handler S3 path / no-S3 path / serverless-misconfig / token-scoping, and client localhost-filtering / presigned passthrough

## Testing
- `tests/test_docker_handler.py` and `tests/test_runpod_client.py` — 64 passing
- Full suite: 1445 passed, 614 integration deselected
- `black --check`, `ruff`, and a cross-family `codex review` pass clean

## Known limitations
- **S3 provisioning is ops config.** The RunPod serverless deployment must set the `ACEMUSIC_S3_*` env vars; this PR adds the code path, not the credentials (consistent with the issue scope).
- **Staging-object lifecycle.** Each remote generation leaves a staging object under `runpod/<task_id>`; the platform stores its own canonical copy on download. In-handler deletion would need cross-component coordination, so cleanup is left to a **bucket lifecycle policy** (presigned-URL expiry defaults to 1h via `ACEMUSIC_S3_URL_EXPIRY`).
- Base64-inline audio delivery was considered and rejected (no existing patterns, poor fit for large files).

Closes #150

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -27,9 +27,10 @@ RUN git clone https://github.com/frankbria/ACE-Step-1.5.git /app/ACE-Step-1.5
 WORKDIR /app/ACE-Step-1.5
 RUN uv venv && uv sync
 
-# Serverless extras: the RunPod SDK plus httpx for the handler's API proxy (httpx is
-# the HTTP client used throughout the platform; the handler reuses it here).
-RUN pip install --no-cache-dir runpod httpx
+# Serverless extras: the RunPod SDK, httpx for the handler's API proxy (httpx is the
+# HTTP client used throughout the platform), and boto3 so the handler can upload
+# generated clips to S3 and return presigned URLs the platform can fetch (US-11.x).
+RUN pip install --no-cache-dir runpod httpx boto3
 
 # Serverless handler (bridges RunPod events to the local ACE-Step API).
 COPY handler.py /app/handler.py
--- a/docker/handler.py
+++ b/docker/handler.py
@@ -29,22 +29,38 @@ Run modes:
   worker loop. ``runpod`` is imported lazily there so this module stays importable
   for unit tests without the runpod SDK installed and never starts a server on import.
 
-Known limitation: the audio URLs returned here point at the worker-local ACE-Step
-server (``http://localhost:8001/v1/audio?...``). Delivering generated audio across
-the internet (S3 upload / presigned URLs) is follow-up infrastructure tracked
-separately; the platform's extractor already tolerates the URL shape emitted here.
+Remote audio delivery (US-11.x): ACE-Step writes clips to the worker-local server
+(``http://localhost:8001/v1/audio?...``), which a remote platform host cannot reach.
+When S3 is configured (``ACEMUSIC_S3_BUCKET`` plus the ``ACEMUSIC_S3_*`` settings the
+platform already uses), the handler downloads each clip on-worker and re-uploads it to
+S3, returning presigned GET URLs the platform can fetch. With no bucket configured it
+returns the worker-local URLs unchanged (local / single-host use, where ``localhost``
+is reachable).
 """
 
 from __future__ import annotations
 
 import json
+import mimetypes
 import os
 import subprocess
 import time
+import urllib.parse
 
 import httpx
 
+try:  # boto3 ships in the worker image; optional so this module imports without it.
+    import boto3
+except ImportError:  # pragma: no cover - exercised only where boto3 is absent
+    boto3 = None
+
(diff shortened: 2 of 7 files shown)


frankbria:
## Review triage (claude[bot] review)

Thanks — addressed below. Two of the three "Issues to Address" were already resolved by commits that landed after that review ran:

1. **`.cache/` files in the diff** — ✅ already fixed. Those were stray untracked files swept in by a `git add -A`; the branch was cleaned and now contains exactly 7 files (handler, Dockerfile, runpod_client, two test files, two demo files). No `.cache/` in the PR.
2. **GitGuardian "Basic Auth String" false positive** — ✅ already fixed. The flagged string was the security-test URL `http://localhost:<port>@attacker.example/...`, which GitGuardian read as `user:pass@host`. Rather than allowlist it, the fixtures now use a userinfo form without a password and the strict `:port@` regression is built from `API_BASE_URL` at runtime (no literal credential). GitGuardian now passes.
3. **Silent fallback on invalid `ACEMUSIC_S3_URL_EXPIRY`** — ✅ done in 45c118f. `_s3_url_expiry()` now logs a warning for invalid/non-positive values before defaulting to 3600s, with parametrized tests.

Minor: documented boto3's AWS credential chain (`AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / instance role) in the handler docstring alongside the `ACEMUSIC_S3_*` vars (45c118f). The broad `except Exception` and fixed 120s download timeout are intentional (noted in code / sized for large clips within the serverless ceiling).

PR #240, opened 2026-06-27
feat(web): Workspace and Clip Library Panel (US-16.5)

### LINKED ISSUE #191: US-16.5: Workspace and Clip Library Panel
## User Story

**As a** musician, **I want** a workspace panel showing all my clips with search, filters, and sorting, **so that** I can manage my generated content alongside creation.

## Description

Build the right-side workspace panel on the Create page with clip cards in list view, workspace breadcrumb, search, filter controls, sort dropdown, and pagination. This panel is the musician's clip library during the creation workflow.

## Functional Requirements

- Workspace breadcrumb showing current workspace path (clickable to navigate)
- Search input filtering clips by title or metadata
- Filters button with active filter count badge; filter panel includes: liked, public, uploads
- Sort dropdown: Newest, Oldest
- Pagination controls (previous/next + page number)
- Clip cards rendered in list view (see US-16.6)
- Panel scrolls independently of main content

## Acceptance Criteria

- [ ] Workspace breadcrumb displays current workspace and navigates on click
- [ ] Search filters clips in real time by title or metadata
- [ ] Filter toggles (Liked, Public, Uploads) narrow the clip list
- [ ] Sort dropdown reorders clips
- [ ] Pagination loads additional pages of clips
- [ ] Panel scrolls independently from the creation form

## Dependencies

**Blocked by:** US-16.4 (prior story in this stage). Part of Stage 16: Creation Pages, which is blocked by Stage 15.

## References

- User stories: `user-stories/03-layer-3-web-ui.md` — Stage 16, US-16.5

frankbria:
## US-16.5: Workspace and Clip Library Panel

Closes #191.

Adds the right-side clip library to the **Create** page: a workspace breadcrumb,
debounced search, Liked/Public/Uploads filters, Newest/Oldest sort, pagination,
and a minimal clip card/list — in a two-column layout that scrolls independently
of the creation form.

> [!NOTE]
> The CodeRabbit issue plan was written when `web/` was empty and uses
> `web/src/...` paths and an `items` response key. It's **stale**. This PR is
> adapted to the real app: flat structure (`app/`, `components/`, `lib/`,
> `hooks/`), PascalCase components, co-located vitest tests, and the actual
> backend contract.

### How it works
- **Search / sort / pagination** are server-driven via `GET /api/v1/clips`
  (params `search`, `sort`, `page`, `per_page`, `workspace_id`), reached through
  new same-origin BFF proxy routes (`/api/clips`, `/api/workspaces`) that forward
  the Bearer token — the established pattern (`app/api/users/me/route.ts`).
- **Breadcrumb** uses `GET /api/v1/workspaces` (default workspace) and navigates
  to `/studio` on click.
- **Filters** are applied **client-side** (the backend has no `liked/public/
  uploads` params): Liked reuses the player's localStorage `likedIds`, Public
  uses `is_public`, Uploads uses `generation_mode === "upload"`. This was an
  explicit, approved interim choice.

### Acceptance criteria
- [x] Breadcrumb displays current workspace + navigates on click
- [x] Search filters in real time (debounced, server `search`)
- [x] Filter toggles (Liked / Public / Uploads) narrow the list
- [x] Sort dropdown reorders clips (server `sort`)
- [x] Pagination loads pages (server `page`)
- [x] Panel scrolls independently of the creation form

### Tests
27 new files; `npm test` → **238 passed**. Typecheck, lint, and `next build`
all clean locally (web is not in CI).

### Known limitations
- **Client-side filters apply to the fetched page only.** Because the backend
  has no `liked/public/uploads` query params, filtering happens after server
  pagination — matches on other pages aren't pulled in, and `totalPages` reflects
  the unfiltered set (raised as P2 by `codex review`; the empty-state wording is
  scoped to "this page" to avoid implying the whole library is empty). Promoting
  these to real server-side filters is a backend follow-up.
- **`ClipCard` is intentionally minimal** — inline edit, play/like/menu actions,
  and full interactivity are US-16.6.
- **No "liked" backend exists**; the Liked filter is only as complete as the
  player's local liked set.

## Summary by CodeRabbit

* **New Features**
  * Added a workspace clip library panel with search, filters, sorting, pagination, and empty/loading states.
  * Workspace navigation now includes a breadcrumb, plus a clearer clip card view with duration and tags.
  * Introduced backend-backed workspace and clip loading for authenticated users.

* **Bug Fixes**
  * Improved handling of missing clip titles by showing a friendly fallback.
  * Added consistent unauthorized responses when workspace or clip data is accessed without authentication.

* **Tests & Documentation**
  * Expanded automated coverage for search, filters, pagination, and data loading.
  * Added demo documentation with verified live-test results.

### DIFF
--- a/docs/demos/us-16.5-workspace-panel.md
+++ b/docs/demos/us-16.5-workspace-panel.md
@@ -1,0 +1,32 @@
+# US-16.5 — Workspace and Clip Library Panel demo
+
+Verified live against `next dev` (localhost:3100) with `agent-browser`. The panel
+is data-driven (clips + workspaces) and lives on the auth-gated `/create` route,
+so it was mounted on a throwaway harness page (`app/demo-workspace`, since
+deleted) under a token'd `AuthContext` and the real `PlayerProvider`, with
+`window.fetch` stubbed to serve a seeded workspace ("My Beats") and 25 clips so
+the real `WorkspacePanel` + `useClips`/`useWorkspaces` hooks + client-side
+filtering all ran against live data. Each acceptance criterion is mapped to
+observed outcome evidence (not just "rendered").
+
+Seed shape: 25 clips, newest-first `c1…c25`. `c1` (Aurora) public/generate,
+`c2` (Bassline) private/upload **and** seeded into the player's liked set,
+`c3` (Crescendo) public/upload, `c4…c25` private/generate.
+
+| # | Acceptance criterion | Evidence |
+|---|----------------------|----------|
+| 1 | Workspace breadcrumb displays current workspace and navigates on click | Breadcrumb read live as `"Workspaces \| My Beats"` (the default workspace from `GET /api/workspaces`). Clicking the `Workspaces` root segment fired the wired `onNavigate` — captured console line `[demo] breadcrumb navigate clicked` (on the real `/create` page this is `router.push("/studio")`). |
+| 2 | Search filters clips in real time by title or metadata | Filling the search box with `Aurora` (debounced) narrowed the list from 20 cards to exactly `1 card: Aurora`. The term round-trips as the server `search` query param via `useClips`. |
+| 3 | Filter toggles (Liked, Public, Uploads) narrow the clip list | Opening **Filters** and toggling **Public** → `2 cards: Aurora, Crescendo` and the button badge showed `1`. **Liked** alone → `1 card: Bassline` (the seeded liked id `c2`, read from the player's `likedIds`). **Uploads** alone → `2 cards: Bassline, Crescendo` (`generation_mode === "upload"`). Clearing all → back to 20. |
+| 4 | Sort dropdown reorders clips | Default **Newest** showed first three `Aurora, Bassline, Crescendo`; selecting **Oldest** reordered the first three to `Track 25, Track 24, Track 23` (server `sort=oldest`). |
+| 5 | Pagination loads additional pages of clips | Footer read `Page 1 of 2` (25 clips / 20 per page) with **Previous** disabled. Clicking **Next** → `Page 2 of 2`, `5 cards` (`Track 21…25`), **Next** now disabled, **Previous** enabled. Clicking **Previous** returned to `Page 1 of 2`. |
+| 6 | Panel scrolls independently from the creation form | The panel's clip-list container is its own scroll context: live DOM read `scrollable=true scrollH=1632 clientH=532` (`overflow-y-auto` with a flex-constrained height), independent of the left form column. |
+
+Functional extras observed: with a filter active on a page whose clips don't
+match, the empty state reads **"No clips on this page match your filters."** —
+deliberately scoped to the page, because the Liked/Public/Uploads filters run
+client-side over the fetched page (the backend has no params for them yet; see
+Known Limitations in PR #240). No feature-related console errors (`0 captured`).
+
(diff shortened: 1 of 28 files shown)


frankbria:
Thanks — all four correctness findings addressed in e679bc5:

1. **Spurious unscoped fetch** — `useClips` now takes an `enabled` flag; `WorkspacePanel` passes `enabled: !workspacesLoading`, so clips aren't fetched until the workspace id resolves (no unscoped flash / double fetch).
2. **Filter change didn't reset page** — added `onFiltersChange` that calls `setPage(1)`, mirroring search/sort.
3. **Stale `debouncedSearch` in empty-state copy** — `narrowed` now reads the live `search`.
4. **`error` blocked the retry skeleton** — replaced the boolean with a per-query `errorQuery`; a changed query makes the prior error stale, so `loading` flips back on for the retry (and a failed first load still drops out of loading to avoid an endless skeleton — covered by a new test). Note: a synchronous `setError(false)` at the top of the effect would trip the repo's `react-hooks/set-state-in-effect` rule, hence the per-query approach.

Two new tests cover the `enabled` gate and the post-error retry skeleton. Full local gate green: typecheck, lint, `npm test` (240), build.

**Minor note (extract `useDebouncedValue`)** — leaving it inline for now (YAGNI); will extract to `hooks/use-debounced-value.ts` when a third copy appears, as suggested.

frankbria:
Follow-up review addressed in 094a0e0:

**Finding 1 — unscoped fetch on zero/error workspace** ✅ Fixed. The `enabled` gate now keys on `defaultWorkspace !== null`, so the workspace-error and zero-workspace cases no longer fall through to an unscoped fetch.

**Finding 2 — Next enabled at `totalPages === 0`** ✅ Fixed. `PaginationControls` clamps once (`const clampedTotalPages = Math.max(totalPages, 1)`) and uses it for both the label and the `disabled` check. Test added.

**Finding 3 — re-render on every playback tick** — Not applied. The diagnosis is correct (`state.currentTime` updates ~4 Hz and the context value is memoized on `[state]`), **but the suggested fix doesn't work**: React context re-renders *every* consumer when the provider value changes, regardless of which fields you destructure — `const { state: { likedIds } } = usePlayer()` still re-renders at 4 Hz. A real fix needs a context selector or splitting the player context, which is a `player-context` architecture change out of scope for this UI story. The per-tick work here is negligible (a `useMemo`-guarded filter + light JSX, only during active playback), so I've left it rather than ship an ineffective change. Happy to open a follow-up to add a selector to `usePlayer` if the broader re-render cost matters.

CodeRabbit inline:
- **WorkspacePanel error handling** ✅ The panel now reads `error` from `useClips` and shows "Couldn't load clips. Please try again." instead of the empty-library copy. Test added.
- **use-clips stale error** — handled in the prior commit via per-query `errorQuery` (a changed query supersedes the stale error; a synchronous `setError(false)` would trip `react-hooks/set-state-in-effect`).
- **use-workspaces error reset** — `useWorkspaces` fetches once per session (no param-driven refetch), so there's no in-session retry to suppress; leaving as-is.
- **buildClipQuery drops `0`** — `page`/`per_page` are `≥ 1` on the backend and never 0 from this panel; the falsy check is intentional for blank `search`/`workspace_id`. No behavior change for any real caller.

Local gate green: typecheck, lint, `npm test` (242), build.

PR #249, opened 2026-07-05
feat(web): Like, Dislike, Share, and Publish inline actions (US-17.6)

### LINKED ISSUE #200: US-17.6: Like, Dislike, Share, and Publish Actions
## User Story

**As a** musician, **I want** inline like, dislike, share, and publish controls on clips and song detail pages, **so that** I can engage with content and control visibility without extra steps.

## Description

Implement inline action buttons that appear on clip cards and song detail pages. These are the primary engagement and visibility controls.

## Functional Requirements

- Like button (heart icon): toggles liked state, updates count, persists to API
- Dislike button: toggles dislike state, affects recommendations
- Share button: opens share modal with copy-link, and social sharing options
- Publish button: toggles between private/unlisted/public visibility states
- Publishing requires a title and at least one style tag; prompt if missing
- Optimistic UI updates with rollback on API failure

## Acceptance Criteria

- [ ] Like/dislike toggle states correctly and persist after page reload
- [ ] Share modal provides a copyable link
- [ ] Publish toggle changes visibility state with API confirmation
- [ ] Publishing without a title or style tag shows a prompt to add them
- [ ] UI updates optimistically and rolls back on error

## Dependencies

**Blocked by:** US-17.5 (prior story in this stage). Part of Stage 17: Song Detail & Editing UI, which is blocked by Stage 16.

## References

- User stories: `user-stories/03-layer-3-web-ui.md` — Stage 17, US-17.6

frankbria:
## US-17.6 — Like, Dislike, Share, and Publish Actions

Closes #200.

Wires the inline clip action controls — which shipped as **local stubs** in US-16.6/17.1 and were explicitly marked "US-17.6 lands the routes" — to real behavior on both clip cards and the song detail page.

### What changed

**Publish (persisted, guarded, optimistic)**
- New `PATCH /api/clips/{id}` BFF route forwards `{is_public}` to the backend.
- Backend: `ClipUpdate` gains `is_public`; new `update_clip_fields` service applies title and/or visibility with a **publish guard** — going public requires a title **and** ≥1 style tag, else `422`. Unpublishing is never guarded.
- Client mirrors the guard (prompts *before* the request via `PublishGuardPrompt`) and is optimistic with **rollback** on failure; a server `422` also rolls back and prompts.

**Like / Dislike (persist across reload)**
- Dislike now lives in the player store + `localStorage` (mirrors like; the two are mutually exclusive per clip), so both survive a page reload.

**Share (client-only)**
- `ShareModal` with a copyable `{origin}/song/{id}` link (copy-to-clipboard + inline confirmation) and X/Facebook share intents. No backend call (share URL is derived from the clip id).

### Scope decision (user-approved: web-first / minimal)

This is a `web`/`layer-3` story, and every prior Stage-17 story was web-only. The stale CodeRabbit plan (written before the web app existed) proposed a full backend reaction subsystem; instead this ships the **one small backend change** needed for truthful publish persistence and keeps everything else in the web layer.

**Deferred to future layer-2 stories** (not required by the acceptance criteria):
- Server-side `ClipReaction` collection ("dislike affects recommendations")
- Three-state `unlisted` visibility (the inline control is binary public/private, matching the existing UI)

### Acceptance criteria
- [x] Like/dislike toggle correctly and persist after page reload
- [x] Share modal provides a copyable link
- [x] Publish toggle changes visibility with API confirmation
- [x] Publishing without a title or style tag shows a prompt to add them
- [x] UI updates optimistically and rolls back on error

### Testing
- **Backend** (real MongoDB): `TestPublishClip` (publish persists, 422 guard for missing title/tags, unpublish unguarded, rename+publish) — 6 passed; `TestUpdateClip`/`TestAuthGate` unaffected.
- **Web**: full suite **595 passed** (85 files); typecheck + ESLint clean; `next build` succeeds. New/updated suites cover the store slice, BFF PATCH route, `updateClipVisibility` lib, `use-song-actions` publish (guard/rollback), `ShareModal`, `PublishGuardPrompt`, and the ClipCard/SongHeader integrations.

### Known limitations
- **Shared links aren't yet viewable by non-owners** (flagged by the cross-family review, P1). The share URL points at the canonical `/song/{id}`, but that page uses `useRequireAuth` and `GET /clips/{id}` is owner-scoped even for public clips (see the intentional `test_other_users_clip_returns_404_even_when_public`). Public *audio streaming* already works anonymously (US-14.2 `stream_router`); a public, unauthenticated **song page** + a public metadata read is a separate layer-2 story and is out of this web-first story's approved scope. The links produced here will "just work" once that page ships. **Recommend a follow-up issue: "Public song page for shared/public clips."**
- Per project convention, this web story wasn't exercised via a live end-to-end browser demo (no GPU/backend/Mongo demo stack locally); verification rests on the automated suites above.
- Combined `{title, is_public:true}` PATCH that fails the guard rejects atomically (title not saved). In the UI, rename and publish are always separate requests, so this only affects direct API callers.

### DIFF
--- a/src/acemusic/api/routers/clips.py
+++ b/src/acemusic/api/routers/clips.py
@@ -2,12 +2,13 @@
 
 * ``GET    /clips``           → paginated list with search/filter/sort (US-9.4)
 * ``GET    /clips/{id}``      → clip metadata (404 if missing/not owned)
-* ``PATCH  /clips/{id}``      → rename (title is the only writable field)
+* ``PATCH  /clips/{id}``      → rename and/or publish (title, is_public; US-17.6)
 * ``DELETE /clips/{id}``      → remove the record and its stored audio
 * ``GET    /clips/{id}/audio``→ stream audio, byte ranges + ``?format=`` (US-9.3)
 
-CRUD is owner-scoped; only the audio endpoint honors ``is_public``. Access and
-filter rules live in :mod:`acemusic.api.services.clips`.
+CRUD is owner-scoped; the audio/stream endpoints honor ``is_public``, which the
+PATCH publish toggle sets (guarded — see :func:`acemusic.api.services.clips.update_clip_fields`).
+Access and filter rules live in :mod:`acemusic.api.services.clips`.
 """
 
 import asyncio
@@ -87,11 +88,16 @@ class ClipSearchParams(BaseModel):
 
 
 class ClipUpdate(BaseModel):
-    """Rename payload. ``extra="forbid"`` rejects any non-title field with 422."""
+    """Clip edit payload. ``extra="forbid"`` rejects any field other than the two
+    client-writable ones (title, is_public) with 422. Both are optional; an
+    omitted field is left unchanged."""
 
     model_config = ConfigDict(extra="forbid")
 
     title: Annotated[str, Field(min_length=1, max_length=CLIP_TITLE_MAX_LENGTH)] | None = None
+    # US-17.6 publish toggle. Going public is guarded (title + style tags) in the
+    # service; unpublishing is always allowed.
+    is_public: bool | None = None
 
     @field_validator("title")
     @classmethod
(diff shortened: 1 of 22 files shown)


frankbria:
Thanks for the review — addressed in c9bebf8:

- **Major (empty guard on server-422 race):** Fixed the root cause in `doPublish`. When the server returns 422 but the locally-captured clip still looks ready (the fields-changed-since-load race), we no longer recompute an all-false guard; we fall back to the server's specific message (`result.message`) instead. Added a regression test — `shows the server message (not an empty guard) on a 422 for a locally-ready clip` — which exercises the previously-uncovered `guardFailed` branch.
- **Minor (stale SongHeader comment):** The local `optimisticPublic` is a live fallback for a *standalone* uncontrolled `SongHeader`, not dead — but the comment was inaccurate now that US-17.6 provides a persisting parent. Rewrote it to describe the controlled-vs-uncontrolled behavior.

Re the codex P1 (shared links not viewable by non-owners): confirmed real but out of this web-first story's scope — it needs a public/unauthenticated song page + a public metadata read (an existing test intentionally 404s public clips for non-owners). Captured in Known Limitations with a recommended follow-up.

frankbria:
Filed the public-song-page follow-up for the documented Known Limitation (shared links not viewable by non-owners): #250.

PR #257, opened 2026-07-07
feat(web): Repaint Mode Integration (US-18.5)

### LINKED ISSUE #206: US-18.5: Repaint Mode Integration
## User Story

**As a** musician, **I want** to select a range in the editor and regenerate that section with a new prompt, **so that** I can use AI to fix or reimagine specific parts of a clip.

## Description

Integrate the Repaint/Edit workflow (spec section 10.2) into the waveform editor. The musician selects a time range, provides new instructions (prompt, style, lyrics), and the AI regenerates only that section while blending seamlessly with surrounding audio.

## Functional Requirements

- Select a region in the waveform editor to activate Repaint mode
- Repaint panel appears with: prompt/instructions textarea, style override field, lyrics override field
- "Regenerate" button submits the selected range and instructions to the API
- Progress indicator during regeneration
- Result replaces the selected region in the editor view
- Surrounding audio is unchanged; crossfade applied at boundaries
- Result can be undone (returns to pre-repaint state)
- Save as new version to persist the change

## Acceptance Criteria

- [ ] Selecting a region enables the Repaint mode panel
- [ ] Providing instructions and clicking Regenerate submits to the API
- [ ] Regenerated section replaces only the selected range in the waveform
- [ ] Surrounding audio remains intact with smooth transitions
- [ ] Repaint result is undoable
- [ ] Saving creates a new clip version in the workspace

## Dependencies

**Blocked by:** US-18.4 (prior story in this stage). Part of Stage 18: Waveform Editor UI, which is blocked by Stage 17.

## References

- User stories: `user-stories/03-layer-3-web-ui.md` — Stage 18, US-18.5

frankbria:
## US-18.5: Repaint Mode Integration

Closes #206. Wire the **existing** Stage-17 repaint pipeline into the waveform editor so a musician selects a region and regenerates just that section with AI.

### Key finding
The repaint API layer, form fields, and submit→poll job lifecycle were **already built** (US-17.3: `submitRepaint`, `RepaintPayload`, `useClipEdit`, `ReplaceSectionModal`). US-18.5 is the **editor integration**, not a rebuild — so this PR is a thin wiring layer that reuses all of it.

### What changed
- **`RepaintPanel`** (new) — appears inline when a region is selected. Required instructions + optional style (reusing `StyleTextarea`), a Regenerate button, and inline progress/error chrome driven by the existing `useClipEdit` poll loop. Submits the selected region as the range.
- **`WaveformEditor`** — renders `RepaintPanel` while a selection is live; on completion decodes the backend's child clip (`decodeClipAudio`) and folds it into the editor via `pushEdit`.
- **`EditOperation`** — new `repaint` variant so the operation rides the US-18.4 save-provenance seam.

### Design decisions
1. **Result handling — deviation from the issue's CodeRabbit plan (Design Choice 2).** The plan proposed navigating the editor to the new child clip with a separate navigation-based undo. Instead the full backend-blended child clip is pushed onto the **existing US-18.4 snapshot stack** via `pushEdit`. Rationale: the existing **Undo/Redo button covers repaint for free** (unified undo — a cleaner fit for the "undoable" AC than a second undo mechanism), no navigation flash, and it reuses the exact non-destructive pipeline every other edit already flows through. No backend/contract/data-model change — the child clip is still created server-side.
2. **No lyrics field** — `RepaintPayload`/the API don't accept lyrics (matches CodeRabbit Design Choice 1).
3. **Inline panel, not a modal** — the AC says "selecting a region enables the Repaint *panel*".

### Acceptance criteria
- [x] Selecting a region enables the Repaint mode panel
- [x] Providing instructions and clicking Regenerate submits to the API
- [x] Regenerated section replaces only the selected range in the waveform *(backend returns a full crossfade-blended child clip; loaded whole)*
- [x] Surrounding audio remains intact with smooth transitions *(backend-blended boundaries, US-6.3)*
- [x] Repaint result is undoable *(Undo/Redo, unified with all other edits)*
- [x] Saving creates a new clip version in the workspace *(the repaint job creates the child clip server-side; "Save as new version" still works for further local edits)*

### Testing
- New: `RepaintPanel.test.tsx` (4) — fields/validation, `Ns`-coord payload, success→decode→apply, failure+retry.
- New: `WaveformEditor.test.tsx` (2) — panel visibility gated on selection; completed repaint lands as an undoable edit.
- Full web suite green: **104 files / 735 tests**. `tsc --noEmit` + eslint clean.

### Known limitations
- The repaint result assumes the job returns a **full-length** child clip with only `[start,end]` regenerated + blended (same assumption the plan's navigate-to-clip approach relies on).
- Repaint targets the **persisted source clip** at the selection's seconds; repainting on a locally-dirty buffer (US-18.3 edits applied) is out of scope for this story.
- Playback-vs-edited desync after repaint is **pre-existing** editor behavior (the player holds the original track; edits re-render the canvas only).
- **Demo gate:** the live end-to-end demo needs the GPU/backend stack, which isn't available in this environment — verification here is the test suite. Requesting maintainer sign-off on that evidence before merge.

### DIFF
--- a/web/components/editor/EditToolbar.test.tsx
+++ b/web/components/editor/EditToolbar.test.tsx
@@ -80,6 +80,31 @@ describe("EditToolbar", () => {
     expect(p.onGainApply).toHaveBeenCalledWith(0.5)
   })
 
+  it("freezes the whole toolbar (incl. open-popover Apply) when disabled", () => {
+    const onGainApply = vi.fn()
+    const base: EditToolbarProps = {
+      hasSelection: true,
+      onFadeIn: vi.fn(),
+      onFadeOut: vi.fn(),
+      onSilence: vi.fn(),
+      onNormalize: vi.fn(),
+      onGainPreview: vi.fn(),
+      onGainApply,
+      onCrossfade: vi.fn(),
+    }
+    const { rerender } = render(<EditToolbar {...base} disabled={false} />)
+    // Open the Gain popover while enabled, then a repaint starts (disabled=true).
+    fireEvent.click(btn("Gain"))
+    expect(btn("Apply gain")).toBeEnabled()
+
+    rerender(<EditToolbar {...base} disabled />)
+    // Triggers freeze, and the already-open Apply must not commit an edit either.
+    expect(btn("Normalize")).toBeDisabled()
+    expect(btn("Apply gain")).toBeDisabled()
+    fireEvent.click(btn("Apply gain"))
+    expect(onGainApply).not.toHaveBeenCalled()
+  })
+
   it("carries a non-default crossfade slider value through to Apply", () => {
     const p = renderToolbar()
     fireEvent.click(btn("Crossfade"))
--- a/web/components/editor/EditToolbar.tsx
+++ b/web/components/editor/EditToolbar.tsx
@@ -24,6 +24,8 @@ const CROSSFADE_DEFAULT_MS = 100
 
 export type EditToolbarProps = {
   hasSelection: boolean
+  /** Freeze the whole toolbar (e.g. while a repaint job is in flight). */
+  disabled?: boolean
   onFadeIn: () => void
   onFadeOut: () => void
   onSilence: () => void
@@ -36,6 +38,7 @@ export type EditToolbarProps = {
 
 export function EditToolbar({
   hasSelection,
+  disabled = false,
   onFadeIn,
   onFadeOut,
   onSilence,
@@ -90,7 +93,7 @@ export function EditToolbar({
         variant="outline"
         size="sm"
         onClick={onFadeIn}
-        disabled={!hasSelection}
+        disabled={disabled || !hasSelection}
       >
         Fade In
       </Button>
@@ -99,7 +102,7 @@ export function EditToolbar({
         variant="outline"
         size="sm"
         onClick={onFadeOut}
-        disabled={!hasSelection}
+        disabled={disabled || !hasSelection}
       >
         Fade Out
       </Button>
@@ -108,11 +111,17 @@ export function EditToolbar({
         variant="outline"
         size="sm"
         onClick={onSilence}
-        disabled={!hasSelection}
+        disabled={disabled || !hasSelection}
       >
         Silence
       </Button>
-      <Button type="button" variant="outline" size="sm" onClick={onNormalize}>
+      <Button
+        type="button"
+        variant="outline"
+        size="sm"
+        onClick={onNormalize}
+        disabled={disabled}
+      >
         Normalize
       </Button>
 
@@ -122,7 +131,7 @@ export function EditToolbar({
             type="button"
             variant="outline"
             size="sm"
-            disabled={!hasSelection}
+            disabled={disabled || !hasSelection}
           >
             Gain
           </Button>
@@ -165,7 +174,7 @@ export function EditToolbar({
 
       <Popover open={crossfadeOpen} onOpenChange={setCrossfadeOpen}>
         <PopoverTrigger asChild>
-          <Button type="button" variant="outline" size="sm">
+          <Button type="button" variant="outline" size="sm" disabled={disabled}>
             Crossfade
           </Button>
         </PopoverTrigger>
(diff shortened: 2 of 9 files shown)


frankbria:
## Review triage (claude-review)

Both findings were real and are now fixed in `9eaa88a`:

**#1 — In-flight job dropped on selection-clear (confirmed, higher severity).** The poll lived in `RepaintPanel`, which unmounted the moment the selection cleared (canvas click / any edit / undo), killing the `useClipEdit` poll and silently losing a paid generation. Fix: the panel reports its active (polling) state up via `onActiveChange`, and `WaveformEditor` keeps it mounted while a job is active (falling back to the last drag region for the prop). Covered by a new test: *keeps the repaint job alive when the selection is cleared mid-poll*.

**#2 — "Try again" after a decode-only failure re-charged a new job (confirmed, moderate).** Split the two failure modes: a decode/fetch failure (job already succeeded server-side) now retries just the fetch of the retained `childId`; only a genuine job failure resubmits. Covered by: *retries only the decode (no resubmit) after the job succeeded but the fetch failed*. Fixing this surfaced a render bug — the sticky `success` phase kept `busy` true and masked the error behind the spinner — now the error branch wins.

**Nitpicks:** `applying` is reset on the success path; `Ns` payload coords are rounded to ms (`coord()`).

The *error-banner-vanishes-if-you-clear-the-selection-during-a-failed-poll* edge is left as a minor known limitation (you moved off the region; the paid-result path is what mattered).

Full web suite green: 104 files / 737 tests. `tsc` + eslint clean.

frankbria:
## Follow-up finding fixed — concurrent-edit data-integrity (`b98a765`)

Good catch: keeping the poll alive (the #1 fix) opened a window where a user could run a concurrent local edit (Normalize/Crossfade/Undo work without a selection) while a repaint polled. When the repaint landed via `pushEdit`, the backend child clip — computed against the clip as it was at submit time — silently reverted that intervening edit, while it stayed in the `operations` save-provenance log: an audio-vs-provenance mismatch and silent loss of the user's edit.

Fix: **freeze all buffer-mutating affordances while `repaintActive`** — `EditToolbar`, `UndoRedoControls`, and the keyboard shortcuts (cut/copy/paste/delete/undo/redo) each gained an optional disable flag, driven off `repaintActive` in `WaveformEditor`. The buffer is about to be replaced wholesale, so freezing edits during the job is also the conceptually correct behavior. Non-mutating interactions (zoom/scroll/seek/select) stay live. The repaint result itself is unaffected — `repaintActive` is already `false` by the time `onRepainted → pushEdit` fires.

New test: *freezes other edits while a repaint is in flight* (asserts Normalize/Fade In/Undo disabled and a keyboard undo is a no-op mid-poll).

Full web suite green: **104 files / 738 tests**. `tsc` + eslint clean.

frankbria:
## Freeze gap closed — popover Apply buttons (`9c02f4f`)

Right — the `disabled` prop only reached the Gain/Crossfade *triggers*, not the *Apply* buttons inside the open `PopoverContent`, so an already-open popover could still commit an edit mid-poll (relying only on Radix's incidental outside-dismiss, not an explicit guard). Wired `disabled` into both Apply buttons so the commit path itself is frozen, not just the entry point.

New test: opens the Gain popover while enabled, flips the toolbar to `disabled`, and asserts *Apply gain* is disabled and commits nothing.

The minor "a new selection dragged mid-poll is cleared when the repaint lands" note is left as-is — it's not data loss (just requires re-selecting) and the in-flight job/op are unaffected (captured at submit).

Full web suite green: **104 files / 739 tests**. `tsc` + eslint clean.

frankbria:
## Decode-window freeze + dead-retry edge fixed (`acc17f1`)

Correct — the freeze was driven off `active = submitting || polling`, which drops at the `success` transition, but the decode (`decodeClipAudio` fetch + `decodeAudioData`) runs afterward in the separate `applying` effect. That re-enabled the toolbar/undo/shortcuts for the entire decode window — the same concurrent-edit → provenance-mismatch hole, one step later. Fix: fold `applying` into the reported `active`, so the freeze spans the poll **and** the decode until the result lands.

Also fixed the minor dead "Try again": `childId` is now state (not a ref), and the retry falls back to a fresh `handleSubmit` when there's no child id (the odd `clipIds: []` success), so the button is never a silent no-op.

New test: holds the decode promise open and asserts Normalize stays disabled through the decode window, then re-enables once the result lands.

Full web suite green: **104 files / 740 tests**. `tsc` + eslint clean.

PR #258, opened 2026-07-10
Add GLM inline review workflow

frankbria:
Adds the reusable GLM-5.2 inline PR reviewer (frankbria/glm-review): CodeRabbit-style inline comments on defective lines with committable suggestions, bugs-only scope. ZHIPU_API_KEY secret is set. Verified on hai-sh#75.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,11 +27,9 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 5
+    # semver-*-days cooldowns are not supported for github-actions
     cooldown:
       default-days: 14
-      semver-major-days: 14
-      semver-minor-days: 7
-      semver-patch-days: 7
 
   - package-ecosystem: "docker"
     directory: "/docker"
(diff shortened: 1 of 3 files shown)


frankbria:
Addressed review feedback in 73734a9:

- **Pinned the reusable workflow ref** to `b877d15a` (main as of 2026-07-10) instead of floating `@main`, since the job carries write permissions.
- **Fixed the failing `.github/dependabot.yml` check** (pre-existing from #259): Dependabot rejects `semver-*-days` cooldown properties for the `github-actions` ecosystem, so that entry now uses `default-days: 14` only.

Not changed, deliberately:
- **`issues: write` / `id-token: write`**: the called workflow declares exactly these permissions (claude-code-action needs id-token, and issue-comment write for progress tracking), so the caller's grants are already the required minimum.
- **Gating divergence (additions OR deletions ≥ 20 vs. total ≥ 20)**: GitHub Actions expressions don't support arithmetic, so matching claude-code-review's total-based gate would need an extra compute job. The only divergent case (e.g. 15+15) is marginal; not worth the machinery.
- **paths-ignore asymmetry**: intentional — GLM reviews workflow/`pyproject.toml` changes (it reviewed this very PR), while the Claude review skips them.

PR #277, opened 2026-07-13
feat(web): US-19.2 track types and clip import

### LINKED ISSUE #208: US-19.2: Track Types and Clip Import
## User Story

**As a** musician, **I want** to create different track types (AI-generated, uploaded audio, sounds/loops, vocal stems), **so that** I can organize my arrangement by source material.

## Description

Support four track types from spec section 24.2. Each type has a visual indicator and accepts clips of its category. Clips can be imported from the workspace or dragged from the clip library panel.

## Functional Requirements

- Track type selector when adding a new track: AI-Generated, Audio, Sound/Loop, Vocal
- Track type indicated by color label and icon
- Clips from the workspace panel can be dragged onto matching track types
- Sounds/loops placed on loop tracks inherit the project tempo
- Multiple clips on the same track arranged sequentially
- Clips on different tracks play simultaneously (layered)

## Acceptance Criteria

- [ ] All four track types can be created
- [ ] Each type is visually distinguished by color and icon
- [ ] Clips from the workspace can be dragged onto tracks
- [ ] Multiple clips on one track play sequentially
- [ ] Clips on different tracks play simultaneously

## Dependencies

**Blocked by:** US-19.1 (prior story in this stage). Part of Stage 19: Multi-Track Studio UI, which is blocked by Stage 18.

## References

- User stories: `user-stories/03-layer-3-web-ui.md` — Stage 19, US-19.2

frankbria:
## Summary
Implements #208: US-19.2 Track Types and Clip Import

- Four typed studio tracks — AI-Generated (purple, `AiMagicIcon`), Audio (blue, `Upload01Icon`), Sound/Loop (green, `RepeatIcon`), Vocal (orange, `Mic01Icon`) — defined in a new `web/lib/track-types.ts` config; the Add Track button opens a dropdown selector with icon + name + description per type
- Track lanes show the type's icon (accessible label `"<Type> track"`) and use the type color for the strip border and clip blocks
- Clip category is inferred from the clip's persisted `generation_mode` (`"upload"` → Audio, `"stems"` → Vocal, `"sound"` → Sound/Loop, everything else incl. null → AI-Generated); no backend change needed — the API already persists the generation request's `mode`
- Strict type matching enforced authoritatively in the reducer for both drops (`ADD_CLIP`) and cross-track moves (`MOVE_CLIP`); dragover shows valid (`bg-accent`) / invalid (`bg-destructive` tint + native no-drop, via not calling `preventDefault`) feedback using a dataTransfer *type* entry, since `getData()` is unavailable during dragover
- Loop-track tempo inheritance: project `bpm` in studio state (default 120, clamped 60–180 matching backend bounds) editable via a header BPM input; playback rate = `projectBpm / clipBpm` is derived per placement at render/schedule time (stored `clipBpm`, not a frozen multiplier — so changing the project tempo re-stretches placed loops), applied to `AudioBufferSourceNode.playbackRate` and clip-block width
- `?song=` preload creates the track with the clip's inferred type so the preloaded clip always lands

## Acceptance Criteria
- [x] All four track types can be created
- [x] Each type is visually distinguished by color and icon
- [x] Clips from the workspace can be dragged onto tracks (matching types only)
- [x] Multiple clips on one track play sequentially
- [x] Clips on different tracks play simultaneously

## Test Plan
- [x] Unit tests written (TDD approach) — 904 tests passing (`npx vitest run`)
- [x] Diff coverage 95% on changed lines (diff-cover, ≥85% gate)
- [x] Linting + typecheck + `next build` clean (web is not in CI — run locally)
- [x] Internal code review (advisory) completed
- [x] Cross-family review pass: opencode (GLM) — verdict APPROVE, no Critical/Major findings
- [x] Test mutation sanity check completed — 6 targeted mutations (type mapping, rate derivation, reducer validation, offset scaling, dragover validity, source rate application) all killed by tests

- Loop tempo inheritance requires the clip to carry BPM metadata; clips without `bpm` play at 1x on loop tracks (no prompt-for-BPM flow — deferred)
- No time-stretching DSP: rate change via `playbackRate` shifts pitch as well as tempo (standard Web Audio behavior; pitch-preserving stretch is out of scope)
- All stem kinds (vocals/drums/bass/other) classify as Vocal: the backend stores every stem as `generation_mode="stems"` with the kind only in the renamable title; finer stem typing needs a backend schema field (follow-up candidate)
- Sound/loop clips generated before the API began persisting `generation_mode` (or via CLI paths that store other values) classify as AI-Generated; reclassification/manual tagging is not in scope
- Track type is fixed at creation — no type change on an existing track (create a new track instead)
- `REMOVE_TRACK` still has no UI affordance (pre-existing from US-19.1)

## Implementation Notes
- The CodeRabbit issue plan predated the US-19.1 studio code and assumed `generation_mode` values that don't exist; the mapping was corrected to the actually-persisted values (`processor.py` stores the request `mode`, so `"song"`/`"sound"` are real values)
- Deviation from the plan's "store a tempoMultiplier on the placed clip": the placement stores `clipBpm` and the rate is derived, which keeps placed loops in sync with later project-tempo changes

Closes #208

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -202,6 +202,29 @@ Reached from the song menu's "Open in Studio" action (`open-studio` in
   centralizes the wire shape so drag sources and the lane's drop handler can't
   drift apart.
 
+## Track Types and Clip Import (US-19.2)
+
+Tracks are typed — AI-Generated, Audio, Sound/Loop, Vocal — each with its own
+color and icon; the Add Track button opens a type selector dropdown.
+
+- `lib/track-types.ts` — the type config (label/description/color/icon),
+  `inferTrackType(generation_mode)` (`"upload"` → audio, `"stems"` → vocal,
+  `"sound"` → loop, everything else → ai — matching what the API persists),
+  and `placementPlaybackRate` (loop-track clips stretch to the project tempo:
+  `projectBpm / clipBpm`, 1x elsewhere or without BPM metadata).
+- Strict type matching is enforced authoritatively in the reducer (`ADD_CLIP`
+  re-infers from the payload's `generationMode`; `MOVE_CLIP` rejects
+  cross-type track moves). During dragover, lanes show valid/invalid feedback
+  from a `dataTransfer` *type* entry (`setDragTrackType`/`readDragTrackType`
+  in `lib/clip-drag.ts`) because `getData()` is unavailable until drop; an
+  invalid lane skips `preventDefault`, so the browser itself disallows the
+  drop.
+- Project tempo lives in studio state (`bpm`, default 120, clamped 60–180 to
+  mirror the backend's bounds) behind the header's BPM input. Placements store
+  `clipBpm` and the playback rate is derived at render/schedule time, so a
+  tempo change re-stretches placed loops — including mid-playback, where an
+  effective `SET_BPM` bumps `seekEpoch` to reschedule the running sources.
+
 ## Adding components
 
 To add components to your app, run the following command:
--- a/web/app/studio/page.test.tsx
+++ b/web/app/studio/page.test.tsx
@@ -37,6 +37,7 @@ class FakeGainNode {
 
 class FakeSourceNode {
   buffer: AudioBuffer | null = null
+  playbackRate = { value: 1 }
   connect = vi.fn()
   disconnect = vi.fn()
   start = vi.fn()
@@ -165,6 +166,30 @@ describe("StudioPage header", () => {
     expect(
       screen.getByRole("button", { name: "Return to start" })
     ).toBeInTheDocument()
+  })
+
+  it("renders a project tempo input defaulting to 120 BPM (US-19.2)", () => {
+    renderPage()
+    expect(
+      screen.getByRole("spinbutton", { name: "Project tempo (BPM)" })
+    ).toHaveValue(120)
+  })
+
+  it("commits an edited tempo on blur, clamped to the supported range", async () => {
+    const user = userEvent.setup()
+    renderPage()
+    const input = () =>
+      screen.getByRole("spinbutton", { name: "Project tempo (BPM)" })
+
+    await user.clear(input())
+    await user.type(input(), "90")
+    await user.tab()
+    expect(input()).toHaveValue(90)
+
+    await user.clear(input())
+    await user.type(input(), "999")
+    await user.tab()
+    expect(input()).toHaveValue(180)
   })
 })
 
@@ -361,6 +386,15 @@ describe("StudioPage ?song= preload", () => {
     stubStudioFetch()
     renderPage()
     await waitFor(() => expect(screen.getByText("My Song")).toBeInTheDocument())
+  })
+
+  it("creates the preloaded track with the clip's inferred type (US-19.2)", async () => {
+    searchParamsRef.current = new URLSearchParams("song=clip-1")
+    stubStudioFetch({ ...SONG_CLIP_JSON, generation_mode: "sound" })
+    renderPage()
+    await waitFor(() =>
+      expect(screen.getByLabelText("Sound/Loop track")).toBeInTheDocument()
+    )
   })
 
   it("fetches the clip named by the song query param", async () => {
--- a/web/app/studio/page.test.tsx
+++ b/web/app/studio/page.test.tsx
@@ -192,6 +192,22 @@ describe("StudioPage header", () => {
     expect(input()).toHaveValue(180)
   })
 
+  it("re-labels the bars-beats ruler from the project tempo (US-19.2)", async () => {
+    const user = userEvent.setup()
+    renderPage()
+    // 120 BPM: bar = 2s → the "2.1" tick sits at 200px (100px/sec, zoom 1).
+    const tick = () => screen.getByText("2.1").parentElement as HTMLElement
+    expect(tick().style.left).toBe("200px")
+
+    const input = screen.getByRole("spinbutton", { name: "Project tempo (BPM)" })
+    await user.clear(input)
+    await user.type(input, "60")
+    await user.tab()
+
+    // 60 BPM: bar = 4s → the same bar label moves to 400px.
+    expect(tick().style.left).toBe("400px")
+  })
+
   it("reverts to the previous tempo when the input is committed empty (Number('') === 0 footgun)", async () => {
     const user = userEvent.setup()
     renderPage()
--- a/web/app/studio/page.test.tsx
+++ b/web/app/studio/page.test.tsx
@@ -191,6 +191,23 @@ describe("StudioPage header", () => {
     await user.tab()
     expect(input()).toHaveValue(180)
   })
+
+  it("reverts to the previous tempo when the input is committed empty (Number('') === 0 footgun)", async () => {
+    const user = userEvent.setup()
+    renderPage()
+    const input = () =>
+      screen.getByRole("spinbutton", { name: "Project tempo (BPM)" })
+
+    await user.clear(input())
+    await user.type(input(), "140")
+    await user.tab()
+    expect(input()).toHaveValue(140)
+
+    // Clearing and blurring must not commit SET_BPM 0 (which would clamp to 60).
+    await user.clear(input())
+    await user.tab()
+    expect(input()).toHaveValue(140)
+  })
 })
 
 describe("StudioPage playhead", () => {
--- a/web/app/studio/page.tsx
+++ b/web/app/studio/page.tsx
@@ -73,7 +73,9 @@ function TempoInput() {
   const [draft, setDraft] = useState<string | null>(null)
 
   function commit() {
-    if (draft !== null) {
+    // Number("") === 0, so an emptied field would otherwise commit SET_BPM 0
+    // and clamp the project to BPM_MIN — treat blank as "revert" instead.
+    if (draft !== null && draft.trim() !== "") {
       const bpm = Number(draft)
       if (Number.isFinite(bpm)) dispatch({ type: "SET_BPM", bpm })
     }
--- a/web/app/studio/page.tsx
+++ b/web/app/studio/page.tsx
@@ -203,6 +203,7 @@ function StudioTimeline() {
             pxPerSec={pxPerSec}
             durationSec={durationSec}
             displayMode={state.displayMode}
+            bpm={state.bpm}
             onSeek={(sec) => dispatch({ type: "SEEK", sec })}
           />
         </div>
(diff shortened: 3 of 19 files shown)


frankbria:
## Cross-family review — opencode (zai/glm-5-turbo)

## PR #277 Review

### Critical

**`web/lib/track-types.ts:80-86` — `inferTrackType("stems")` maps to `"vocal"` but stems can contain any instrument, not just vocals.** A stem extraction might produce drums, bass, or piano stems. Mapping all stems exclusively to Vocal tracks means those clips can never be placed on AI/Audio/Loop tracks, silently failing every drop. Scenario: user extracts drum stems and can only drag them to Vocal tracks — they appear stuck.

**`web/contexts/studio-context.tsx:127-130` — `ADD_CLIP` rejects clips whose inferred type mismatches, but `MOVE_CLIP` (line 155) only checks `source.trackType !== dest.trackType`. A clip added before US-19.2 (no `generationMode`) silently becomes `"ai"`-type via `inferTrackType(undefined)`, so moving it from an AI track to any other track is silently rejected. Scenario: existing projects with pre-typing clips on any track become immovable cross-type, with no user feedback explaining why.

### Major

**`web/lib/clip-drag.ts:46-55` — `readDragTrackType` iterates `dataTransfer.types` looking for a `TRACK_TYPE_PREFIX` match. If a browser or future drag source happens to include a type string containing the prefix (e.g., `"application/x-ams-track-type-ai-future"`), the `slice` would extract `"ai-future"`, which fails the `includes` guard and returns null — safe but fragile. More concerning: if a malicious page sets `application/x-ams-track-type-ai`, it could spoof a track type to bypass the dragover gate.** Not exploitable beyond the local browser, but the dragover gate could be fooled into showing "valid" feedback for a mismatched payload.

**`web/hooks/use-studio-playback.ts:107-112` — `placementPlaybackRate` is called at schedule time using `state.bpm`, but if the user changes BPM while playing, already-scheduled `BufferSourceNode`s retain their old rate.** Scenario: user drops a 90 BPM loop into a 120 BPM project, starts playback, then changes project tempo to 150 BPM mid-play — the loop continues at the old 4/3 rate for the rest of its scheduled duration.

**`web/components/studio/TrackLane.tsx:76-82` — `onDragOver` checks `dragValidity` and conditionally calls `preventDefault()`, but `onDragEnter` at line 166 also calls `dragValidity` to set the visual state. If a drag starts over a child `ClipBlock` (which doesn't call `stopPropagation`), `dragEnter` may not fire on the lane, leaving stale `"valid"`/`"invalid"` highlighting.** The `dragLeave` child-exit guard helps, but the enter path has no equivalent — a drag that enters the lane via a child will show the wrong highlight color.

**`web/components/studio/TrackLane.tsx:82-95` — `onDrop` does not re-validate the clip type. It trusts the `dragover` gate to have filtered, but the `dragValidity` function returns `"valid"` for drags with no track-type entry (`dragType === null`).** Scenario: a user drags a file from the OS desktop — `readDragTrackType` returns null → `"valid"` → `preventDefault` allowed → drop fires → `parseClipDragData` returns null (good) but any future code that adds a file-drop path would bypass type checking.

### Suggestion

**`web/lib/track-types.ts:80-86` — Consider mapping `"stems"` to `"audio"` instead of `"vocal"`, or introducing a separate `"stem"` track type.** Vocal stems are a subset; mapping all stems to vocal is a UX footgun.

**`web/contexts/studio-context.tsx:127-130` — When `ADD_CLIP` rejects a mismatched type, the reducer silently returns the unchanged state.** The `onDrop` handler in `TrackLane.tsx` dispatches and moves on — the user gets no toast/error explaining why their clip wasn't placed. Adding a rejected-type action or an error callback would close the feedback loop.

### Nitpick

**`web/lib/timeline.ts:174` — The `rateFor` default parameter `() => 1` is correct but the function signature is now 4 lines long.** Fine as-is.

**`web/lib/track-types.ts:72-73` — `BPM_MIN = 60` and `BPM_MAX = 180` are duplicated from the backend.** Consider deriving from a shared constant or at least a code comment linking to the backend source, so drift is caught on review.

---

**Verdict: REQUEST_CHANGES**

The `stems → vocal` mapping (Critical) will silently block users from placing non-vocal stems on any track except Vocal, which is incorrect behavior. The silent `ADD_CLIP` rejection with no user feedback (Major) means users will think the drag-and-drop is broken rather than intentionally filtered. Both should be addressed before merge.

frankbria:
## Triage of cross-family review (verified against code before acting)

| Finding | Severity claimed | Verdict | Action |
|---|---|---|---|
| `inferTrackType("stems")` → vocal, but stems include drums/bass/other | Critical | Partially correct, contract-bound | Rebutted + documented. The failure scenario ("can never be placed / appear stuck") is wrong — all stem clips place on Vocal tracks. The `stems → Vocal` mapping is the issue #208 contract ("Vocal: `stems`"). The backend stores every stem kind as `generation_mode="stems"` with the kind only in the user-renamable `title`, so finer typing needs a backend schema field — out of scope; added to Known Limitations. |
| Pre-US-19.2 clips become immovable cross-type | Critical | Incorrect | Rebutted. Studio arrangements are not persisted (in-memory per session); no placement can predate the typed `ADD_CLIP` path, so the scenario cannot occur. |
| BPM change mid-play leaves scheduled sources at the old rate | Major | **Correct** | **Fixed in 3e753e7**: `SET_BPM` bumps `seekEpoch`, reusing the seek reschedule path, so in-flight playback restarts its sources at the new tempo-derived rates. Covered by a reducer test; mutation-checked. |
| Drag type entry spoofable during dragover | Major | Correct but cosmetic-only | Rebutted. The dragover entry only drives hover highlighting; the drop is validated authoritatively in the reducer (`ADD_CLIP` re-infers from the JSON payload, `MOVE_CLIP` checks track types), so a spoofed entry cannot place a mismatched clip. |
| `dragEnter` may not fire when entering via a child ClipBlock | Major | Incorrect | Rebutted. React synthetic `dragEnter` bubbles; entering the lane through a child still invokes the lane's handler and recomputes validity. |
| `onDrop` does not re-validate type | Major | Incorrect | Rebutted. The reducer is the authoritative re-validation on every drop; `onDrop` merely forwards the payload. A null-type external drag falls through to `parseClipDragData` → null → no dispatch. |
| Map stems → audio or new stem type | Suggestion | — | Skipped: contradicts the issue contract; needs backend metadata. |
| Toast/error on rejected drop | Suggestion | — | Skipped: silent rejection with dragover feedback is the behavior specified in the issue plan ("reject the drop silently"). |
| BPM constants duplicated from backend | Nitpick | — | Already addressed: `track-types.ts` comments the mirror of `constants.py` bounds. |

frankbria:
## Triage of Claude bot + GLM precision reviews

| Finding | Verdict | Action |
|---|---|---|
| TempoInput commits an emptied field as `SET_BPM 0` → clamped to 60 (Claude Bug 1; GLM precision minor) | Correct | **Fixed in 25cd5db** — blank draft reverts instead of committing; regression test added and mutation-checked. |
| BPM change mid-playback doesn't reschedule loop sources (Claude Bug 2) | Correct | **Already fixed in 3e753e7** (pushed while the review ran) — `SET_BPM` bumps `seekEpoch`, reusing the seek reschedule path. |
| `ClipBlock` width unguarded against `playbackRate = 0` (Claude minor) | Skipped | The only producer, `placementPlaybackRate`, returns 1 for any `clipBpm <= 0` and can never yield 0/negative; a runtime clamp for an impossible input is defensive noise. |

frankbria:
## Demo verification (Phase 11 gate) — PASSED at 25cd5db

Live stack: FastAPI (:8212, fresh Mongo db) + `next dev` (:3212), authenticated agent-browser session, four seeded WAV clips (one per `generation_mode` category). Full narrated demo with screenshots: `docs/demos/us-19.2-track-types.md` (local, not committed).

| Criterion | Action | Outcome evidence | Status |
|---|---|---|---|
| All four track types can be created | Add Track dropdown → clicked each option | 4 lanes with type labels AI-Generated/Audio/Sound-Loop/Vocal | VERIFIED |
| Types visually distinguished | Read live lane strips via JS | 4 distinct border colors (purple/blue/green/orange) + 4 distinct icon aria-labels | VERIFIED |
| Workspace clips drag onto matching tracks | 4 real browser drags + 1 mismatched drag | 4 blocks placed at correct widths; AI clip dropped on Vocal track left it unchanged (1 block, "vocals") | VERIFIED |
| Loops inherit project tempo | 90 BPM 8s loop @ 120 BPM project; edited BPM live | Block width 600px (=8s ÷ 4/3); live BPM 90 re-stretched to 800px; instrumented source node played at `rate 1.333` | VERIFIED |
| Same-track clips play sequentially | Two clips arranged 0–6s / 6.2–12.2s, Play | Instrumented `source.start`: `when 0` then `when 6.2`; playhead advanced 100px/s through both | VERIFIED |
| Cross-track clips play simultaneously | Overlapping clips on 3 tracks, Play | Three sources all scheduled `when 3.68` on one running context | VERIFIED |

frankbria:
## Final Triage Summary (PR #277)

Cutoff progression: PR open 17:23Z → cycles at 17:39Z, 17:50Z, 18:08Z, 18:23Z. All reviewer feedback triaged; final GLM precision pass at HEAD (6b2879a): **no defects found**.

### Fixed
- [claude-bot + GLM] TempoInput commits emptied field as SET_BPM 0 → fixed in 25cd5db
- [opencode cross-family] BPM change mid-playback keeps stale loop rates → fixed in 3e753e7
- [claude-bot] SET_BPM bumps seekEpoch on no-op commits (spurious audio restart) → fixed in 09a442d
- [claude-bot] stale seekEpoch field comment → fixed in 09a442d
- [GLM precision] TimeRuler ignored project tempo (no bpm prop) → fixed in eaa5980
- [claude-bot] aria-label on role=generic span (ARIA-prohibited) → fixed in 6b2879a

- [opencode] stems→vocal mapping wrong for drum/bass/other stems — contract-bound: issue #208 defines Vocal ← `stems`; the backend has no per-stem-kind field; documented in Known Limitations
- [opencode] pre-US-19.2 clips immovable — inaccurate: studio state is not persisted, no placement predates the typed ADD_CLIP path
- [opencode] drag type entry spoofable — cosmetic only: reducer re-validates authoritatively on drop
- [opencode] dragEnter may miss child entries — inaccurate: React synthetic dragEnter bubbles
- [opencode] onDrop doesn't re-validate — inaccurate: the reducer is the re-validation
- [claude-bot] ClipBlock playbackRate=0 guard — impossible input: placementPlaybackRate returns 1 for any clipBpm ≤ 0
- [claude-bot] token missing from playback effect deps — pre-existing, predates this PR, benign (re-read on every play)
- [opencode] toast on rejected drop — issue plan specifies silent rejection with dragover feedback

Demo gate: PASSED (see demo verification comment). Merging.

PR #280, opened 2026-07-14
feat(web): US-19.5 master bus controls (volume, EQ, compressor, limiter, metering)

### LINKED ISSUE #211: US-19.5: Master Bus Controls
## User Story

**As a** musician, **I want** a master bus with volume, EQ, compressor, and limiter, **so that** I can shape the overall mix before exporting.

## Description

Build the master bus section with master volume, 3-band EQ (low/mid/high shelf), compressor (threshold, ratio, attack, release), and limiter. These controls affect the summed output of all tracks.

## Functional Requirements

- Master volume fader
- 3-band EQ: low shelf, mid peak, high shelf with frequency and gain controls
- Compressor: threshold, ratio, attack, release knobs
- Limiter: ceiling knob
- All controls update the audio output in real time
- Visual metering (peak and RMS levels for left/right channels)

## Acceptance Criteria

- [ ] Master volume controls the overall output level
- [ ] EQ bands audibly affect the frequency spectrum
- [ ] Compressor and limiter respond to dynamics as expected
- [ ] Visual metering shows peak and RMS levels in real time

## Dependencies

**Blocked by:** US-19.4 (prior story in this stage). Part of Stage 19: Multi-Track Studio UI, which is blocked by Stage 18.

## References

- User stories: `user-stories/03-layer-3-web-ui.md` — Stage 19, US-19.5

frankbria:
## Summary
Implements #211 (US-19.5): Master Bus Controls for the multi-track studio.

- Master bus Web Audio chain spliced between the track sum and the destination: `sum → low shelf → mid peak → high shelf → compressor → master volume → limiter (fixed 20:1, 1ms attack, 50ms release, hard knee) → destination`, with a `ChannelSplitter → 2× AnalyserNode` metering tap on the final output. Built once per AudioContext; survives play/pause cycles.
- Master volume sits **before** the limiter (but after the compressor) so the configured ceiling is a true output cap even at +6 dB fader boost, while fader moves never change compression amount.
- New `masterBus` state slice in the studio reducer (`SET_MASTER_VOLUME`, `SET_MASTER_EQ`, `SET_MASTER_COMPRESSOR`, `SET_MASTER_LIMITER_CEILING`) with range clamping and NaN guards; live retune effect pushes values onto the running audio graph without rescheduling playback (same ref pattern as US-19.4 track controls).
- `MasterBusPanel` (dispatch-only sliders for volume / 3-band EQ / compressor / limiter ceiling, with a live limiter-active indicator reading the node's `reduction`) and `StereoMeter` (rAF-driven peak + RMS bars, 1.5s peak-hold with decay, clip indicators, dB scale ticks). The studio's right aside is now tabbed: Workspace | Master.
- Pure metering math (`lib/metering.ts`) and master-bus constants/types (`lib/master-bus.ts`) kept side-effect-free and unit-tested.

## Acceptance Criteria
- [x] Master volume controls the overall output level — reducer → `dbToGain` → GainNode, live-retuned without rescheduling
- [x] EQ bands audibly affect the frequency spectrum — three BiquadFilterNodes in the signal path, all params live-wired
- [x] Compressor and limiter respond to dynamics as expected — DynamicsCompressorNodes with user params / fixed high-ratio limiter + ceiling
- [x] Visual metering shows peak and RMS levels in real time — rAF loop over post-limiter stereo analysers

## Test Plan
- [x] Unit tests written (TDD approach) — red confirmed before each implementation step
- [x] All tests passing — 1056/1056 (125 files)
- [x] Diff coverage 98% on changed lines (gate: ≥85%)
- [x] Linting clean (0 errors; 1 pre-existing warning in gitignored coverage artifacts), typecheck clean, production build succeeds — note: web/ is not in CI, verified locally
- [x] Internal code review (advisory) completed — 1 Major finding fixed (see notes)
- [x] Cross-family review pass: opencode (GLM) — verdict APPROVE, 1 suggestion applied
- [x] Test mutation sanity check — 5/5 targeted mutations (dB math, chain wiring, reducer clamp, panel dispatch, meter read) each broke at least one test

- The limiter is a high-ratio `DynamicsCompressorNode` (soft limiter, hard knee), not a true look-ahead brickwall — per the issue plan's chosen design. Live-demo measured behavior: a -6 dB ceiling caps a -0.4 dBFS tone at ≈ -2 dBFS (consistent and reproducible) rather than exactly -6, because Chromium's DynamicsCompressorNode applies fixed internal makeup gain derived from threshold/knee/ratio. Exact ceiling adherence needs the already-deferred AudioWorklet limiter, which can replace the node later without touching the UI contract.
- Master bus params are set via direct `AudioParam.value` assignment (same as US-19.4 track gain/pan); very fast slider drags on EQ could produce zipper artifacts. Matches the existing engine-wide pattern; smoothing ramps would be an engine-wide change.
- Master bus settings are session-only (not persisted), consistent with the rest of the studio arrangement state.
- Compressor knee is left at the Web Audio default (30 dB) and not exposed in the UI.
- Meters are active only after the first playback run creates the AudioContext (browser autoplay policy).

## Implementation Notes
- The CodeRabbit plan on the issue assumed `web/` was greenfield; paths and patterns were adapted to the real US-19.1–19.4 infrastructure (reducer context, playback-hook engine, existing `dbToGain`).
- Sliders instead of rotary knobs — no knob primitive exists in the codebase and the horizontal Shadcn Slider is the established control.
- The internal review caught that an earlier revision placed master volume after the limiter, letting +6 dB of fader boost bypass the ceiling; the chain was corrected and the wiring test's connect-order expectations updated to the corrected contract (commit 3bf4255).
- One pre-existing test assertion updated: master-chain construction adds one GainNode, so a `gains.toHaveLength(3)` count in the US-19.4 wiring test became 4 (commented inline; no assertion weakened).

Closes #211

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -274,6 +274,36 @@ action, on top of the name edit from US-19.1:
   the finished clip to the end of the track once generation completes;
   regeneration is additive, never a replacement.
 
+## Master Bus Controls (US-19.5)
+
+The studio's right aside is now tabbed **Workspace | Master**
+(`app/studio/page.tsx`); the Master tab holds `MasterBusPanel` with a live
+`StereoMeter` on top. The playback engine is owned by `StudioView` (called
+once) so both the timeline and the panel observe the same `AudioContext`.
+
+- **Audio chain** — `hooks/use-studio-playback.ts` splices the bus between
+  the track sum and the destination: sum → low shelf → mid peak → high shelf
+  (BiquadFilters) → compressor → master volume → limiter
+  (`DynamicsCompressorNode`, fixed 20:1 / 1ms attack / 50ms release / hard
+  knee) → destination, with a `ChannelSplitter` → dual `AnalyserNode` (L/R)
+  metering tap on the final output. Built once per `AudioContext`; survives
+  play/stop. Master volume sits *before* the limiter so the configured
+  ceiling stays a real output cap, and *after* the compressor so fader moves
+  don't change compression amount.
+- **State** — `masterBus: MasterBusState` on `StudioState`
+  (`lib/master-bus.ts` has the types, defaults, and ranges), set via
+  `SET_MASTER_VOLUME` / `SET_MASTER_EQ` / `SET_MASTER_COMPRESSOR` /
+  `SET_MASTER_LIMITER_CEILING` (clamped + NaN-guarded in the reducer). A
+  dedicated effect retunes the live nodes on change — same
+  never-restart-the-run ref pattern as the per-track controls.
+- **Metering** — `lib/metering.ts` (pure peak/RMS dBFS math + peak-hold
+  decay) drives `components/studio/StereoMeter.tsx`'s rAF loop: peak + RMS
+  bars, 1.5s peak-hold, clip indicators, -60 dBFS floor.
+- **Limiter caveat** — the ceiling is a soft limiter built on
+  `DynamicsCompressorNode`; Chromium's fixed internal makeup gain means a
+  -6 dB ceiling caps a near-0 dBFS tone at ≈ -2 dBFS rather than exactly -6.
+  Exact adherence needs an AudioWorklet limiter (deferred by design).
+
 ## Adding components
 
 To add components to your app, run the following command:
--- a/web/app/studio/page.test.tsx
+++ b/web/app/studio/page.test.tsx
@@ -26,8 +26,15 @@ import StudioPage from "@/app/studio/page"
 import { AuthContext } from "@/contexts/auth-context"
 import { PlayerProvider } from "@/contexts/player-context"
 import { TRACK_STRIP_PX } from "@/lib/timeline"
+import {
+  FakeAnalyserNode,
+  FakeBiquadFilterNode,
+  FakeChannelSplitterNode,
+  FakeDynamicsCompressorNode,
+} from "@/test/audio-stubs"
+import { stubRaf } from "@/test/raf-stub"
 
-// --- Minimal Web Audio + rAF stand-ins (jsdom has neither), mirroring
+// --- Minimal Web Audio stand-ins (jsdom has none), mirroring
 // hooks/use-studio-playback.test.tsx's stubs. ---
 
 class FakeGainNode {
@@ -51,41 +58,6 @@ class FakeSourceNode {
   stop = vi.fn()
 }
 
-// US-19.5's master bus chain runs on every play, so this page-level harness
-// needs the same node stand-ins as hooks/use-studio-playback.test.tsx even
-// though this file doesn't assert on them directly.
-class FakeBiquadFilterNode {
-  type: BiquadFilterType = "lowpass"
-  connect = vi.fn()
-  disconnect = vi.fn()
-  frequency = { value: 350 }
-  gain = { value: 0 }
-  Q = { value: 1 }
-}
-
-class FakeDynamicsCompressorNode {
-  connect = vi.fn()
-  disconnect = vi.fn()
-  threshold = { value: -24 }
-  knee = { value: 30 }
-  ratio = { value: 12 }
-  attack = { value: 0.003 }
-  release = { value: 0.25 }
-  reduction = 0
-}
-
-class FakeChannelSplitterNode {
-  connect = vi.fn()
-  disconnect = vi.fn()
-}
-
-class FakeAnalyserNode {
-  connect = vi.fn()
-  disconnect = vi.fn()
-  fftSize = 2048
-  getFloatTimeDomainData = vi.fn()
-}
-
 function stubAudioContext() {
   const box: { instance: { currentTime: number } | null } = { instance: null }
   class FakeAudioContext {
@@ -105,33 +77,6 @@ function stubAudioContext() {
   }
   vi.stubGlobal("AudioContext", FakeAudioContext)
   return box
-}
-
-function stubRaf() {
-  let nextId = 1
-  const callbacks = new Map<number, FrameRequestCallback>()
-  vi.stubGlobal(
-    "requestAnimationFrame",
-    vi.fn((cb: FrameRequestCallback) => {
-      const id = nextId++
-      callbacks.set(id, cb)
-      return id
-    })
-  )
-  vi.stubGlobal(
-    "cancelAnimationFrame",
-    vi.fn((id: number) => {
-      callbacks.delete(id)
-    })
-  )
-  return {
-    /** Invoke every currently-scheduled rAF callback once, at time `t`. */
-    tick(t: number) {
-      const due = [...callbacks.entries()]
-      callbacks.clear()
-      for (const [, cb] of due) cb(t)
-    },
-  }
 }
 
 const authValue = {
--- a/web/app/studio/page.test.tsx
+++ b/web/app/studio/page.test.tsx
@@ -51,6 +51,41 @@ class FakeSourceNode {
   stop = vi.fn()
 }
 
+// US-19.5's master bus chain runs on every play, so this page-level harness
+// needs the same node stand-ins as hooks/use-studio-playback.test.tsx even
+// though this file doesn't assert on them directly.
+class FakeBiquadFilterNode {
+  type: BiquadFilterType = "lowpass"
+  connect = vi.fn()
+  disconnect = vi.fn()
+  frequency = { value: 350 }
+  gain = { value: 0 }
+  Q = { value: 1 }
+}
+
+class FakeDynamicsCompressorNode {
+  connect = vi.fn()
+  disconnect = vi.fn()
+  threshold = { value: -24 }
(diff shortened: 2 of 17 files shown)


frankbria:
## Cross-family review — opencode (zai/glm-5-turbo)


---

## PR #280 Review: feat(web): US-19.5 master bus controls

### Findings

**Major — `web/hooks/use-studio-playback.ts:169-171` — `ensureMasterChain` reads `masterGainRef.current!` which is null until `ensureContext` runs, but the function is called *after* `ensureContext` inside a closure that already checks context. This is safe in practice but the non-null assertion is fragile.** Failure scenario: if the call site ordering changes (e.g., someone calls `ensureMasterChain` before the gain is created), you get a runtime crash with no useful error message — just `TypeError: Cannot read properties of null`. This is more of a latent risk than a current bug.

**Major — `web/components/studio/StereoMeter.tsx:139-144` — The rAF `tick` function re-schedules itself *before* `setState`, so the next frame callback can read stale `analyserLeft.current`/`analyserRight.current` if refs are swapped mid-frame.** Failure scenario: The refs are `RefObject` (not `MutableRefObject`), and the parent replaces `useStudioPlayback`'s return value each render. Because the effect's closure captures the *ref objects* (which are stable), this is actually safe — the `.current` is read at tick time. Downgrading this to **no finding** on closer inspection — the `useEffect` dependency is `[analyserLeft, analyserRight]` which are stable ref objects, and `.current` is read inside the rAF callback. Correct.

**Major — `web/components/studio/MasterBusPanel.tsx:86` — `onValueChange={([v]) => onValueChange(v)}` destructures the first element of the array without guarding against empty arrays.** Failure scenario: If the Slider component ever emits `onValueChange([])` (e.g., during a rapid drag-release race), `v` would be `undefined` and dispatched as `NaN` to the reducer. The reducer does guard against `NaN` for all master bus actions, so this degrades to a no-op rather than a crash. **Severity: Suggestion** — the reducer guard makes this safe, but the destructuring is still brittle.

**Major — `web/hooks/use-studio-playback.ts:200` — The limiter connects to `ctx.destination` AND to the `splitter`, meaning the metering analysers tap *post-limiter* (correct per the comment), but this means the meter shows levels *after* limiting, not the pre-limiter signal.** This is actually a design choice, not a bug — the comment at line 192 says "the metering splitter taps the limiter's output — the actual signal reaching the destination." This matches AC#4 ("peak and RMS levels in real time"). **No finding** — correct behavior.

**Suggestion — `web/lib/metering.ts:12-16` — `linearToDbfs` uses `abs <= 0` which treats `-0.0` as silence.** Failure scenario: In IEEE 754, `-0.0 <= 0` is true, so negative zero returns `METER_FLOOR_DB` instead of `0 dBFS`. In practice, AnalyserNode time-domain data shouldn't produce `-0.0` as a meaningful signal level, so this is cosmetically correct but technically imprecise. Very minor.

**Suggestion — `web/components/studio/MasterBusPanel.tsx:111-120` — `LimiterIndicator` runs a perpetual rAF loop even when the component is on a hidden tab (the Workspace tab hides the Master tab's content).** Failure scenario: When the user is on the Workspace tab, the LimiterIndicator is unmounted (TabsContent unmounts by default in Radix), so the rAF is actually cleaned up. **No finding** — Radix TabsContent unmounts hidden content by default, so this is handled.

**Suggestion — `web/lib/metering.ts:57-58` — The decay formula `prev.db - (PEAK_DECAY_DB_PER_SEC * (elapsed - PEAK_HOLD_MS)) / 1000` can produce very large negative numbers if `nowMs` is much larger than `heldAtMs` (e.g., tab was backgrounded for minutes, then foregrounded).** Failure scenario: `decayed` could be `-2000 dB` or worse. The `decayed <= incomingDb` guard on line 59 catches this and clamps to `incomingDb`, and `dbToPercent` in the meter clamps to `METER_FLOOR_DB`. Safe, but the intermediate value is mathematically extreme. Harmless.

**Nitpick — `web/lib/master-bus.ts:83-85` — `clamp` is a trivial one-liner that duplicates `Math.min`/`Math.max` patterns already used inline elsewhere in the codebase.** Not a problem, but note that `studio-context.tsx` line 12 imports `clamp` from here while `SET_TRACK_VOLUME` (line 231-234) uses inline `Math.min/Math.max`. Minor inconsistency.

### Contract Verification

| AC | Status | Evidence |
|---|---|---|
| 1. Master volume controls overall output | **Pass** | `use-studio-playback.ts:199`: `compressor → masterVolume → limiter → destination`. Master volume gain node in the chain. Tests verify live retuning without rescheduling. |
| 2. EQ bands affect frequency spectrum | **Pass** | 3-band chain: `lowshelf → peaking → highshelf` (line 194-196). Tests verify live retuning of all params. Reducer clamps and validates. |
| 3. Compressor/limiter respond to dynamics | **Pass** | Compressor node with user-adjustable threshold/ratio/attack/release. Limiter with fixed 20:1 ratio, fast attack (1ms), ceiling as threshold. LimiterIndicator reads live `.reduction`. |
| 4. Visual metering shows peak/RMS in real time | **Pass** | StereoMeter with rAF loop, peakDbfs/rmsDbfs pure math, peak-hold decay, clip indicator. Tests verify loud/silent differentiation, peak hold, and cleanup. |

### Verdict

**APPROVE** — All four acceptance criteria are met with solid test coverage, correct audio graph topology, proper input validation/clamping in the reducer, and clean separation of concerns (pure math in `metering.ts`, state in `master-bus.ts`, audio wiring in the hook, UI in the panel). The findings above are suggestions and latent risks, not correctness issues.

---
### Author triage (severity policy)
- No Critical/Major findings survived the reviewer's own verification (two provisional Majors were self-downgraded to no-finding/suggestion in the review text).
- Slider empty-destructure suggestion: skipped — the reducer's Number.isFinite guards make it a no-op by design.
- `-0.0`/decay-extreme/`clamp` nitpicks: skipped — harmless per the reviewer's own analysis; intermediate values are clamped downstream.
- Non-null assertion in `ensureMasterChain`: skipped — call order is enforced by the single call site right after `ensureContext()`, matching the existing `masterGainRef` idiom.

frankbria:
Triage of the claude-review findings:

1. **SET_MASTER_EQ same-value short-circuit** — applied in 7cac23c (early return + reducer identity test, matching the SET_BPM pattern).
2. **Move generic `clamp` to `lib/utils.ts`** — skipped: it has a single consumer outside its module and sits next to the range constants it clamps; will move if a second non-audio consumer appears.
3. **Consolidate test fakes/stubRaf into shared helpers** — skipped for this PR: self-contained per-file harnesses are the pre-existing convention in these test files (the FakeAudioContext duplication between page.test and hook.test predates this PR's extensions); consolidating is a cross-cutting test refactor better done separately.

frankbria:
## Final Triage Summary (PR #280)

Cutoff cycles: PR open → 06:33Z (Phase 10) → 06:47Z → 07:05Z (final). All CI checks green on HEAD 01f35b0.

### Fixed
- [claude-review] SET_MASTER_EQ no-field identity short-circuit → 7cac23c
- [claude-review] Missing `limiter.connect(ctx.destination)` wiring assertion → b70d0e4 (mutation-checked: deleting the connect fails the test)
- [claude-review] Duplicated test stubs/stubRaf → consolidated into `web/test/audio-stubs.ts` + `web/test/raf-stub.ts` → 2d918a0
- [claude-review] SET_MASTER_COMPRESSOR no-field identity short-circuit (parallel gap) → 01f35b0
- [opencode pre-PR] Limiter release left at Web Audio default → 6e77b8b
- [internal advisory] Master volume placed after the limiter defeated the ceiling → chain corrected in 3bf4255
- [live demo] Limiter soft knee (30 dB default) prevented the ceiling from capping realistic material → hard knee in 9694346, re-verified live
- [claude-review] Stale chain-order bullet in PR description → PR body updated

- [claude-review] `clamp` → `utils.ts` — single external consumer, co-located with the ranges it clamps (reviewer accepted)
- [opencode] compressor knee constant — documents an untouched browser default; YAGNI
- [opencode] slider empty-destructure — reducer's Number.isFinite guards make it a no-op by design
- [opencode] `-0.0` / decay-extreme nitpicks — clamped downstream, no failure scenario
- [claude-review 07:01] "orphaned useAuth in StudioTimeline" — **inaccurate**: `accessToken` is consumed at `app/studio/page.tsx:212` (`token={accessToken}`)
- [claude-review 07:10] "destination assertion is vacuous, FakeAudioContext has no destination property" — **inaccurate**: `hooks/use-studio-playback.test.tsx:85` defines `destination = {}` as a runtime instance property; the assertion at line 1213 checks identity against that object, and the mutation check (deleting `limiter.connect(ctx.destination)`) fails the test.

No unresolved Critical/Major findings. Demo evidence: all four acceptance criteria verified with numeric meter measurements (docs/demos/us-19.5-master-bus.md).

Issue #283, opened 2026-07-15
P1.2 Public rate limiter collapses to one shared bucket behind the BFF proxy


frankbria:
## Problem

The public streaming rate limiter keys on the **peer IP** (`request.client.host`), and deliberately does not trust `X-Forwarded-For`:

```python
# src/acemusic/api/utils/rate_limit.py
def _client_key(request): ...  # raw peer IP; XFF intentionally untrusted
```

US-20.0 (#250) added the first same-origin BFF proxies that put real browser traffic in front of the two rate-limited public routes (`GET /clips/{id}/public` and `GET /clips/{id}/stream`). Those proxies call the backend server-side and forward no client-identifying header, so **every visitor arrives as the same IP** — the Next server's egress address.

The limiter therefore no longer limits *per visitor*; it limits *the whole site* to one shared bucket. Default is `stream_rate_limit_per_minute = 100` (`src/acemusic/api/settings.py:177`).

Effect: one user seeking around a track (each seek is a `Range` request) or a modest number of concurrent viewers can exhaust the shared budget and cause **429s for every other visitor** on every `/song/*` page. It is an unauthenticated availability problem against the public sharing feature, and it gets worse as Stage 20 (Explore #213, Search #214, Feed #216, Profile #217) drives more anonymous public traffic through the same proxies.

Note direct-to-backend traffic is unaffected — it still rate-limits per real IP correctly. Only BFF-proxied traffic collapses into one bucket.

## Why it wasn't a problem before

Before #250, nothing in the web app called `/stream` (`clipAudioUrl` pointed at a dead `/api/v1/...` path), so the rate-limited routes carried no proxied browser traffic.

## Suggested approach

Needs a deployment-topology decision, which is why it isn't fixed inline in #250:

- Have the BFF forward the real client IP (from its own inbound request) as a header, and teach `_client_key` to trust that header **only** when the peer is a configured trusted proxy (e.g. `ACEMUSIC_API_TRUSTED_PROXIES`). Falling back to the peer IP when unset preserves today's behavior — a safe default. The current "never trust XFF" stance is correct precisely because the backend may be directly reachable, so blanket-trusting the header would make the limiter trivially spoofable.
- Or rate limit at the edge/BFF layer instead, where the real client IP is known.
- Or key on a per-visitor signed cookie issued by the BFF (weaker: cookies are cheap to discard).

Whatever is chosen, keep the limiter meaningful for direct backend access too.

## Acceptance criteria

[...]

Issue #282, opened 2026-07-15
P1.1 Private clips cannot play: <audio> can't authenticate to the stream proxy


frankbria:
## Problem

`<audio src>` cannot attach an `Authorization` header, so audio playback only works for **public** clips. A signed-in user cannot play **their own private/unpublished clip** anywhere in the app — workspace clip cards, the waveform editor, the song page, and the full-song section preview all go through `trackFromClip` → `clipAudioUrl` → `GET /api/clips/{id}/stream`, which reaches the backend with no credentials and is therefore resolved as anonymous → `404` for a private clip.

This is **pre-existing, not a regression**. Before US-20.0 (#250), `clipAudioUrl` returned `/api/v1/clips/{id}/audio` — a path this Next app never served (no `/api/v1` route, no `rewrites()` in `next.config.ts`), so playback was dead for *every* clip, public or private. US-20.0 repointed it at the same-origin `/stream` proxy, which fixed public clips and left private ones where they were.

Since unpublished clips are the default state of everything a user creates, this affects the primary creation loop.

## Evidence

- `web/lib/clips.ts` — `clipAudioUrl()` and its comment ("A private clip therefore does not play through this path").
- `web/hooks/use-audio-engine.ts:56` — the value is consumed as a real `<audio>` src.
- `src/acemusic/api/services/clips.py:172-188` — `get_clip_for_streaming(clip_id, None)` 404s a private clip for a headerless request.
- `trackFromClip` callers: `components/workspace/ClipCard.tsx`, `components/editor/WaveformEditor.tsx`, `components/song/SongPlayer.tsx`, `components/song/full-song/SectionPreviewPlayer.tsx`.

## Suggested approach

The BFF already holds an httpOnly session cookie (`ams_refresh_token`) and runs server-side, so it can authenticate a cookie-bearing `<audio>` request that the browser cannot sign itself:

- In `web/app/api/clips/[id]/stream/route.ts`, when no `Authorization` header is present, fall back to the session cookie: exchange it for an access token server-side (as `POST /api/auth/refresh` does) and forward that to the backend.
- Careful with the rotating single-use refresh token — a per-request rotation on every media request is wrong. Probably wants a short-lived cached access token or a dedicated media token, not a refresh-per-range-request.
- Alternative: a signed, short-lived, clip-scoped media URL (`?token=…`) issued by the BFF. Avoid putting the real access token in the URL — it leaks via logs/referrers.
[...]

Issue #219, opened 2026-06-22
US-20.7: Publish and Visibility Controls


frankbria:
## User Story

**As a** musician, **I want** to set my clips as private, unlisted, or public from any clip location, **so that** I control who can see my work.

## Description

Implement the three-state visibility toggle (private, unlisted, public) from spec section 33. Publishing requires a title and at least one style tag.

## Functional Requirements

- Visibility toggle accessible from clip card, song detail page, and workspace
- Three states: Private (default), Unlisted (link-only access), Public (visible in feeds/search/explore)
- Publishing to Public requires: title is set and at least one style tag exists
- If requirements are not met, show an inline prompt to add the missing fields
- State change persists immediately via API
- Visual badge on clip cards indicating current visibility state

## Acceptance Criteria

- [ ] Visibility toggle switches between private, unlisted, and public
- [ ] Publishing without a title or style tag shows a prompt
- [ ] Visibility change persists after page reload
- [ ] Clip cards show a badge indicating current visibility state
- [ ] Unlisted clips are accessible via direct link but not in search/explore

## Dependencies

**Blocked by:** US-20.6 (prior story in this stage). Part of Stage 20: Discovery & Social UI, which is blocked by Stage 19.

## References

- User stories: `user-stories/03-layer-3-web-ui.md` — Stage 20, US-20.7

Issue #217, opened 2026-06-22
US-20.5: Profile Page


frankbria:
## User Story

**As a** listener, **I want** to view a musician's public profile showing their avatar, bio, published songs, playlists, and follower counts, **so that** I can explore their work and follow them.

## Description

Build the public profile page at `/@username` from spec section 32. Displays the creator's identity, published content, and social connections.

## Functional Requirements

- Avatar, display name, and bio prominently displayed
- Style tags rendered as pill badges
- Published songs grid (paginated)
- Playlists section
- Follower and following counts
- Follow/unfollow button (for authenticated users viewing others' profiles)
- Tab navigation between songs, playlists, and about sections
- Profile URL matches the user's handle (`/@username`)

## Acceptance Criteria

- [ ] Profile page loads at `/@username` with correct user data
- [ ] Published songs render in a grid with clip cards
- [ ] Follow button toggles follow state and updates follower count
- [ ] Style tags render as pill badges
- [ ] Playlists section shows the user's public playlists

## Dependencies

**Blocked by:** US-20.4 (prior story in this stage). Part of Stage 20: Discovery & Social UI, which is blocked by Stage 19.

## References

- User stories: `user-stories/03-layer-3-web-ui.md` — Stage 20, US-20.5

frankbria:
## Adapted plan — web-only (Stage-20 precedent)

Following US-20.1–20.4 (#304–#307), US-20.5 ships **web-only** against `lib/*` mock seams rather than CodeRabbit's proposed `Follow` Beanie collection + FastAPI endpoints — the public discovery/social API doesn't exist yet, and each seam getter swaps to a `fetch` when it lands.

**Files (web/):**
- `lib/profiles.ts` (+test) — `PublicProfile` seam: `getProfileByHandle`, `profileClips` (Explore pool), `publicPlaylistsFor` (public `initialPlaylists`).
- `app/[handle]/page.tsx` — root dynamic route; `notFound()` on non-`@`/unknown handle.
- `components/profile/ProfileView.tsx` (+test) — header (avatar glyph, name, bio, style pills, follower/following) + Follow button + Songs/Playlists/About tabs.

**Reuse:** `ExploreClipCard`/`SectionRow`, `initialPlaylists`/`playlistClips`, `ui/{tabs,badge,button}`, `useAuth`.

**Deferred (Known Limitations):** real Follow backend, follower/following lists, cross-session persistence, deep pagination.

PR #284, opened 2026-07-15
feat: US-20.0 public song page and anonymous public read (#250)

### LINKED ISSUE #250: US-20.0: Public Song Page & Anonymous Public Read
## Problem

US-17.6 (#200, PR #249) shipped the inline **Share** control that copies a `{origin}/song/{id}` link for a clip. But a recipient who is **not the clip owner** (or is anonymous) cannot open that link:

- `app/song/[id]/page.tsx` → `SongDetail` uses `useRequireAuth` (redirects unauthenticated visitors to login).
- The metadata fetch `GET /api/v1/clips/{id}` is **owner-scoped even for `is_public` clips** — this is intentional today and pinned by `tests/test_clips_crud_api.py::TestGetClip::test_other_users_clip_returns_404_even_when_public`.

So "publish + share" produces a link that only the owner can open. Public **audio streaming** already works anonymously (US-14.2, `stream_router` / `get_clip_for_audio_access` honor `is_public`), but there is no public **song page** or public metadata read.

## Goal

Make a shared link to a **public** (`is_public == true`) clip openable by anyone — signed-out visitors and non-owner accounts — while keeping private/unlisted clips owner-only.

## Acceptance criteria

- [ ] A signed-out visitor opening `/song/{id}` for a **public** clip sees the song page (title, metadata, playable audio, style tags), not a login redirect or 404.
- [ ] A non-owner authenticated user can open a public clip's page.
- [ ] Private clips still 404 for non-owners/anonymous (no metadata leak).
- [ ] Owner-only actions (publish toggle, delete, edit/remix menu) are hidden or disabled for non-owners on the public page.
- [ ] Existing owner-scoped behavior for the CRUD `GET /clips/{id}` is preserved for the authenticated app; the public read is a distinct, clearly-scoped path.

## Technical notes / suggested approach

- **Backend (layer-2):** add a public metadata read that returns only public clips to anonymous/non-owner callers — e.g. a `stream_router`-style unauthenticated `GET /api/v1/clips/{id}/public` (mirrors how audio streaming already splits onto `stream_router` with `get_current_user_optional`), or relax the existing GET for public clips. Don't weaken the owner-scoped CRUD GET that `test_other_users_clip_returns_404_even_when_public` guards — add a separate path and its own tests.
- **BFF:** a same-origin proxy route for the public read that doesn't require an `Authorization` header.
- **Web:** a public/unauthenticated variant of the song detail page (or make `SongDetail` fall back to the public read + hide owner-only controls when the viewer isn't the owner). Reuse the existing public audio stream for playback.
- Consider whether `unlisted` (deferred from US-17.6) belongs here too: unlisted = openable by link but not listed/discoverable.

## Context

- Surfaced by the cross-family review on PR #249 (codex P1) and captured there as a Known Limitation.
- Deferred from US-17.6 because it needs backend auth changes + a new public page — out of that web-first story's scope.

frankbria:
Closes #250.

Makes a shared `{origin}/song/{id}` link actually open for the person you send it to. Before this, US-17.6's Share button copied a link only its own author could use: the page redirected anonymous visitors to `/login`, and the metadata read was owner-scoped even for `is_public` clips.

This is the read side of link sharing, and the prerequisite the rest of Stage 20 (Explore #213, Search #214, Feed #216, Profile #217) needs — all of them display other users' public content.

## What changed

**Backend — a separate, narrower public read.** New `GET /api/v1/clips/{id}/public` on the existing un-gated `stream_router`, resolving visibility through the pre-existing `get_clip_for_streaming` (US-14.2's rules — anonymous gets an indistinguishable 404 on private-or-unknown; an authenticated non-owner gets 403). The owner-scoped CRUD `GET /clips/{id}` is untouched and still 404s on another user's public clip; `test_other_users_clip_returns_404_even_when_public` passes unchanged.

**The response is redacted for non-owners.** `PublicClipResponse` adds a server-computed `is_owner` (so the owner's `user_id` never reaches the wire) and blanks `workspace_id`, `parent_clip_ids`, `seed` and `inference_steps` unless you own the clip — internal ids that correlate a user's clips, plus the generation recipe. Nothing a visitor sees renders them; the owner keeps all four for the remix/lineage flows.

```
anonymous → workspace_id: null,  seed: null,  parent_clip_ids: [], is_owner: false
owner     → workspace_id: "…00f", seed: 4242, parent_clip_ids: […], is_owner: true
```

**Web — anonymous-capable proxies, and a repaired audio path.** New BFF routes for the public read and for streaming; neither 401s without a token. The stream proxy targets the backend `/stream` rather than the token-gated `/audio`, because an `<audio src>` cannot attach a Bearer header, and it forwards `Range` so seeking works.

This also fixes a real bug on the way past: `clipAudioUrl` pointed at `/api/v1/clips/{id}/audio` — **a path this app never served** (no `/api/v1` route, no rewrites). Clip playback was dead at all four `trackFromClip` call sites. It now resolves same-origin.

**Page.** `SongDetail` drops `useRequireAuth` and reads through the public endpoint. The actions menu and Publish toggle are hidden for non-owners; Like/Dislike/Share stay so anyone can re-share the link.

## Demo

Full evidence in [`docs/demos/us-20.0-public-song-page.md`](docs/demos/us-20.0-public-song-page.md), verified against a live stack, not just tests:

- **Signed-out visitor** opens the public link with **no cookies at all** → page renders, no `/login` redirect. Audio **actually played**: 529,544 bytes fetched anonymously through the stream proxy, and after a real click the transport flipped to Pause with elapsed time advancing 0:00 → 0:03. (The engine builds the element with `new Audio()`, so "an `<audio>` exists" would have proven nothing.)
- **Authenticated non-owner** → page opens, owner controls absent, session confirmed genuinely signed in.
- **Private clip** → "Song not found" for a visitor, asserted to leak neither title nor lyrics.
- **Owner** → same URL, controls *appear*. That's the load-bearing half: they're gated on `is_owner`, not globally absent.

## Deviation from the approved plan, worth a look

The plan's Design Choice 2 said to reuse `ClipResponse` verbatim because it "already redacts ... raw generation params". It does not — it carries `workspace_id`, `seed` and `inference_steps`, and the plan contradicts itself two sentences later. Reusing it would have published the generation recipe and an internal workspace id to anonymous visitors, so the response is redacted instead (@frankbria picked this).

I then extended that redaction to `parent_clip_ids`, which the agreed field list had left visible to non-owners — a review flagged that raw ancestor ObjectIds are the same correlation vector as `workspace_id`, embed a creation timestamp, and would reveal that a *private* parent exists. `get_lineage` already refuses to leak ancestors across users, so this makes the metadata read consistent. Costs the UI nothing (the lineage panel is an owner-scoped authenticated read). **One-line revert if you disagree.**

I also dropped the plan's `get_clip_for_public_read` helper (it would have been an exact alias of `get_clip_for_streaming`) and its per-action `ownerOnly` flag (every action in the menu is an owner action, so the flag would be `true` everywhere; the AC says hide the menu).

## Tests

- Backend: 15 new (`TestGetClipPublic`) — visibility matrix, redaction, rate limiting. 82/82 clips integration green, 1674 unit green, black + ruff clean.
- Web: 14 new BFF route tests, 7 new public-page tests. 1124/1124 green; typecheck, lint, `next build` clean. (`web/` isn't in CI — run locally.)

Mutation-checked the three claims that matter: deleting the redaction, bypassing the visibility helper, and re-introducing the `accessToken` gate in the hook's `loading` formula each fail a test.

## Known limitations

- **Private clips still don't play** via `<audio>` (pre-existing — it was equally broken before, when `clipAudioUrl` pointed at a dead path). Needs cookie-auth on the proxy → #282.
- **The public rate limiter now collapses into one shared bucket** behind the BFF, since all proxied traffic arrives as the Next server's IP. This PR adds the first real traffic to those routes, so it surfaces here. The fix needs a trusted-proxy decision I shouldn't guess at → #283.
- Authenticated non-owners lose Download on a public clip (the whole menu is hidden). Not a regression — the page didn't open for them at all before.
- Cover art still doesn't load (`clipArtworkUrl` has the same dead-path problem; out of scope).
- `unlisted` deferred to US-20.7 (#219) — `Clip` has no such field; that's the write side.

## Review

Cross-family review: opencode/GLM stalled with no output after ~45 min, so `codex review` was used as the fallback. Internal review found the two issues fixed in 3e6672d and the two filed as #282/#283.


frankbria:
## Cross-family review (codex) — findings + triage

opencode/GLM (the usual primary) stalled with zero CPU and no output after ~45 minutes, so this is `codex review --base main` as the documented fallback.

Codex raised **one P1**, quoted in full:

> **[P1] Preserve owner playback for private clips** — `web/lib/clips.ts:51-52`
> When a signed-in owner plays a private clip from the workspace cards, song page, or editor, `trackFromClip` now gives the HTML audio element this `/stream` URL, but the audio element cannot send the in-memory Bearer token. The new proxy therefore forwards no auth, `get_clip_for_streaming` treats the request as anonymous, and private clips return 404, so the app can no longer play the user's default/private clips. Keep a token-capable path for owner/private playback or add an authenticated/signed same-origin stream instead of using the anonymous public stream for every clip.

### Triage: defect is real, "regression" is not — tracked as #282

The underlying limitation is real and I'd already flagged it in the PR body and filed it as #282. But the characterization — *"the app can no longer play"*, *"a core regression for the existing clip library"* — is factually wrong, and it's worth being precise because acting on it as stated would make things worse.

Private clips did not play before this PR either. `main`'s `clipAudioUrl` returned `/api/v1/clips/{id}/audio`, a path this Next app has never served (no `/api/v1` route directory, and `next.config.ts` declares no `rewrites()`). Measured against the running stack:

| Path | Clip | Result |
|---|---|---|
| `/api/v1/clips/{id}/audio` — **what `main` ships** | public | **404** |
| `/api/clips/{id}/stream` — this PR | public | **200** |
| `/api/clips/{id}/stream` — this PR | private | 404 |

So playback was dead for *every* clip on `main`, public and private. This PR moves public 404 → **200** and leaves private at 404 → 404. Strictly better, nothing regressed.

This matters practically: the suggested remedy ("instead of using the anonymous public stream for every clip") points at reverting the `clipAudioUrl` repoint — which would return public playback to 404 and break this story's central acceptance criterion ("playable audio" for a signed-out visitor), in order to fix something that was never working.

Fixing private playback properly needs the BFF to authenticate the request itself from the httpOnly session cookie (an `<audio src>` can't sign anything), and doing that well means deciding how to avoid rotating a single-use refresh token on every Range request. That's its own story, not a rider on this one → **#282**.

### Not flagged by codex, found by the internal review and already fixed here

- **`parent_clip_ids` leaked past the redaction** — raw ancestor ObjectIds are the same correlation vector as `workspace_id` and embed a creation time, so a public clip revealed that a *private* parent existed. Fixed in 3e6672d.
- **416 lost its `Content-Range`** through the stream proxy, stranding a seeking client. Fixed in 3e6672d.
- **The rate limiter now collapses into one shared bucket** behind the BFF (all proxied traffic arrives as the Next server's IP). Needs a trusted-proxy decision → **#283**.

No SSRF or header-injection risk in the new proxies: `BACKEND_URL` is a fixed constant and `id`/`format` only ever land in an `encodeURIComponent`'d path/query segment.

frankbria:
Thanks — #1 is a real bug and is now fixed. Triage of all four, with the two I'm not taking argued rather than waved off.

### 1. Expired token → error screen ✅ fixed in e2d33cb

Correct and reachable: the access token lives in memory with a 30-minute life, so anyone who sits on a tab past expiry and then opens a shared link hits it. Confirmed with a failing test first (the page rendered `song-error` for a public clip), then fixed.

I went with **retry-without-the-token**, not the `401 → notFound` option. Mapping it to `notFound` would tell a viewer with a merely-stale session that a *public* song doesn't exist — worse than the bug, since the whole point of the story is that this link opens for anyone. On a 401 the hook now re-reads anonymously; `is_owner` comes back `false`, which is the honest answer for a viewer we can no longer identify.

### 2. Stale `is_owner` after sign-in — not taking, and the suggested fix would fail CI

Two reasons.

The proposed `setOutcome(null)` inside the effect body violates this repo's lint rule `react-hooks/set-state-in-effect`, which is an **error**, not a warning — state resets have to happen in handlers or async callbacks. It would fail lint.

More importantly I can't find a path that reaches it. The effect gates on `authLoading`, so the initial `null → token` resolve is already covered: no fetch happens until auth settles. Reaching the stale window needs a completed anonymous fetch *followed by* a token appearing — but sign-in here is a full OAuth redirect (`useRequireAuth` → `/login` → callback), so the page remounts with fresh state rather than gaining a token in place. A background refresh is `token → newToken` for the same user, where `is_owner` is unchanged, so nothing visibly flips.

If a genuine sign-in-in-place flow ever lands, the lint-safe fix is to tag the outcome with the token it was fetched under (`outcome.token !== accessToken` → treat as loading), mirroring the existing `id`-tagging idiom. Happy to add it then; today it'd be dead code guarding an unreachable state.

### 3. Stream proxy forwards only `format` — intentional, keeping the allowlist

This is deliberate, and it matches the sibling `/api/clips/[id]/audio` route, which also forwards `format` alone. The backend's `stream_clip_audio` signature accepts exactly one query param:

```python
format: Literal["wav", "mp3"] | None = Query(default=None, ...)
```

So the hypothetical `quality`/`bitrate` wouldn't work if forwarded — FastAPI would ignore them at the far end instead of the near end. Forwarding `searchParams.toString()` wholesale would let a caller push arbitrary attacker-controlled query content at the backend on an **unauthenticated** route, which is exactly the kind of thing worth *not* doing by default on the anonymous surface. An allowlist is the safer default and YAGNI until a second param actually exists.

### 4. Pydantic field widening — accepting the risk, but the suggested fixes don't address it

The LSP point is fair in the abstract: `str | None` genuinely isn't a subtype of `str`. But this repo runs **no Python type checker** — CI is `black --check` → `ruff check` → `pytest` (`.github/workflows/ci.yml`), ruff is lint-only (`E7`, `E9`, `F`, `I`), and there's no mypy/pyright in `pyproject.toml` or `.pre-commit-config.yaml`. So there's nothing to go silent *until*.

Neither suggested remedy would fix the stated concern anyway: `revalidate_instances="always"` changes validation of model instances and has nothing to do with static analysis, and a bare `Field()` annotation doesn't make an incompatible override compatible. The only real fixes are to stop subclassing (duplicating 17 fields) or to widen the parent (which would weaken every owner-scoped endpoint that relies on `workspace_id` always being set). Runtime behaviour is pinned by 15 tests including both redaction directions. Leaving as-is, deliberately.

---

Also worth noting two things this review didn't reach, both already handled: `parent_clip_ids` was leaking past the redaction (raw ancestor ObjectIds embed a creation time and would reveal a *private* parent's existence) — fixed in 3e6672d; and the rate limiter now collapses into a single shared bucket behind the BFF — filed as #283.

frankbria:
@claude The GLM precision review's single finding — the 401-from-expired-token path in `use-public-clip.ts` — is **fixed in e2d33cb**, landed before this comment arrived. It converged on the same fix independently ("fall back to an anonymous read on 401"), which is what I implemented, driven by a test that first reproduced `song-error` on a public clip.

Your added detail that it **recurs for the rest of the session on client-side nav** (the token is minted once on mount, never proactively refreshed, and `decodeAccessToken` ignores `exp`) makes it worse than I'd scoped it — a single expiry would have poisoned every shared link the viewer opened until a full reload. Worth recording as the reason the retry belongs in the hook rather than at a call site.

Two notes on the verification list, since it's useful that we agree: the `parent_clip_ids` blanking you verified was itself added mid-review (3e6672d) after the internal pass caught it leaking — raw ancestor ObjectIds embed a creation time and would reveal a *private* parent's existence. And the `workspace_id: string | null` widening is tolerated by `ClipMultiSelector` deliberately: it skips the fetch entirely when null (`enabled: workspaceId !== null`) rather than querying an empty workspace.

Known-and-tracked, not fixed here: private-clip playback (#282 — pre-existing; `main`'s `clipAudioUrl` pointed at a path that 404s, measured) and the BFF rate-limit bucket collapse (#283).

frankbria:
Picking up **finding #5** (`usePublicClip` / `useClip` duplication), which I missed when triaging 1–4 — I'd truncated the comment before it. Not taking it, but the reasoning turned up a real pre-existing bug, now filed.

### The two hooks want *opposite* behaviour on the exact case that motivates sharing

The argument for extracting `useFetchedClip(url, headers)` is "when a fix is needed (e.g. the 401 case above), it has to be applied to both". I checked whether that's true, and it's the reverse — applying the 401 fix to `useClip` would be a bug:

- **`usePublicClip`** backs `/song/{id}`, which is public. An expired token there should degrade to the anonymous read, because the endpoint serves anonymous callers and the clip may well be public. That's what e2d33cb does.
- **`useClip`** backs `/studio` and `/release` only (`app/studio/page.tsx:44`, `app/release/page.tsx:99`) — both owner-only tools sitting behind `useRequireAuth`. It reads the **owner-scoped** `GET /clips/{id}`, which is gated by `Depends(get_current_user)` at the router. A headerless retry there can only ever 401 again. Degrading to anonymous is not a fallback, it's a pointless second round-trip.

So the shared primitive would have to parameterise: whether to gate on `accessToken`, which statuses collapse to `notFound` (404 vs 404+403), whether to retry on 401, and the `loading` formula. That's four knobs over two callers, where the knobs *are* the logic — the machinery they share (a tagged union, an `active` flag) is the boring part, and the part that differs is the part that matters. Deferring until there's a third caller to triangulate from.

### The real bug hiding behind it → #285

Tracing "would `useClip` need the same fix" surfaced something worse, and pre-existing:

- `AuthProvider` calls `/api/auth/refresh` **only on mount** (`contexts/auth-context.tsx:40`) — no interval, no refresh-ahead.
- `decodeAccessToken` never inspects `exp` (`lib/auth.ts`), so `isAuthenticated` stays `true` on a dead token.
- Therefore `useRequireAuth` does **not** redirect once the token ages out.

An App Router client nav doesn't remount the provider, so any tab left open past the access-token lifetime keeps a stale token and `useClip` 401s → `/studio` and `/release` render *"Couldn't load this song"* until a hard reload, with no redirect to re-auth. Not caused by this PR and not fixable here (it wants a refresh-ahead or `exp`-aware `useRequireAuth`, and the right recovery for an owner-only page is re-auth, not degradation) → **#285**.

This is also the second finding that traces back to the same root as #1 — the token is minted once and never refreshed — which is worth fixing at the source rather than per-hook.

frankbria:
@claude Good catch on the `isOwner` default — **taken, fixed in 272378a.**

You're right that it's the wrong safe direction. My reasoning for `true` was purely "don't make existing callers pass it," which is a convenience argument applied to an ownership gate — the wrong trade once this header renders on a public page. Failing closed costs one explicit prop at the one call site that needs it.

Your read of the blast radius matches mine: no live bug, since `SongDetail` is the only production caller and always passes it explicitly. But the second half of your finding was the more valuable part — the unit tests leaned on the unsafe default, so `SongHeader.test.tsx` only ever exercised the owner path and the non-owner rendering was covered *indirectly*, via `SongDetail`'s integration tests. A component-level regression would have slipped past its own suite.

What landed:
- Default flipped to `false`.
- The two publish tests now pass `isOwner: true` explicitly, so they state what they're testing instead of inheriting it.
- Added the two missing cases at this level: `isOwner: false` hides the toggle while keeping Like/Dislike/Share, and an **omitted** prop hides it too.

I mutation-checked the last one rather than trusting it — flipping the default back to `true` fails `fails closed: an omitted isOwner hides the publish toggle`, so the bias is pinned now rather than incidental.

Web suite 1127/1127, typecheck and lint clean. (`web/` isn't in CI, so those are local.)

Also picked up your **finding #5** (the `useClip` duplication) in a separate comment — not taking the shared primitive, because the two hooks want opposite 401 semantics, but tracing it surfaced a real pre-existing bug: the access token is never refreshed and `isAuthenticated` ignores `exp`, so `/studio` and `/release` break after expiry with no re-auth → #285.

frankbria:
## Final Triage Summary (PR #284)

Cutoff: `2026-07-15T05:36:57Z` (last code push, `272378a`)
New findings since cutoff: **0** — the final review pass (05:41:22Z) raised no new items and confirmed the prior triage stands.

### Fixed across the review cycle

| Finding | Source | Resolution |
|---|---|---|
| `parent_clip_ids` leaked past the redaction — raw ancestor ObjectIds are the same correlation vector as `workspace_id` and embed a creation time, revealing a *private* parent's existence | internal review | Fixed in `3e6672d` |
| 416 dropped `Content-Range` through the stream proxy, stranding a seeking client | internal review | Fixed in `3e6672d` |
| Expired token → 401 → error screen on a **public** clip | general review + GLM precision review (found independently) | Fixed in `e2d33cb` — retries anonymously |
| `SongHeader` defaulted `isOwner = true` — wrong safe direction on a public page; unit tests leaned on the unsafe default, leaving the non-owner path covered only indirectly | general review | Fixed in `272378a` — fails closed, mutation-checked |

- **[general review] Map 401 → `notFound`** — rejected: would hide a *public* song from a viewer whose token merely expired, which is worse than the bug. Retried anonymously instead.
- **[codex] "Private-clip playback is a core regression"** — inaccurate. Measured: the path `main` ships (`/api/v1/clips/{id}/audio`) returns **404** — playback was dead for *every* clip before this PR. Public now 200s, private stays 404. Its suggested remedy (revert the `clipAudioUrl` repoint) would re-break AC1. Real limitation tracked → #282.
- **[general review] Stream proxy forwards only `format`** — intentional. Backend `stream_clip_audio` accepts exactly one query param; forwarding `searchParams.toString()` wholesale would push unvetted content at an unauthenticated route. Matches the sibling `/audio` route.
- **[general review] Pydantic `workspace_id: str | None` widening** — accepted risk. No Python type checker in CI (`black` → `ruff` → `pytest`; ruff is lint-only). Neither suggested remedy (`revalidate_instances`, bare `Field()`) addresses a static-analysis complaint. Runtime pinned by 15 integration tests covering both redaction directions.
- **[general review] Stale `is_owner` after sign-in-in-place** — unreachable: sign-in is a full OAuth redirect, so the page remounts; a background refresh is `token → newToken` for the same user. The suggested `setOutcome(null)` in an effect body would also fail this repo's `react-hooks/set-state-in-effect` lint error.
- **[general review] `usePublicClip`/`useClip` duplication** — the two hooks need **opposite** 401 semantics (public → degrade to anonymous; owner-scoped → auth-gated endpoint, a headerless retry can only 401 again). A shared primitive would have to parameterise the logic that differs. Tracing it surfaced a real pre-existing bug → **#285**.
- **[deslop] `ponytail:` marker in `web/lib/clips.ts`** — inaccurate: it's an established repo convention with 10+ existing uses, including `SongMetadata.tsx` and `SongLyrics.tsx` in the same directory.

### Deferred to filed issues (out of scope, disclosed in PR body)

- **#282** — private-clip playback (`<audio src>` can't attach a Bearer token; pre-existing)
- **#283** — BFF rate-limit bucket collapse (needs a trusted-proxy decision)
- **#285** — access token never refreshed; `isAuthenticated` ignores `exp`

### Demo re-run

The prior demo was captured at `e8e2b7a`, three commits before HEAD — including `272378a`, which changed the very ownership-gate default AC4 rests on. Re-run in full against `5633c68`; all five criteria verified with outcome evidence. The public clip now has a **private parent**, so the ancestry redaction is exercised against real lineage.

PR #303, opened 2026-07-19
fix(web): refresh access token ahead of expiry so owner pages survive it (#285)

### LINKED ISSUE #285: P1.3 Access token is never refreshed: owner-only pages break after expiry with no re-auth
## Problem

The access token is minted once per page load and **never refreshed**, and the client never checks its expiry. Once it ages out, owner-only pages break with a generic error and the app does not send the user to re-authenticate.

Three facts combine:

1. `AuthProvider` calls `POST /api/auth/refresh` **only on mount** (`web/contexts/auth-context.tsx:40`) — there is no interval, no refresh-ahead, and no retry on 401.
2. `decodeAccessToken` never inspects the `exp` claim (`web/lib/auth.ts`), so `isAuthenticated` stays `true` for an expired token.
3. Because `isAuthenticated` stays `true`, `useRequireAuth` (`web/hooks/use-require-auth.ts`) does **not** redirect to `/login`.

Next's App Router does not remount the provider on client-side navigation, so a tab left open past `access_token_expire_minutes` keeps a dead token for the rest of the session.

## Impact

`useClip` (`web/hooks/use-clip.ts`) sends the stale token to the owner-scoped `GET /clips/{id}`, gets a 401, and falls through `if (!res.ok) throw` to its `error` outcome. Its two callers are owner-only tools:

- `web/app/studio/page.tsx:44` → `/studio`
- `web/app/release/page.tsx:99` → `/release`

Both render **"Couldn't load this song"** until a full page reload, with no redirect to re-auth. Any other authenticated fetch (`use-clips`, `use-lineage`, `use-subscription-tier`, the edit/mutation flows) degrades the same way once the token dies.

## Not a regression

Pre-existing; predates #250/#284. Surfaced while triaging a review comment there ("would `useClip` need the same 401 fix as `usePublicClip`?"). The public song page has since been made resilient — `usePublicClip` retries anonymously on 401 (e2d33cb) — but that fix is **only correct for the public read**: it degrades to an anonymous view, which an owner-only page must not do.

## Suggested approach

Fix at the source rather than per-hook (this root cause has now produced two separate findings):

- Make `AuthProvider` refresh-ahead: decode `exp` and schedule a refresh before expiry, and/or refresh on a 401 from any authenticated fetch and replay the request once.
- Make `decodeAccessToken` / `isAuthenticated` `exp`-aware so `useRequireAuth` actually redirects on a dead session instead of letting pages render into a broken state.
- Mind the rotating single-use refresh token — concurrent refreshes must not race each other into a revoked token.
- Owner-only pages should re-auth, not degrade. Do **not** apply `usePublicClip`'s anonymous-retry to `useClip`: the endpoint it reads is auth-gated at the router, so a headerless retry can only 401 again.

## Acceptance criteria

- [ ] A tab left open past the access-token lifetime can still load `/studio` and `/release` without a manual reload.
- [ ] An unrecoverable session (refresh token revoked/expired) redirects to `/login` rather than rendering "Couldn't load this song".
- [ ] Concurrent authenticated requests around an expiry boundary don't race the rotating refresh token into revocation.
- [ ] The public song page still renders for a signed-out visitor (no regression to US-20.0 / #284).

frankbria:
Fixes #285.

## Problem
The access token was minted once on `AuthProvider` mount and **never refreshed**, and the client never inspected `exp`. A tab left open past `access_token_expire_minutes` kept a dead token for the rest of the session — App Router doesn't remount the provider on client-side nav. Owner-only pages (`/studio`, `/release` via `useClip`) then rendered **"Couldn't load this song"** on the 401, with no redirect to re-auth.

## Fix (at the source, in `AuthProvider`)
- **Single-flight `refresh()`** shared by mount, the scheduler, and the visibility listener — concurrent triggers share one request, so they can't race the backend's **single-use** refresh token into revocation.
- **Refresh-ahead**: schedules a renewal ~60s before the JWT `exp`; each success chains the next.
- **Refresh-on-visibility**: a backgrounded tab throttles timers, so when it returns and its token is near/past expiry it renews immediately.
- **Unrecoverable vs transient**: a `401/403` clears the token → `isAuthenticated` flips false → `useRequireAuth` redirects to `/login`. A transient `5xx/429/network` error keeps the current token and re-arms a short retry (signing a valid user out on a backend blip would be worse than the blip).
- `decodeTokenExp()` helper added to `lib/auth.ts`.

**Why no per-hook changes:** every authed hook already lists `accessToken` in its effect deps, so a rotated token automatically re-fires their fetches and recovers the page. The fix collapses to one provider instead of ~10 hooks. `usePublicClip` is untouched.

## Acceptance criteria
- [x] A tab left open past the access-token lifetime can still load `/studio` and `/release` without a manual reload — refresh-ahead + visibility keep the token live; a rotated token re-fires `useClip`.
- [x] An unrecoverable session (refresh revoked → 401) redirects to `/login` rather than rendering "Couldn't load this song".
- [x] Concurrent authenticated requests around an expiry boundary don't race the rotating refresh token — single-flight `refresh()`.
- [x] The public song page still renders for a signed-out visitor — `usePublicClip` path unchanged.

## Tests
- `lib/auth.test.ts`: `decodeTokenExp` (exp→ms, missing/malformed → null).
- `contexts/auth-context.test.tsx`: refresh-ahead applies the rotated token; a scheduled refresh finding the session revoked signs out; a transient 5xx keeps the session; concurrent triggers coalesce to one request.
- Full web suite: **1144 passing**; `tsc` + `eslint` clean; production `build` green.

## Review
Cross-family review via `codex` (GLM/opencode timed out). It flagged the transient-5xx-signs-out case (a MAJOR) — **fixed** in this branch (only 401/403 clears the token; transient errors retry). Its MINOR (setState-after-unmount) is not applicable: React 19 no longer warns and `AuthProvider` is the app-root provider.

## Known limitations
- **Cross-tab single-use race** (pre-existing, noted in `api/auth/refresh/route.ts`): two tabs refreshing the *same* pre-rotation cookie in the same instant — the second 401s. In practice the shared httpOnly cookie is already rotated by the time the second tab reads it. Per-tab single-flight fully covers the within-tab case.
- **No per-hook 401 replay**: recovery relies on the provider keeping the token fresh (it re-fires consumers via their `accessToken` dep). A refresh is not triggered by arbitrary user interaction — only by the scheduler, the retry chain, or `visibilitychange`.
- CI is Python-only; the web gates above were run locally.

### DIFF
--- a/web/contexts/auth-context.test.tsx
+++ b/web/contexts/auth-context.test.tsx
@@ -220,13 +220,15 @@ describe("AuthProvider", () => {
       releaseRefresh = resolve
     })
     let call = 0
-    const fetchMock = vi.fn(async (url: string) => {
+    let refreshSignal: AbortSignal | undefined
+    const fetchMock = vi.fn(async (url: string, init?: RequestInit) => {
       if (url === "/api/auth/logout") return ok({})
       if (url !== "/api/auth/refresh") return unauthorized
       call += 1
       // Mount restores an (already-expired) token so the visibility listener
       // will start a background refresh; that second refresh hangs on the gate.
       if (call === 1) return ok({ access_token: tokenExpiring(-10) })
+      refreshSignal = init?.signal ?? undefined
       await gate
       return ok({ access_token: tokenExpiring(15 * 60) })
     })
@@ -250,9 +252,12 @@ describe("AuthProvider", () => {
     await waitFor(() =>
       expect(screen.getByTestId("state")).toHaveTextContent("out")
     )
+    // logout aborts the in-flight refresh so the browser never applies its
+    // rotated Set-Cookie (the HTTP/2-reorder cookie-resurrection vector).
+    expect(refreshSignal?.aborted).toBe(true)
 
-    // The in-flight refresh now resolves 2xx — it must NOT flip the session
-    // back to authenticated.
+    // Even if the (aborted) refresh still resolves 2xx, it must NOT flip the
+    // in-memory session back to authenticated.
     await act(async () => {
       releaseRefresh()
       await gate
--- a/web/contexts/auth-context.test.tsx
+++ b/web/contexts/auth-context.test.tsx
@@ -193,8 +193,11 @@ describe("AuthProvider", () => {
           <Probe />
         </AuthProvider>
       )
+      // Advance 0ms: flush the mount refresh without firing the expired token's
+      // scheduled retry (armed at REFRESH_RETRY_MS), so `call` reflects only the
+      // mount request here.
       await act(async () => {
-        await vi.runOnlyPendingTimersAsync()
+        await vi.advanceTimersByTimeAsync(0)
       })
       expect(call).toBe(1)
       expect(screen.getByTestId("state")).toHaveTextContent("in:")
--- a/web/contexts/auth-context.test.tsx
+++ b/web/contexts/auth-context.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor } from "@testing-library/react"
+import { act, render, screen, waitFor } from "@testing-library/react"
 import userEvent from "@testing-library/user-event"
 import { afterEach, describe, expect, it, vi } from "vitest"
 
@@ -8,6 +8,16 @@ import { useAuth } from "@/hooks/use-auth"
 const accessToken = `h.${Buffer.from(
   JSON.stringify({ sub: "u1", email: "musician@example.com" })
 ).toString("base64url")}.s`
+
+/** A token that carries an `exp` claim `secondsFromNow` in the future. */
+const tokenExpiring = (secondsFromNow: number) =>
+  `h.${Buffer.from(
+    JSON.stringify({
+      sub: "u1",
+      email: "musician@example.com",
+      exp: Math.floor(Date.now() / 1000) + secondsFromNow,
+    })
+  ).toString("base64url")}.s`
 
 function Probe() {
   const { isLoading, isAuthenticated, user, logout } = useAuth()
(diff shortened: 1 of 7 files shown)


frankbria:
### Review triage (GLM + Codex + Claude) — addressed in 38df6cb

| Finding | Severity | Action |
| --- | --- | --- |
| Retry `setTimeout` orphans on teardown mid-refresh → loops every 15s | confirmed defect (GLM + Claude) | **Fixed** — `active` guard in the scheduler effect cancels a pending retry on cleanup |
| Transient 5xx signs a valid user out | MAJOR (Codex) | **Fixed** in the first commit — only 401/403 clears the token; 5xx/429 retries |
| Expired-on-arrival token never scheduled (foreground tab that never backgrounds) | minor (Claude #1) | **Fixed** — schedules at the retry cadence instead of skipping; retry cadence also caps any busy-loop |
| `useClipAudio` re-decodes whole clip on every rotation | minor (Claude #2) | **Fixed** — keyed on token *presence* (token via a ref); +test |
| `model-selection` re-fetches `/me` on rotation; comment now false | minor (Claude #3) | **Fixed** — same presence-gate; comment corrected |
| `decodeTokenExp` duplicates `decodeAccessToken` decode | minor (Claude #4) | **Fixed** — extracted shared `decodeJwtPayload()` |
| setState-after-unmount | MINOR (Codex) | **N/A** — React 19 doesn't warn; `AuthProvider` is app-root, never unmounts mid-session |
| visibility listener resubscribes per rotation; split `REFRESH_SKEW_MS` into two constants | minor (Claude #5/#6) | **Won't fix** — resubscribe cost is negligible; the two uses are genuinely one safety-window knob, splitting is premature |

Full web suite **1147 passing**; `tsc` + `eslint` + `next build` green.

frankbria:
### Re-review MAJOR addressed in `9eb5016`

**Logout-resurrection race** (new finding, confirmed by GLM precision + Claude): an in-flight `refresh()` resolving after `logout()` could re-authenticate a just-ended session (AuthProvider is root-mounted, never unmounts on the nav to /login). **Fixed** with a `sessionEpoch` ref that `logout()` bumps; `refresh()` drops its result if the epoch changed while in flight. Regression test added (`does not let an in-flight refresh resurrect the session after logout`).

The two lower-severity notes (extract a shared `useLatestRef` hook; expose `getAccessToken()` from the provider) are valid follow-up refactors, not defects — deferred to keep this PR scoped to the bug.

Full web suite **1148 passing**; tsc + eslint + build green.

frankbria:
### Re-review MAJOR (cookie vector) addressed in `de8a03f`

Correct catch: the epoch guard closed only the in-memory vector; the in-flight refresh response still carried a rotated `Set-Cookie` that HTTP/2 reordering could land *after* logout's cookie-clear. **Fixed** — `logout()` now `abort()`s the in-flight refresh (via an `AbortController` held in a ref) before clearing the token and hitting `/logout`, so the browser never applies the aborted response's cookies; `/logout`'s clear is the last write. Epoch guard kept for the response-received-but-handler-pending case (defense in depth). Test asserts the in-flight refresh's signal is aborted by logout.

1148 web tests · tsc · eslint · build all green.

PR #364, opened 2026-07-29
feat(video): US-22.3 Video Rendering Progress and Delivery (#313)

### LINKED ISSUE #313: US-22.3: Video Rendering Progress and Delivery
## User Story

**As a** musician, **I want** to see real-time progress while my video renders and receive it when complete, **so that** I know the status and can plan accordingly.

## Description

Video rendering is a long-running job (minutes to tens of minutes depending on duration and resolution). The progress view keeps the musician informed and delivers the final video with download and publish options.

## Functional Requirements

- Progress view replaces the generation form after submission
- Progress bar with percentage, estimated time remaining, and current phase (analyzing audio, generating scenes, encoding)
- In-browser notification when rendering completes (even if the user navigated away)
- Video player with the rendered result, full playback controls
- Download button: MP4 (H.264 by default, H.265 for 4K)
- "Publish to Platform" button — makes the video publicly visible on the song detail page
- "Export for Social" — provides optimized versions for YouTube (16:9), TikTok/Reels (9:16), and Instagram (1:1)

## Acceptance Criteria

- [ ] Progress bar updates at least every 5 seconds during rendering
- [ ] Notification fires on completion if user navigated away
- [ ] Completed video is playable in-browser
- [ ] Download produces a valid MP4 with correct resolution and audio
- [ ] Publish makes the video visible on the song detail page

## Dependencies

**Blocked by:** US-22.2 (prior story in this stage). Part of Stage 22: Music Video Generator, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 22, US-22.3

frankbria:
## Summary
Implements #313 (US-22.3): the **delivery** half of the music-video feature, on top of US-22.1 (generate/status backend) and US-22.2 (creation page + progress view).

- **Backend** (`src/acemusic/api`): `Video.published` flag; service helpers (`get_video_by_id`, `get_owned_video`, `publish_video`, `get_published_video_for_clip`); `POST /videos/{id}/publish` (owner-only) plus a public sub-router — `GET /videos/{id}` (metadata), `GET /videos/{id}/stream` (Range + `?download`), `GET /videos/for-clip/{clip_id}`. Streaming/visibility mirror the clip-audio pattern (owner always; a published video is viewable to whoever can view its clip; a private/unpublished video is an indistinguishable 404).
- **Web BFF proxies**: stream (Range + httpOnly-cookie fallback so a bare `<video src>`/download link can authenticate private media, per #282), publish, detail, for-clip.
- **Delivery client + UI**: `lib/video` gains `videoStreamUrl`/`publishVideo`/`fetchVideoDetail`/`fetchPublishedVideoForClip`; a `VideoDelivery` view (in-browser player + MP4 download + Publish) replaces the completed-render placeholder in `VideoProgress`.
- **Cross-navigation completion notification (AC2)**: a layout-level `VideoJobsProvider` keeps polling in-flight renders across client-side navigation and fires the new `video_complete` notification when one finishes — so it still arrives if the user left the create page.
- **Song page (AC5)**: a `SongVideo` section renders the clip's published video on the public song detail page.

## Acceptance Criteria
- [x] Progress bar updates ≥ every 5s during rendering — existing hook polls every 2.5s (US-22.2)
- [x] Notification fires on completion if user navigated away — `VideoJobsProvider` (in-memory, survives SPA navigation)
- [x] Completed video is playable in-browser — `<video>` in `VideoDelivery`, streamed via the Range-aware proxy
- [x] Download produces a valid MP4 with correct resolution + audio — Download button → `?download=1` (Content-Disposition), serves the stored render bytes
- [x] Publish makes the video visible on the song detail page — publish flag → `SongVideo` on `/song/{id}`

## Test Plan
- [x] Unit/integration tests written (TDD)
- [x] All tests passing — backend `pytest tests/test_video_api.py` (46 integration + auth-gate); web `vitest run` (1544 tests, 199 files)
- [x] Diff coverage on new lines — manual check (coverage-v8 not installed): every new function/branch exercised (video modules 94–100% on the focused pytest --cov run)
- [x] Lint clean — ruff + black (Python), eslint + tsc + `next build` (web)
- [x] Internal review (advisory) completed
- [x] Cross-family review: opencode (zai/glm) — see PR comment
- [x] Test mutation sanity check (Phase 7d): the completion-notify branch and `publish_video` both fail their tests under mutation

- **"Export for Social" (YouTube 16:9 / TikTok-Reels 9:16 / Instagram 1:1)** — deferred. It is in the story's functional-requirements list but **not** in any acceptance criterion, and aspect ratio is already chosen at render time (there is no post-render re-frame/transcode pipeline). Follow-up issue can add a re-export endpoint if desired.
- **Codec selection (H.265 for 4K)** — the delivery endpoints serve the stored render bytes as-is; no on-the-fly transcode. Whatever codec the provider produced is what downloads.
- **Notification durability** — `VideoJobsProvider` is in-memory: the completion notification survives client-side navigation but not a full page reload (a reload drops the in-flight watch list). Persist to `localStorage` if reload-durable notifications become a requirement.
- **No real video provider in this environment** — `POST /videos/generate` 503s without provider credentials, so end-to-end renders don't advance here; the delivery endpoints were demoed against a seeded `Video` + MP4.
- **Stream buffers the whole file into memory** — `GET /videos/{id}/stream` loads the full object via `storage.download()` before slicing a Range, exactly as the existing clip-audio streaming does. Fine for the current audio-sized objects but a memory concern for large (100 MB+) 1080p/4K videos under concurrency. Proper fix is a streaming/partial-read `StorageBackend` API (S3 range GET + local seek) that also benefits clip audio — out of scope here, tracked as a follow-up. Flagged by cross-family review (opencode).

## Implementation Notes
- Self-authored plan (issue had no plan comment); no architectural fork required approval.
- The delivery proxy routes reuse the existing `[jobId]` dynamic segment (Next forbids a second slug name at the same path level); the URL segment carries a video id for those routes.

Closes #313


frankbria:
## Cross-family review — opencode (zai/glm)

## Review: US-22.3 Video Rendering Progress and Delivery

**Major** — `src/acemusic/api/routers/videos.py:341`: `storage.download(video.storage_path)` buffers the entire video into RAM before serving even a byte-range slice; a 1080p render can exceed 100 MB, so a handful of concurrent viewers risk OOM or event-loop stalls that kill all requests on the process.

**Major** — `web/contexts/video-jobs-context.tsx:89`: `notify` sits in the `useEffect` dependency array but its referential stability depends entirely on `useNotify()` returning a memoized function — if it doesn't, the 3 s poll timer resets on every layout render and completion notifications silently never fire, violating AC2. Wrap `notify` in a ref (same pattern already used for `tokenRef`/`activeRef`) or remove it from deps.

**Suggestion** — `src/acemusic/api/services/video.py:82-88`: `publish_video` does a non-atomic read-then-`save()` instead of `find_one_and_update({"_id": oid, "user_id": uid, "published": False}, {"$set": {"published": True}})`. Harmless today (idempotent field), but the pattern will bite when a non-idempotent mutation lands on this code path.

**Suggestion** — `src/acemusic/api/routers/videos.py:355-377`: When `parse_range_header_multi` returns `None` (malformed `Range` header), the endpoint silently serves a full 200 body — RFC 7233 §4.4 says the server SHOULD ignore the header, which this does, but logging the malformed input would help diagnose broken player implementations.

**Suggestion** — `web/components/video/VideoDelivery.tsx:78`: The `<a>` tag carries both `href` with `?download=1` (backend sets `Content-Disposition: attachment`) and a bare `download` attribute — the HTML attribute is redundant and, if the BFF proxy ever streams via a different origin, would be silently ignored anyway.

---

APPROVE

---

### Author triage
- **Major (RAM buffering, videos.py stream)** — valid; documented under *Known Limitations*. Mirrors the existing clip-audio streaming (`storage.download()` full-read then slice); a proper fix needs a streaming/partial-read `StorageBackend` API that also benefits clip audio, so it's tracked as a follow-up rather than fixed here.
- **Major (notify referential stability, video-jobs-context)** — `notify` is in fact memoized today (`useCallback([])` in the provider + stable `noop` fallback), so it wasn't broken. But the ref fix is a real robustness/consistency win — applied in `8b862e9`: `notify` now lives in a ref and the poll effect depends on nothing, so its timer mounts once and never restarts.
- **Suggestion (atomic publish)** — skipped: `publish_video` is idempotent and owner-scoped, so the read-then-save race is harmless (both writers land `published=True`).
- **Suggestion (log malformed Range)** — skipped: serving a full 200 on an unparseable Range matches RFC 7233 §4.4 and the existing clip-audio behavior.
- **Suggestion (redundant `download` attr)** — skipped: the same-origin `download` attribute is harmless and adds resilience if a proxy ever drops `Content-Disposition`.

frankbria:
## Demo verification (Phase 11) — outcome evidence

Delivery ACs exercised end-to-end against the real backend (FastAPI + MongoDB + local storage) serving a genuinely playable **720p H.264 + AAC MP4** (ffmpeg-generated, ffprobe-verified). AC2's client behavior is proven by the vitest context suite.

| Criterion | Action | Outcome evidence | Status |
|---|---|---|---|
| AC1 progress updates ≥ every 5s | `GET /videos/{job}/status` across a job's lifecycle | Reported `rendering 20% (eta 90s)` → `rendering 55% (eta 40s)` → `encoding 92% (eta 8s)` → `complete 100% video_id=…`; web hook polls every 2.5s (< 5s) | VERIFIED |
| AC2 notify if navigated away | app-level `VideoJobsProvider` observes a job's `status=complete` | `contexts/video-jobs-context.test.tsx`: the `notify` mock **receives** `{type:"video_complete", href:"/video/{songId}"}` from the provider's poll loop (mounted in the root layout → survives route changes); backend status→complete transition (above) is the data it consumes | VERIFIED |
| AC3 playable in-browser | `GET /videos/{id}/stream` | 200, `content-type: video/mp4`, `accept-ranges: bytes`; **served bytes ffprobe → h264 video + aac audio** (exact render); Range `bytes=0-99` → 206 partial (seekable) | VERIFIED |
| AC4 download valid MP4 (resolution + audio) | `GET /videos/{id}/stream?download=1` | `Content-Disposition: attachment; filename="video-{id}.mp4"`; **downloaded file ffprobe → width=1280 height=720 h264 + aac audio** | VERIFIED |
| AC5 publish → visible on song page | `for-clip` before/after `POST /videos/{id}/publish` | Before: `for-clip` → **404** (no video section). After publish: `for-clip` (anonymous) → **200** returning the video; anonymous `stream` → 200 with the full MP4 (a signed-out visitor can watch it on the public song page) | VERIFIED |

Demo artifacts (MP4 + script) live in the session scratchpad — not committed (outside the repo tree).

**Note:** no real video provider is configured in this environment (`POST /videos/generate` 503s), so the render pipeline itself doesn't advance here; the delivery half was demoed against a seeded `Video` + real MP4, which is exactly what the endpoints serve in production.

frankbria:
## Triage of @claude review notes (fixed in 0d0b76d)

Both notes verified and addressed:

- **Minor — stream proxy `download` truthiness** — fixed. The public BFF stream route now forwards download only for `?download=1` (`=== "1"`), so `?download=0`/`false` no longer force `Content-Disposition: attachment`. Added a proxy test for the falsy case.
- **Test coverage — 416 unsatisfiable range** — added `TestVideoStream::test_unsatisfiable_range_returns_416`, asserting 416 + `Content-Range: bytes */{len}`, mirroring the clip-audio endpoint's regression test.

Thanks for the thorough pass — the RAM-buffering and non-atomic-publish points remain tracked (#365) / justified as noted.

frankbria:
## Triage of @claude re-review (fixed in 765447e)

Thanks for the second pass. Both re-reviews confirm the earlier fixes landed. Dispositions on the new observations:

- **Minor — metadata BFF proxies used bare `fetch`** (`videos/[jobId]`, `videos/for-clip/[clipId]`) — **fixed**. Both now use `fetchWithTimeout` + `clientIpHeaders`, matching `stream/route.ts` and `clips/[id]/public`: fail closed (502) on a stalled backend, and forward the real client IP for backend rate limiting (#283).
- **Minor — `Cache-Control` can outlive a clip visibility flip** — valid; **folded into #365** (it pairs with the streaming `StorageBackend` rework: shorter max-age / revalidation on access-controlled media). Not changed here to stay consistent with the clip-audio pattern this mirrors (not a regression).
- **Nit — `ponytail:` comment reads as a stray word** — left as-is: `ponytail:` is an existing deliberate-simplification marker in this repo (e.g. `src/acemusic/api/models/clip.py:57`), so it's consistent with the codebase convention.

No blocking findings outstanding.

PR #366, opened 2026-07-29
perf(storage): streaming/partial-read media delivery (#365)

### LINKED ISSUE #365: P2.9: Streaming/partial-read StorageBackend API for media delivery (video + audio)
## Context
Surfaced by cross-family review on PR #364 (US-22.3 video delivery).

`GET /api/v1/videos/{id}/stream` (and the pre-existing `GET /clips/{id}/audio` + `/clips/{id}/stream`) load the **entire** object into memory via `storage.download(path) -> bytes` and then slice it for a Range request. This is fine for audio-sized clips but a real memory/latency concern for large video renders (a 1080p/4K MP4 can exceed 100 MB); a handful of concurrent viewers could OOM or stall the event loop.

## Problem
`StorageBackend` (`src/acemusic/storage.py`) only exposes `download() -> bytes`, so no endpoint can serve a byte range without buffering the whole object first.

## Proposed work
- Add a streaming / partial-read method to `StorageBackend`, e.g. `open_range(path, start, end) -> Iterator[bytes]`:
  - **S3**: `get_object(Range="bytes=start-end")`
  - **Local**: `open(...).seek(start)` + chunked read
- Have the video stream endpoint (and, ideally, the clip audio endpoints) use it with FastAPI `StreamingResponse` so a Range request never buffers more than a chunk.
- Keep the full-body path for small objects / non-Range requests.

## Priority
P2 — performance/infra hardening, not launch-blocking (comparable tier to US-28.1 Performance Optimization). Benefits both video (US-22.3) and existing audio delivery.

## References
- PR #364 (adds the buffered video stream endpoint; documented there under Known Limitations)
- `src/acemusic/storage.py`, `src/acemusic/api/routers/videos.py` (`stream_video`), `src/acemusic/api/routers/clips.py` (`stream_clip_audio`)

frankbria:
## What & why

Closes #365 (P2.9). `GET /videos/{id}/stream`, `/clips/{id}/stream`, and `/clips/{id}/audio` all called `storage.download() -> bytes`, buffering the **entire** object into memory before slicing a Range. Fine for audio, an OOM/latency risk for large (1080p/4K) video renders where a few concurrent viewers could stall the event loop.

## Changes

- **`StorageBackend` gains two methods** (`src/acemusic/storage.py`):
  - `size(path) -> int` — object byte length without downloading it (Local: `stat()`; S3: `head_object` ContentLength; missing → `FileNotFoundError`).
  - `open_range(path, start, end) -> Iterator[bytes]` — yields the inclusive byte range a chunk at a time (Local: `seek` + chunked read; S3: `get_object(Range=)` → `iter_chunks`, body closed in `finally`).
  - Both are **concrete base-class defaults** (buffered via `download()`), so any existing subclass that implements only the original four methods keeps working; `LocalStorage`/`S3Storage` override for the efficient streaming path.
- **Shared async helper** `stream_stored_object(...)` (`api/utils/range_requests.py`): single-range (206) and full-body (200) responses stream via `StreamingResponse` (sync generator → Starlette threadpool, so the event loop isn't blocked), with correct `Content-Range` / `Content-Length`. Multi-range `multipart/byteranges` stays buffered — it's rare and already size-bounded by `MAX_MULTIRANGE_SPECS` + a total-bytes cap.
- **Endpoints rewired**: `stream_video`, `stream_clip_audio`, `get_clip_audio` resolve `size()` first (existence check → 404 before any streaming headers), then delegate to the helper. Conversion paths still download the whole object (a re-encode has a different byte layout, so Range doesn't apply).
- **Cache staleness (issue comment)**: lowered `Cache-Control: max-age` from `3600` → `300` on access-controlled media, bounding how long a shared cache can keep serving a resource after a visibility flip (publish → later make-private). No ETag/revalidation infra added — shorter max-age is the minimal fix.

## Tests / evidence

- New `tests/test_storage.py` cases for `size` + `open_range` on Local and S3 (mocked), missing → `FileNotFoundError`, chunked streaming of a large object, and a base-class-default regression guard (a subclass overriding only `download` still instantiates and works).
- Existing endpoint suites (`test_video_api`, `test_clips_api`, `test_clips`) stay green — 206/multipart/416/404/full-body all preserved.
- Manual check: a 5 MiB object serves `size()` via stat, `open_range` in multiple chunks with exact bytes, a 206 single-range with correct `Content-Range`/`Content-Length`, and a full-body 200 streamed in 20 chunks.

## Known limitations

- **Multi-range** (`multipart/byteranges`) still buffers the whole object via `download()`. It's rare (browsers seek with a single range) and bounded by the existing multi-range caps, so it wasn't worth a streaming multipart assembler.
- **Truncation race**: `size()` is read before streaming; if the object is truncated/deleted between `size()` and iteration, the stream ends short (or errors mid-transfer) rather than returning a clean 404. Narrow window, matches the prior data-integrity assumptions.

## Review

Cross-family pre-PR review via `codex review` flagged one P1 — new abstract methods would break existing `StorageBackend` subclasses (e.g. test `FailingStorage`) and the integration CI job. Fixed by making `size`/`open_range` concrete defaults; verified `FailingStorage` instantiates and the integration tests collect. (opencode/GLM hit a transient server error this run.)

### DIFF
--- a/src/acemusic/api/routers/clips.py
+++ b/src/acemusic/api/routers/clips.py
@@ -22,14 +22,13 @@ Access and filter rules live in :mod:`acemusic.api.services.clips`.
 import asyncio
 import logging
 import math
-import secrets
 from datetime import datetime
 from typing import Annotated, Literal
 
 from fastapi import APIRouter, Depends, HTTPException, Query, Request, Response, status
 from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
 
-from acemusic.storage import get_storage_backend
+from acemusic.storage import StorageBackend, get_storage_backend
 
 from ..auth.dependencies import CurrentUser, get_current_user, get_current_user_optional, require_existing_user
 from ..models import Clip, VisibilityState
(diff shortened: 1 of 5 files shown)


frankbria:
### Cross-family pre-PR review (codex)

`codex review --base main` on the branch diff flagged one blocking issue:

- **[P1] New abstract methods break existing `StorageBackend` subclasses** — making `size()`/`open_range()` `@abstractmethod` makes any concrete subclass that doesn't implement them impossible to instantiate, breaking test `FailingStorage` and the integration CI job.

**Resolved**: `size()`/`open_range()` are now **concrete base-class defaults** (buffered via `download()`); `LocalStorage`/`S3Storage` override for the streaming path. Verified `FailingStorage` instantiates and uses the defaults, and the integration tests collect. A regression test (`TestStorageBackendDefaults`) locks this in.

opencode/GLM hit a transient server error on this run, so codex was the cross-family reviewer.

PR #368, opened 2026-07-30
feat(plugin): US-23.1 JUCE Project Setup and Cross-Platform Build (#315)

### LINKED ISSUE #315: US-23.1: JUCE Project Setup and Cross-Platform Build
## User Story

**As a** developer, **I want** a JUCE-based VST3 plugin project that builds on Windows, macOS, and Linux, **so that** we have a solid foundation for the DAW integration features.

## Description

Set up the JUCE project in the `plugin/` directory of the monorepo. Configure CMake for cross-platform builds producing VST3 as the primary format and AU as a secondary format for macOS. The plugin should load in a DAW and display a minimal UI.

## Functional Requirements

- JUCE project in `plugin/` with CMakeLists.txt
- Build targets: VST3 (all platforms), AU (macOS only)
- CI build verification for Windows (MSVC), macOS (Xcode/clang), Linux (GCC)
- Plugin loads in a DAW host without crashing (validated with pluginval or equivalent)
- Minimal placeholder UI renders (plugin name, version, empty panels)
- Installed binary size target: ~20MB
- Non-audio thread infrastructure for HTTP communication (no network calls on audio thread)

## Acceptance Criteria

- [ ] VST3 builds successfully on all three platforms
- [ ] AU builds on macOS
- [ ] Plugin passes pluginval validation
- [ ] Plugin loads in at least one DAW (Reaper recommended for testing) and shows UI
- [ ] Binary size is under 25MB

## Dependencies

**Blocked by:** Stage 21. First story of Stage 23.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 23, US-23.1

frankbria:
## US-23.1: JUCE Project Setup and Cross-Platform Build

Closes #315. First story of Stage 23 — the VST3 plugin foundation.

`plugin/` was empty; this adds a JUCE 8.0.9 plugin project that builds VST3 on
all three platforms, AU on macOS, and a Standalone dev build, plus the
off-audio-thread work queue the rest of Stage 23 is built on.

### What's here

| Path | Purpose |
| --- | --- |
| `plugin/CMakeLists.txt` | JUCE via `FetchContent` (pinned tag `8.0.9`, shallow), plugin + test targets |
| `plugin/source/PluginProcessor.*` | Stereo/mono passthrough effect. `processBlock` allocates nothing, takes no locks, never touches the network |
| `plugin/source/BackgroundTaskQueue.*` | The off-audio-thread seam every ACE-Step call will use — `juce::ThreadPool` + message-thread result marshalling |
| `plugin/source/PluginEditor.*` | Placeholder UI: name, version, and the Connection / Generation / Results panels that US-23.2–23.4 fill in |
| `plugin/tests/` | 14 `juce::UnitTest` cases in a console runner, wired to `ctest` |
| `.github/workflows/plugin.yml` | Linux/macOS/Windows build → unit tests → size gate → pluginval strictness 10 |
| `plugin/README.md` | Local build/test/validate steps and the manual Reaper check |

### Acceptance criteria

| Criterion | Evidence |
| --- | --- |
| VST3 builds on all three platforms | Verified locally on Linux/GCC; macOS + Windows `Build VST3` steps green in the `Plugin` workflow on this PR |
| AU builds on macOS | macOS `Build AU` step green (the target only exists on Apple platforms) |
| Plugin passes pluginval | pluginval 1.0.3 **strictness 10, 25/25 groups, 0 failures, exit 0** locally; same job on all three OSes in CI |
| Plugin loads in a DAW and shows UI | pluginval's `Open plugin (cold/warm)`, `Editor`, and `Open editor whilst processing` groups pass — it instantiates the processor and repaints the editor as a host would. See Known Limitations for the Reaper check |
| Binary size under 25MB | **4744KB Linux / 3480KB macOS / 5384KB Windows**, against a 25600KB CI gate on every platform |

### Decisions worth knowing

- **`JUCE_USE_CURL=0`.** The plugin talks to a *local* ACE-Step server over plain
  HTTP on `localhost:8001`, which JUCE's built-in socket `WebInputStream`
  handles. Dropping curl removes a system dev-package requirement from Linux
  builds. Upgrade path is documented in `CMakeLists.txt` and the README: flip to
  `1` and add `libcurl4-openssl-dev` if a later story needs HTTPS to a remote host.
- **JUCE is fetched, not vendored or submoduled.** `-DFETCHCONTENT_SOURCE_DIR_JUCE=…`
  reuses an existing checkout; CI caches `plugin/build/_deps`. The cold-fetch
  path (exactly what CI does) was verified locally.
- **Plugin sources compile into both the plugin and the test runner** rather than
  through a shared static library, so the test binary needs nothing from JUCE's
  plugin-wrapper machinery.
- **`isBusesLayoutSupported` accepts mono→mono, stereo→stereo, and a disabled
  input bus only.** The output-clearing loop in `processBlock` exists for that
  last case, and is tested against it rather than against a layout the processor
  would reject.
- **Path-filtered workflow.** `plugin.yml` only runs when `plugin/**` or the
  workflow itself changes, so Python-only PRs don't pay for a 3-OS C++ build.

### Verification

- **Unit tests**: 14/14 pass (`ctest --test-dir build -C Release`).
- **Mutation check**: two logic mutations were introduced and both were caught —
  removing the output-clearing loop failed `clears output channels the host did
  not supply as inputs`; running `enqueue` inline on the caller's thread failed
  four `BackgroundTaskQueue` cases. The tests bite.
- **Build warnings**: zero (`-Wall` via `juce_recommended_warning_flags`; the only
  remaining line is GCC's benign `lto-wrapper: using serial compilation` note).
- **Cross-family review** (opencode / GLM-5-turbo, pre-PR): processBlock RT-safety,
  BackgroundTaskQueue thread-safety and lifetime, JUCE API usage, CMake, and the
  tests all came back clean. Its one finding — `actions/checkout@v7` "does not
  exist" — is a false positive: v7.0.1 is the current release and `main`'s CI is
  already green on it. No change made.
- **Demo**: `docs/demos/us-23.1-juce-plugin.md` — every criterion with executed
  output, plus a captured screenshot of the rendered UI.
- **Self-review fixes** (`cce7efa`), both real:
  - `waitForAll` compared against an absolute `getMillisecondCounter()` deadline,
    which reports a spurious timeout across the counter's ~49-day wrap. Now uses
    unsigned elapsed-time subtraction.
  - The `callOnMessageThread` test captured stack locals by reference; a late
    callback firing after the scope exited would have touched dead memory. Now
    captures shared state by value.
- **Review-round fix** (`9143f62`) — the review pressed on whether the queue's
  destructor bound actually held when a task ignores interruption. It didn't, and
  the header overstated it. `~juce::ThreadPool()` already does
  `removeAllJobs(true, 5000)` then `stopThreads()`, so the explicit
  `removeAllJobs(true, 2000)` only stacked another timeout onto the worst case;
  removed. Worse, `enqueue` took a plain `std::function<void()>`, so no task could
  observe the stop request even though the docs told authors to be cancellable —
  the advice was unfollowable. Added `isStopping()` for tasks to poll, and measured
  the real cost with a throwaway probe: an uncooperative 30s task makes teardown
  take **5502ms** and is then abandoned still running. The header now states that
  number; a new test covers the cooperating path (teardown returns in <1s and the
  task confirms it saw the stop request).

### Known limitations

- **The Reaper load check is manual.** The story's functional requirements accept
  "pluginval **or equivalent**", and pluginval at strictness 10 is the automated
  evidence — it opens the plugin cold and warm, instantiates and repaints the
  editor, fuzzes parameters, and exercises every bus layout. A real DAW cannot be
  driven headlessly in CI, so the Reaper step is documented in
  `plugin/README.md` as a pre-release manual check rather than automated here.
- **macOS AU and Windows VST3 are proven by CI on this PR, not locally** — this
  was developed on Linux. Both jobs must be green before merge.
- **AU is not code-signed or notarised**, so macOS hosts will need the usual
  local-development allowances. Distribution signing is out of scope for #315.
- **No persisted state yet.** `getStateInformation` writes zero bytes;
  US-23.2 stores the server URL, API key, and model selection there.
- **No parameters exposed to the host.** The plugin has nothing automatable yet,
  which is why pluginval's automation groups pass trivially.
- **An uncooperative background task stalls teardown for ~5.5s** and is then
  abandoned (measured, see above). `BackgroundTaskQueue::isStopping()` exists for
  tasks to poll; US-23.2's connection test is the first consumer obliged to use it.
- **JUCE licensing is undecided.** `JUCE_DISPLAY_SPLASH_SCREEN=0` is set, which is
  only permitted under a paid JUCE licence — so the build as it stands already
  assumes the commercial arm rather than AGPLv3. Surfaced in review, tracked in
  **#369**. Blocks distributable builds only, not US-23.2–23.5.
- **No checksum verification on the JUCE / pluginval fetches.** Both are pinned
  (tag `8.0.9`, release `v1.0.3`) and fetched over HTTPS from official sources, so
  the exposure is narrow, but a `sha256sum` gate would be worth adding if this
  workflow becomes the template for further dependency fetches.


frankbria:
## Cross-family review — opencode / GLM-5-turbo

Third-party pass on the branch diff (1129 lines), asked to look for real defects
only across six dimensions. Verbatim verdicts:

| Dimension | Verdict |
| --- | --- |
| `processBlock` real-time safety | **Clean** — no allocation, no locks, no network. `ScopedNoDenormals` and `buffer.clear` are both RT-safe; the channel-count accessors read cached values. |
| `BackgroundTaskQueue` thread-safety / lifetime | **Clean** — `removeAllJobs(true, timeout)` + `~ThreadPool()` join is a correct two-phase shutdown; the static `callOnMessageThread` correctly delegates to `MessageManager::callAsync`; `waitForAll` polls `getNumJobs()` under the pool's internal lock. |
| JUCE API usage | **Clean** — `isBusesLayoutSupported`, the `getStateInformation`/`setStateInformation` empty-state contract, `createEditor`'s host-owns-the-pointer contract, and `using juce::AudioProcessor::processBlock` all correct. |
| CMake | **Clean** — FetchContent pins 8.0.9 with `GIT_SHALLOW ON`; compiling the sources into both targets correctly avoids linking the test binary against JUCE's plugin wrapper; `JUCE_MODAL_LOOPS_PERMITTED=1` and `JUCE_UNIT_TESTS=1` are `PRIVATE` to the test target only. |
| Tests | **Clean** — "All meaningful. Each test asserts a concrete observable property. No vacuous assertions found." |
| GitHub Actions workflow | **One finding** (below). Everything else correct: bash on all three runners, empty `$XVFB` word-splits cleanly, `du -sk \| cut -f1` portable across Git Bash, the macOS `pluginval.app/Contents/MacOS/pluginval` path, and the cache key hashing the file that holds the JUCE tag. |

### The one finding — not actioned, and why

> **Major** — `.github/workflows/plugin.yml`: `actions/checkout@v7` — this tag does
> not exist (latest is `@v4`). The workflow will fail immediately on every run.

**False positive, verified before acting:**

```
$ gh api repos/actions/checkout/releases --jq '.[0].tag_name'
v7.0.1
$ gh api repos/actions/checkout/git/ref/tags/v7 --jq '.ref'
refs/tags/v7
```

`v7` exists, `.github/workflows/ci.yml` on `main` already pins it, and `main`'s CI
is green. The reviewer's cutoff predates the release. No change made — and the
`Plugin` workflow runs on this PR confirm the checkout step works.

### Claude-side findings (fixed in `cce7efa`)

Two real defects the cross-family pass did not catch:

1. **`BackgroundTaskQueue::waitForAll` wrap bug.** It compared against an absolute
   `juce::Time::getMillisecondCounter()` deadline. That counter is a `uint32` that
   wraps roughly every 49 days; across the wrap, `now >= deadline` is immediately
   true and the method reports a spurious timeout on a healthy queue. Now uses
   unsigned elapsed-time subtraction, which is wrap-correct.
2. **Dangling stack captures in the async test.** The `callOnMessageThread` test
   captured stack locals by reference. If the message-thread callback arrived
   after the enclosing scope exited (the failure path — exactly when it's most
   likely), it would write to dead memory. Shared state is now captured by value.

### Mutation check

Verified the tests actually bite rather than passing vacuously:

| Mutation | Caught by |
| --- | --- |
| Removed the output-clearing loop in `processBlock` | `PluginProcessor / clears output channels the host did not supply as inputs` |
| Made `enqueue` run its task inline on the caller's thread | 4 `BackgroundTaskQueue` cases (`runs queued work off the calling thread`, `reports pending work`, `waitForAll reports failure on timeout`, `callOnMessageThread delivers on the message thread`) |

Both reverted; 13/13 green on the committed code.

frankbria:
Thanks — the shutdown question was the right one to press on, and the answer was
worse than the header claimed. Fixed in `9143f62`.

## The shutdown bound — you were right, and it was measured

`~juce::ThreadPool()` is:

```cpp
ThreadPool::~ThreadPool()
{
    removeAllJobs (true, 5000);
    stopThreads();          // signalThreadShouldExit(), then stopThread(500) per worker
}
```

So my explicit `removeAllJobs (true, shutdownTimeoutMs /* 2000 */)` was **redundant
and actively misleading**: the pool destructor does the same thing immediately
afterwards with a longer timeout, so the 2s call only stacked another timeout onto
the worst case without changing the outcome. Removed, along with the
`shutdownTimeoutMs` constant that documented a bound that never existed.

Then I measured it rather than reasoning about it, exactly as you suggested — a
throwaway probe enqueueing a task that sleeps 30s and ignores interruption:

```
PROBE teardown with an UNCOOPERATIVE 30s task: 5502 ms
PROBE task had finished by then: no
```

5000ms + 500ms + overhead, and the task is **abandoned still running**. In a DAW
that is a five-and-a-half-second freeze on plugin close plus a leaked thread — not
the ~2s the header implied.

## The deeper problem your question exposed

The header told task authors to be "cancellable by their owner", but `enqueue`
takes a plain `std::function<void()>`, which has no way to observe JUCE's
`signalJobShouldExit`. **No task could have cooperated even if it wanted to.** The
advice was unfollowable.

So there's now a `isStopping()` the task can poll, set by the destructor before the
pool tears down:

```cpp
private:
    // Declared before `pool` so it is destroyed *after* it: members tear down in
    // reverse order, so a job still winding up inside ~ThreadPool() can safely
    // read this flag.
    std::atomic<bool> stopping { false };
    juce::ThreadPool pool;
```

Member declaration order is load-bearing there, so it's commented. And the header
now states the measured 5502ms figure and what it costs in a DAW, instead of
implying a bound it didn't have.

A new test (14 now, up from 13) covers the cooperating path — a task that polls
`isStopping()` around its slow steps lets teardown return in under a second, and
asserts the task actually observed the stop request rather than just happening to
finish. I deliberately did **not** commit the uncooperative probe: it leaves an
abandoned thread sleeping for 30s, which has no business running in CI. The header
says that's why the number came from a probe.

US-23.2's connection test will be the first real consumer — it'll poll
`isStopping()` and call `juce::WebInputStream::cancel()`.

## Your other points

- **LTO asymmetry on the test target** — fair, everything else got a comment. Added
  one: LTO roughly doubles link time and the test binary is never shipped.
- **No `PluginEditor` tests** — agreed, and deliberate for now: `PlaceholderPanel`
  is two `paint` calls with no logic, and pluginval's `Editor` /
  `Open editor whilst processing` groups already prove it instantiates and
  repaints. It stops being acceptable the moment US-23.2 puts a URL field and a
  connection state machine in there, so that's where editor tests land.
- **No checksum on the JUCE / pluginval fetches** — noted as a real gap, not
  actioned here. JUCE is pinned to tag `8.0.9` (an immutable ref) and pluginval to
  release `v1.0.3`, both over HTTPS from official sources, so the exposure is
  narrow. Worth a `sha256sum` gate if this workflow becomes the template for
  further dependency fetches; that's a workflow-hardening change rather than
  something US-23.1 should carry.
- **JUCE GPLv3/commercial licensing** — correct and out of scope for this PR, but a
  real pre-distribution decision. Filed separately rather than dropped, so it
  can't get lost before a shippable build: **#369**. Note that `JUCE_DISPLAY_SPLASH_SCREEN=0` is already set, which is only permitted under a paid licence — so the build as it stands already assumes the commercial arm. That inconsistency is captured in the issue.

## Re-verified after the change

| Check | Result |
| --- | --- |
| Unit tests | 14/14 pass |
| Build warnings | 0 |
| pluginval strictness 10 | exit 0, 25/25 groups, 0 failures |
| VST3 size | 4744KB (limit 25600KB) |

PR #370, opened 2026-07-30
fix(ci): skip Docker Publish when Docker Hub secrets are absent

frankbria:
## Why Docker Publish fails on main

Every `Docker Publish` run on `main` — going back at least to 2026-07-19 — fails in the same place:

```
X Log in to Docker Hub
ANNOTATIONS
X Username and password required
```

The workflow is correct; the credentials simply do not exist. `gh secret list` on this repo returns only `CLAUDE_CODE_OAUTH_TOKEN` and `ZHIPU_API_KEY` — **`DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` have never been configured.** `docker/login-action` fails hard on empty inputs, so the job dies ~15s in and the build/push steps never run.

The reason it looks like a recurring regression is the path filter: the workflow triggers on changes to `.github/workflows/docker-publish.yml` itself, so each Dependabot action bump that merges to `main` (#286, #288, #289, #294, #356) immediately re-triggers a run that fails for the same missing-secret reason.

Failed runs, all identical: [30416412958](https://github.com/frankbria/auto-music-studio/actions/runs/30416412958), [29673446436](https://github.com/frankbria/auto-music-studio/actions/runs/29673446436), [29673444348](https://github.com/frankbria/auto-music-studio/actions/runs/29673444348), [29673442038](https://github.com/frankbria/auto-music-studio/actions/runs/29673442038), [29673437897](https://github.com/frankbria/auto-music-studio/actions/runs/29673437897).

## What this PR changes

Guards the credential-dependent steps on both secrets being present, and emits a warning annotation when they are not — so the job **skips** instead of failing.

- A single `HAS_DOCKERHUB_CREDS` flag is computed in job-level `env`, because the `secrets` context is not available in `if:` conditions but `env` is.
- Gated on *both* secrets: a username set without a token fails `docker/login-action` just as hard as neither being set (caught in pre-PR review).
- Validated with `actionlint` (clean).

## What this PR does *not* do

It does not make publishing work — I can't create the Docker Hub token. **To actually publish `frankbria/ace-step` again:** add `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` under Settings → Secrets and variables → Actions, then run the workflow from the Actions tab. The guard is a no-op once they exist.

Note that the build/push steps have never successfully executed on CI, so `docker/Dockerfile` is unverified in this pipeline — the first run with real credentials may surface build issues.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -7,6 +7,9 @@
 # Required repository secrets (Settings → Secrets and variables → Actions):
 #   - DOCKERHUB_USERNAME  Docker Hub account name (e.g. frankbria)
 #   - DOCKERHUB_TOKEN     Docker Hub access token with read/write scope
+#
+# When those secrets are absent the job logs a warning and skips rather than failing, so
+# unrelated pushes to docker/ do not report a red build on main.
 name: Docker Publish
 
 on:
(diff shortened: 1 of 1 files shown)


frankbria:
### Pre-PR third-party review

`opencode`/GLM failed with an upstream server error (`err_d6bc536c`), so `codex review` ran as the fallback per repo standard. It raised one finding:

> **[P2] Check both Docker Hub secrets before publishing** — when `DOCKERHUB_USERNAME` is set but `DOCKERHUB_TOKEN` is absent, the condition still runs the login step and the workflow fails instead of warning and skipping as documented.

Valid. Fixed in 8d038cf by collapsing the guard to a single `HAS_DOCKERHUB_CREDS` flag that requires both secrets to be non-empty. `actionlint` clean after the change.

Issue #369, opened 2026-07-30
P2.1 Decide JUCE licensing (GPLv3 vs commercial) before distributing the VST3 plugin


frankbria:
## Context

US-23.1 (#315, PR #368) added a JUCE 8.0.9 plugin in `plugin/`, fetched via CMake
`FetchContent` from the official repo at tag `8.0.9`. Surfaced during review of
that PR — not a defect in it, and not a blocker for Stage 23 development.

JUCE 8 is dual-licensed: the **JUCE 8 End User Licence Agreement** (commercial,
tiered by revenue) or **AGPLv3**. `plugin/` currently carries no licence note, and
the repo has not recorded which arm applies.

## Why this needs a decision

The choice is only binding at **distribution** time, but it constrains things well
before that:

- Under AGPLv3, distributing the plugin obliges us to offer the plugin's complete
  corresponding source under AGPLv3 — which reaches the plugin code in `plugin/`,
  and arguably shapes how the plugin may talk to the rest of the platform.
- Under the commercial EULA, there is no such obligation, but it requires a paid
  tier and pins us to its terms (including its splash-screen and revenue-reporting
  rules).
- `JUCE_DISPLAY_SPLASH_SCREEN=0` is already set in `plugin/CMakeLists.txt`.
  Disabling the splash screen is **only permitted under a paid licence** — under
  AGPLv3 the splash screen must stay. So the build as it stands already assumes the
  commercial arm without that having been decided.

That last point is the concrete reason this can't sit indefinitely: the current
configuration is inconsistent with the free arm.

## Done when

- [ ] The licensing arm is chosen and recorded (in `plugin/README.md` and the repo's
      top-level licence documentation).
- [ ] `JUCE_DISPLAY_SPLASH_SCREEN` is set consistently with that choice — `0` only
      if a paid licence is actually held, otherwise removed.
- [ ] If commercial: the licence is obtained and the revenue tier confirmed.
- [ ] If AGPLv3: the source-offer obligation is understood and documented, and the
      splash screen is re-enabled.
- [ ] `plugin/` has a licence header or `LICENSE` note matching the decision.

## Not blocked by / does not block

Independent of US-23.2–23.5 (#316–#319) — those can proceed. This blocks any
**distributable or public** plugin build.

## References

- Raised in review of PR #368 (US-23.1).
- JUCE licensing: https://juce.com/legal/juce-8-licence/

Issue #317, opened 2026-07-23
US-23.3: Generation Panel


frankbria:
## User Story

**As a** musician, **I want** to enter a prompt, configure generation parameters, and trigger music generation from inside my DAW, **so that** I can create AI music without leaving my production environment.

## Description

The generation panel mirrors the core creation controls from the web app — prompt, lyrics, vocal language, instrumental toggle, BPM, key, duration, seed, and quality preset. The musician fills in the fields and clicks Generate. The request is sent via HTTP REST on a non-audio thread.

## Functional Requirements

- Prompt textarea for style/caption text
- Lyrics textarea (collapsible) with support for structure tags ([Verse], [Chorus], etc.)
- Vocal language dropdown (50+ languages)
- Instrumental toggle
- BPM field (numeric input, or "Auto")
- Key selector ("Any" or specific musical key)
- Duration field (seconds, numeric input)
- Seed field (numeric or "Random")
- Quality preset selector: Turbo / Standard / High (maps to inference steps)
- Generate button — submits to the ACE-Step REST API via non-audio HTTP thread
- Progress bar during generation (polls job status)
- Generation disabled when connection status is Red
- Text-to-Music and Cover modes selectable via a mode toggle

## Acceptance Criteria

- [ ] Filling in a prompt and clicking Generate produces 2 audio clips
- [ ] Progress bar updates during generation and completes when clips are ready
- [ ] All parameter fields are sent correctly to the API (verified via request logging)
- [ ] Generation does not cause audio dropouts or DAW UI freezing
- [ ] Cover mode accepts a source audio reference

## Dependencies

**Blocked by:** US-23.2 (prior story in this stage). Part of Stage 23: VST3 Plugin — Core, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 23, US-23.3

frankbria:
## Design constraint carried over from US-23.2 (#316)

US-23.2 added the first real HTTP traffic (`probeAceStepServer` → `GET /v1/stats`).
While reviewing it I found a bound that **US-23.3 has to solve properly**, because
generation requests are long-running where a stats probe is not.

### The problem

A probe runs on a `BackgroundTaskQueue` worker and cannot interrupt its own blocking
socket call. `juce::WebInputStream::withConnectionTimeout` bounds the *connect*, and
JUCE's non-curl Linux read path does `poll(fd, timeoutMs)` before each read — so
[...]

PR #371, opened 2026-07-30
feat(plugin): US-23.2 Connection Status Panel (#316)

### LINKED ISSUE #316: US-23.2: Connection Status Panel
## User Story

**As a** musician, **I want** to configure and verify the connection to my local ACE-Step server from within the plugin, **so that** I know the AI engine is ready before I try to generate.

## Description

The connection panel is the first thing a musician sees when opening the plugin. It shows the server URL, API key field, model selector, connection test button, and a clear status indicator. The plugin requires a running ACE-Step-1.5 server at `localhost:8001` (configurable).

## Functional Requirements

- Server URL text field (default: `http://localhost:8001`)
- API key text field (optional, for secured setups)
- "Test Connection" button — calls the server's health endpoint
- Status indicator: Green (connected, model loaded), Yellow (connecting/testing), Red (offline/error)
- Model selector dropdown — populated from the server's available models on successful connection
- Connection settings persist between DAW sessions (saved to local config file)
- Auto-connect on plugin load (non-blocking, updates indicator when complete)

## Acceptance Criteria

- [ ] With ACE-Step running: Test Connection shows green indicator and populates model list
- [ ] With ACE-Step stopped: Test Connection shows red indicator with "Server unreachable" message
- [ ] Custom server URL is saved and restored across DAW sessions
- [ ] Model dropdown lists all available models from the server
- [ ] Auto-connect attempt occurs on plugin load without blocking the DAW

## Dependencies

**Blocked by:** US-23.1 (prior story in this stage). Part of Stage 23: VST3 Plugin — Core, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 23, US-23.2

frankbria:
## US-23.2: Connection Status Panel

Closes #316. Second story of Stage 23, on top of US-23.1 (#315, merged `7d08633`).

Replaces the Connection placeholder with a working panel, and adds the plugin's
first real HTTP traffic — the payoff for the `BackgroundTaskQueue` seam.

### The contract the issue gets wrong

The issue says *"Test Connection button — calls the server's health endpoint"*.
**ACE-Step has no `/health`.** The availability probe this repo already uses is
`GET /v1/stats` (`AceStepClient.get_stats` in `src/acemusic/client.py`,
`compute_status.LOCAL_STATS_PATH`), envelope
`{"data": {"models": [{"name": …}], …}}`, auth `Authorization: Bearer <key>`. The
plugin mirrors the Python client rather than the issue text, and the model list
comes from `data.models[].name`.

### What's here

| Path | Purpose |
| --- | --- |
| `source/AceStepClient.{h,cpp}` | `probeAceStepServer()` — `GET /v1/stats`, envelope parsing, optional Bearer auth, cancellable |
| `source/ConnectionManager.{h,cpp}` | Settings + status + model list, **owned by the processor**; lock-free |
| `source/ConnectionSettings.{h,cpp}` | `juce::PropertiesFile` persistence, 0600 on POSIX |
| `source/ConnectionPanel.{h,cpp}` | URL, masked API key, Test button, traffic-light + message, model dropdown |
| `tests/StubAceStepServer.h` | A **real** HTTP server on a real loopback socket |
| `scripts/demo_stub_acestep.py` | Stand-in ACE-Step for the demo |

### Acceptance criteria

| Criterion | Evidence |
| --- | --- |
| Server running → green + model list | Screenshot of a live connection + `ConnectionPanel` test asserting the indicator colour *off the component that paints it* and the dropdown contents |
| Server stopped → red + "Server unreachable" | Screenshot + three tests against a genuinely closed port |
| URL saved/restored across sessions | The written settings file, then a relaunch restoring it; plus a two-"session" round-trip test over one file |
| Model dropdown lists all server models | Dropdown holding the server's three models, first auto-selected |
| Auto-connect on load, non-blocking | **Both screenshots were reached with zero clicks.** Test asserts `autoConnect()` returns in <100ms and only *then* reaches Connected |

Full mapping in `docs/demos/us-23.2-connection-panel.md`.

### Decisions worth knowing

- **Connection state lives on the processor, not the editor.** The editor is
  destroyed whenever the user closes the plugin window; auto-connect happens on
  *plugin* load and its result has to outlive that. There's a test that connects,
  destroys the editor, builds a new one, and asserts the live status is still there.
- **Lock-free by construction.** `testConnection()` snapshots the settings into the
  worker lambda, so no manager state is read off the message thread. The result
  returns via `callOnMessageThread` behind a `juce::WeakReference`, so a probe
  finishing after teardown is a no-op. Member declaration order in
  `PluginProcessor` is load-bearing (queue declared first → destroyed last) and
  commented as such.
- **Settings in a `PropertiesFile`, not project state**, because the requirement is
  to survive *sessions* — project state only returns inside the same project.
- **No mocks.** Tests run against a real loopback HTTP server, because a mocked
  client would pass even if US-23.1's curl-free `WebInputStream` were broken. That
  assumption was spiked before any of this was written.
- **`PluginProcessor` gained a test-seam constructor** so tests never write to the
  user's real config or hit the network on construction.

### Verification

- **56 unit tests** pass (42 new), zero build warnings, `xvfb-run -a ctest`.
- **pluginval strictness 10**: exit 0, 25/25 groups, 0 failures. VST3 4868KB.
- **Mutation check** — four mutations, each caught by exactly the intended test:
  dropping the Bearer header; reporting a 2xx-with-no-models as success; keeping a
  stale `Connected` after the server changed; skipping the settings `chmod`.
- **Cross-family review** (opencode / GLM-5-turbo, pre-PR): no critical or major
  defects; thread-safety, lifetime, and the member-ordering argument all confirmed
  sound. Two of its three minors were worth fixing and are fixed (below); the third
  (a `findClosedPort()` TOCTOU it judged not worth fixing) is left alone.

### Bugs found and fixed during this PR

1. **The probe's worst case didn't fit the teardown budget — my own error, and the
   one that matters.** I documented a bound the code didn't have. JUCE's non-curl
   read polls with the connection timeout but then calls `recv(…, MSG_WAITALL)`,
   which blocks until the requested bytes arrive. Even the *normal* stalled case
   (3s connect + 3s read) overran `juce::ThreadPool`'s ~5s teardown allowance.
   Timeout is now 2s so the normal worst case (~4s) fits, and the residual
   unbounded case is stated plainly in `AceStepClient.h` and `README.md` instead of
   being papered over. Recorded as a design constraint on **#317**, which has to
   build real cancellation for long-running generation requests.
2. **Settings landed in `~/AutoMusicStudio/`** — JUCE resolves `folderName` under
   `$HOME` on Linux, so a bare vendor name drops a *visible* directory into the
   user's home. Now under `.config`, which is what the README already claimed. The
   test asserts the exact resolved path.
3. **The status line rendered as `Connected â—� 3 models available`.** The em-dash in
   a user-facing literal came out as mojibake. **Every test compared the message
   against another literal written the same way, so the entire suite was blind to
   it** — only the demo screenshot caught it. User-facing strings are ASCII now,
   guarded by a new `isAsciiOnly` test.
4. *(review)* The URL scheme check only required a non-empty scheme, so `ftp://` or
   `file://` reached the socket and returned a vague connection error. Now requires
   http/https with a clear message.
5. *(review)* `refreshFromManager` reassigned the ComboBox selection on every status
   broadcast, which could yank an open dropdown. Now only writes when the selection
   is actually wrong.

Three of those five were caught by the demo or by reading JUCE's source, not by the
tests — which is the argument for the demo gate.

### Known limitations

- **A probe is bounded in the normal case, not absolutely** — see fix 1 above and
  the constraint on #317.
- **The amber "Connecting" state is asserted, not screenshotted.** Against a local
  server the probe completes in ~1ms; capturing it would mean holding the response
  artificially. A test delays the stub 1.5s and asserts the in-flight state.
- **The API key is stored in plaintext** (0600 on POSIX, per-user ACL on Windows).
  JUCE has no keychain abstraction. Documented in `plugin/README.md`. Note the
  acceptance criteria only require the *URL* to persist, so dropping key
  persistence is a one-line change if that trade is unwanted.
- **No real ACE-Step server was exercised.** The stub serves the envelope
  documented in `src/acemusic/client.py`; if the real server has drifted from what
  the Python client parses, both are wrong together.
- **macOS and Windows are proven by CI on this PR, not locally** — developed on Linux.
- **No `PluginEditor` tests for the Generation/Results placeholders** — still two
  `paint` calls with no logic. They get tests when US-23.3/23.4 give them behaviour.
- **JUCE licensing is still undecided** (#369) — unchanged by this PR, blocks
  distributable builds only.


frankbria:

Full output below. **No critical or major defects.** Three minors; two fixed, one left alone with reasoning.

### What I did with each finding

| Finding | Action |
| --- | --- |
| `findClosedPort()` TOCTOU (Minor) | **Not fixed.** Bind-then-release is racy in theory, but every alternative has the same window, and the reviewer's own judgement was "not worth fixing" — a collision makes the test time out rather than falsely pass. |
| `setSelectedId` on every refresh could disturb an open dropdown (Minor) | **Fixed** — only writes when the selection is actually wrong, plus a test that a re-probe preserves the user's choice. |
| Scheme check allowed non-http(s) URLs (Minor) | **Fixed** — requires http/https explicitly with a clear message, plus tests for `ftp://`, `file://`, `ws://`. |

### One place the review was too optimistic, and I disagreed with it

On the probe timeout it concluded:

> `readEntireStreamAsString()` is unbounded in theory, but the stub server sends `Content-Length` + `Connection: close`, and the real ACE-Step `/v1/stats` response is small JSON — not a practical concern for localhost.

I'd written the same optimistic claim into the README, and reading JUCE's source showed both of us were wrong. `juce_Network_linux.cpp` polls with the connection timeout but then calls `recv(socketHandle, buffer, bytesToRead, MSG_WAITALL)`, which blocks until the requested byte count arrives. A server that sends a partial body and then neither sends more nor closes blocks the worker indefinitely — and the *normal* stalled case (3s connect + 3s read) already overran `juce::ThreadPool`'s ~5s teardown allowance, so teardown would abandon the thread.

Fixed the part that's fixable (2s timeout, so the normal worst case ~4s fits) and stated the residual unbounded case plainly in `AceStepClient.h` and `plugin/README.md` rather than leaving a guarantee the code doesn't provide. Recorded as a design constraint on #317, which has to build real `WebInputStream::cancel()`-based cancellation for long-running generation anyway.

This is the same class of mistake as US-23.1's shutdown bound, which is now a written lesson: don't document a timing guarantee that comes from a third-party destructor without reading its source *and* measuring it.

### Two more bugs neither review caught — the demo did

1. Settings landed in `~/AutoMusicStudio/` (a **visible** directory in $HOME) because JUCE resolves `folderName` under $HOME on Linux. The README already claimed `~/.config/`; the code now matches, asserted on the exact resolved path.
2. The status line rendered as `Connected â—� 3 models available` — an em-dash in a user-facing literal came out as mojibake. **Every test compared the message against another literal written the same way, so the whole suite was blind to it.** Only the screenshot found it. User-facing strings are ASCII now, guarded by an `isAsciiOnly` test.

---

## PR Review: US-23.2 Connection Panel

### 1. Thread-safety and lifetime — Clean

The lock-free design is sound:
- **WeakReference contract**: `ConnectionManager::~ConnectionManager()` calls `masterReference.clear()` before any member is destroyed. Worker lambdas capture `weakSelf` by value; `weakSelf.get()` returns null after the destructor body runs. Since `applyResult` is only called via `callOnMessageThread` (message-thread-only), and the destructor also runs on the message thread, the callback cannot interleave with destruction.
- **queuePtr raw pointer**: Safe because `BackgroundTaskQueue` is declared *before* `ConnectionManager` in `PluginProcessor`, so it's destroyed *after*. The probe is a ThreadPool job — `~ThreadPool()` won't finish until all jobs complete. Inside the job, `queuePtr->isStopping()` reads the `stopping` atomic, which is declared before `pool` in `BackgroundTaskQueue` and thus destroyed after `pool`. Valid throughout.
- **Snapshot pattern**: `testConnection()` copies `settings.serverUrl` and `settings.apiKey` by value into the lambda. The worker never reads `ConnectionManager` members directly. No data race.
- **Member declaration order** in `PluginProcessor` (line 81–84): `backgroundQueue`, `properties`, `connectionManager`. Destruction is reverse — connectionManager dies first, clearing the WeakReference; backgroundQueue dies last, waiting for any in-flight probe to complete. Comment at line 78–80 correctly documents this as load-bearing.

### 2. PropertiesFile lifetime — Clean

`properties` (raw pointer) is only dereferenced on the message thread — in the constructor, `setSettings`, `setSelectedModel`, and `applyResult` (the latter guarded by `weakSelf.get()`). It points to the `properties` member of `PluginProcessor`, which is destroyed after `connectionManager`. The WeakReference guard ensures `applyResult` never runs after `connectionManager`'s destructor.

### 3. HTTP correctness — Clean

- **URL building** (`AceStepClient.cpp:168`): `trimCharactersAtEnd("/")` before appending `/v1/stats` prevents double-slash. Multiple trailing slashes are handled. Correct.
- **Status code classification** (lines 204–216): 401/403 get a specific user-facing message distinguishing "no key" from "wrong key"; all other non-2xx get a generic message. Correct.
- **cancel()**: Called at line 221 after a successful connect+read when `stopRequested()` becomes true. When connect fails (line 188–200), the stream is already disconnected — nothing to cancel. Not called mid-connect (JUCE can't interrupt `WebInputStream::connect`), but the 3s timeout is deliberately below ThreadPool's ~5s teardown budget. Correct.
- **Timeout**: `withConnectionTimeout(3000)` bounds the connect phase. `readEntireStreamAsString()` is unbounded in theory, but the stub server sends `Content-Length` + `Connection: close`, and the real ACE-Step `/v1/stats` response is small JSON — not a practical concern for localhost.

### 4. Security — Clean

- API key is sent as `Authorization: Bearer` header (line 186), never in the URL.
- No error message, status label, or log line includes the key value. The messages only reference "API key" generically (lines 207–208).
- Plaintext storage with chmod 0600 is documented in README with the `~/.aws/credentials` comparison. No keychain abstraction in JUCE — acceptable posture for a desktop plugin.

### 5. Tests — Sound with one minor note

- **StubAceStepServer**: Real HTTP on loopback, port scan 18200–18299, `Connection: close`, sliced delay loop for clean shutdown. `stop()` calls `signalThreadShouldExit()` + `listener.close()` + `stopThread(2000)`. Reads and writes are under a CriticalSection. Atomic `requestCount`. No issues.
- **findClosedPort() TOCTOU** (`StubAceStepServer.h:2304–2319`, **Minor**): Binds a port, immediately closes it, returns the port number. Another process could bind it in between. The "Server unreachable" test would then see a different error (e.g., "no models") instead. Extremely unlikely in CI (range 18400–18499) and the test would time out rather than falsely pass. Not worth fixing.
- **No vacuous tests**: The "probe outliving manager" test uses `expect(true, ...)` but its value is crash/ASAN detection, which is standard for lifetime tests.
- **`pumpUntil` + `waitForAll` sequencing**: The "probe outliving manager" test correctly calls `waitForAll` then manually pumps the message loop (20 × 10ms) to flush the `callAsync` callback. This handles the fact that `waitForAll` only drains the thread pool, not the message queue.

### 6. UI — Clean with one minor note

- **Text editor guard** (`ConnectionPanel.cpp:743–747`): `hasKeyboardFocus(true)` is checked before setting URL and API key text. Won't clobber an in-progress edit.
- **Model selector** (`ConnectionPanel.cpp:757–774`): The dropdown list is only rebuilt when the actual model names change (`current != models` check). The `setSelectedId` at the bottom runs on every refresh, but uses `dontSendNotification` so it doesn't trigger `onChange` or cause recursion. **Minor**: if a user has the dropdown open when a status broadcast fires, the selection could visually jump. Practically impossible since broadcasts only fire on state transitions (Connecting → Connected/Error), and the user would need to have the popup open at that exact instant.

### Summary

| # | File | Lines | Severity | Finding |
|---|---|---|---|---|
| 1 | `tests/StubAceStepServer.h` | 2304–2319 | Minor | `findClosedPort()` TOCTOU — bind-then-release is racy in theory; no practical impact |
| 2 | `source/ConnectionPanel.cpp` | 769–774 | Minor | `setSelectedId` runs on every refresh, could visually interfere with an open dropdown during a probe |
| 3 | `source/AceStepClient.cpp` | 169–171 | Minor | Scheme check allows non-http(s) URLs (e.g. `ftp://`); produces a confusing error rather than an early rejection |

**No critical or major defects.** The lock-free WeakReference + snapshot pattern, member declaration ordering, and message-thread-only state mutation are all correctly implemented. The code is production-quality.

</details>

PR #372, opened 2026-07-30
feat(plugin): US-23.3 Generation Panel (#317)

### LINKED ISSUE #317: US-23.3: Generation Panel

frankbria:
## US-23.3: Generation Panel

Closes #317. Third story of Stage 23, on US-23.2 (#316, merged `6155625`).

Text prompt → parameters → Generate → two clips on disk, all from inside the DAW.

### The contract the issue understates

The issue says the request is "sent via HTTP REST". **It is three calls, not one.**
Verified against `src/acemusic/client.py`:

| Step | Endpoint | Notes |
| --- | --- | --- |
| Submit | `POST /release_task` | returns `data.task_id` (or `data.id`) |
| Poll | `POST /query_result` | `{"task_id_list":[id]}`; status `0` queued/running, `1` ok, `2` failed |
| Download | `GET /v1/audio?path=…` | one call per clip |

`result` is a **JSON string** holding `[{"file": "/v1/audio?path=…"}]`, not a nested
array — a parser that assumed otherwise passes a naive test and fails the real server.

### What's here

| Path | Purpose |
| --- | --- |
| `source/GenerationRequest.*` | The 11 controls + mode → a `/release_task` body |
| `source/AceStepClient.*` | `submitGeneration` / `queryTask` / `downloadAudio`, plus a shared `prepareEndpoint` helper |
| `source/GenerationManager.*` | The submit→poll→download state machine, on the processor |
| `source/GenerationPanel.*` | The controls, Generate/Cancel, progress |
| `scripts/demo_stub_acestep.py` | Now serves the whole plugin-facing API + real WAV audio |
| `scripts/demo_x_input.py` | XTEST input injection, so the demo types and clicks for real |

### Acceptance criteria

| Criterion | Evidence |
| --- | --- |
| Prompt + Generate produces 2 clips | Typed and clicked with **real X events** in the demo; server log shows `batch_size: 2` and two downloads; **two valid 44.1kHz WAVs on disk** |
| Progress updates and completes | Screenshots of running and complete; test asserts submitting → queued → running → downloading → complete |
| All parameters sent correctly (via request logging) | The server's own log of the submit payload, plus a test asserting the captured `/release_task` body field by field |
| No dropouts / UI freezing | `processBlock` unchanged from US-23.1; test asserts the Generate click returns in well under a second while work continues |
| Cover mode accepts a source reference | `task_type: "cover"` + `src_audio_path` asserted; see Known Limitations |

Full mapping in `docs/demos/us-23.3-generation-panel.md`.

### Decisions worth knowing

- **Auto / Any / Random are omitted, not sent as sentinels.** The server reads an
  absent key as "you choose" but would take `-1` or `""` literally. Seed `0` is
  deliberately kept — `-1` is the sentinel, and dropping `0` would silently
  randomise a run the user wanted reproducible.
- **Quality → `inference_steps` is not invented**: `client.py:121` documents
  "Turbo: 8, Standard: 32–64". Turbo 8 / Standard 48 / High 120.
- **Progress is state-based, not a percentage.** `/query_result` returns a status
  integer and nothing else. A crawling bar would be inventing server state, so the
  bar is indeterminate and the panel reports phase + elapsed time.
- **The worker touches no member state.** `runGeneration` is a *static* function
  taking a `RunContext` by value, with the stop flag in a `shared_ptr` so
  superseding or destroying the manager can signal a run without the worker ever
  dereferencing it. An earlier draft of mine read `currentRun` and
  `connection.getSettings()` off the object and captured raw `this` — the same
  race/lifetime class fixed in review on #316. Caught and restructured before it
  left my machine.
- **No watchdog.** I had claimed on #317 that this story "must" build one; that
  over-specified it and I posted a correction. Generation is many short requests,
  not one long one, so cancellation falls out of the poll loop (checked in 100ms
  slices). The residual single-hung-request exposure is unchanged from US-23.2.
- **`prepareEndpoint` extracted** from `probeAceStepServer` — four call sites now,
  so it is real dedup rather than speculative structure.

### Verification

- **104 test groups** (45 new), zero warnings, `xvfb-run -a ctest`.
- **pluginval strictness 10**: exit 0, 25/25, VST3 4948KB.
- **Mutation check** — six mutations. Five caught by exactly the intended test
  (seed `0` dropped; `task_type` sent for text2music; completed-with-no-audio
  reported as success; empty download accepted; truncated download accepted).
  **The sixth survived**, which is the interesting one — see below.
- **Cross-family review** (opencode / GLM-5-turbo, pre-PR): thread-safety, the poll
  loop, request serialisation and the panel all came back clean. One Critical and
  two Minors; the Critical was a false positive (below), both Minors fixed.

### The mutation that survived, and what I did about it

I mutated away the run-id staleness guard in `applyState` and **no test failed**. I
had just written a test that looked like it covered exactly that.

Reasoning it through rather than patching the test: **the guard is currently
unreachable.** `currentRun` only advances in `start()`; `start()` refuses while
`isBusy()`; `isBusy()` stays true until the worker's terminal state is *dispatched*;
and after that the worker posts nothing more. A stale post cannot coexist with a
newer run today.

So I deleted the test rather than ship something that only looks like coverage, and
documented the guard as deliberate defence-in-depth that becomes live the moment
`start()` is allowed to pre-empt a running job. The null-check beside it *is*
reachable and *is* tested (a run outliving its manager).

### The review's Critical, and why it wasn't one

> **CRITICAL** — `post()` overwrites the Authorization header with Content-Type;
> `withExtraHeaders` is a setter, not an appender. Every POST loses the API key.

Verified before acting, against JUCE's Linux, Windows and macOS implementations —
all three do `headers << extraHeaders`. It **appends**. No change was needed.

But the reviewer's supporting observation was correct and worth acting on: *nothing
covered it*, because every other client test runs without a key. Added tests that a
POST and a poll each carry `Authorization` alongside `Content-Type`, so the claim is
now permanently settled either way.

### Known limitations

- **Cover mode's `src_audio_path` is a path on the server, not an upload.** Fine for
  a local ACE-Step, silently broken against a remote one, and there is no upload
  endpoint to fall back on. The field is wired and validated; actually *choosing* a
  source needs local clips to exist, which is US-23.4/23.5.
- **Progress has no percentage** — see above. The bar is already there if ACE-Step
  ever reports one.
- **Cancellation is at poll granularity** (~2s), and a single hung request can still
  cost up to the probe timeout, as documented for US-23.2.
- **Clips are not yet playable or insertable** (US-23.4). They land in
  `~/.config/AutoMusicStudio/clips/run-N/`, which is provisional — US-23.5 owns
  cache location and management.
- **The demo used a stand-in server**, not real ACE-Step. It serves the envelope
  documented in `client.py` and real synthesised WAV audio, so parsing, downloading
  and file handling are exercised for real; but if the live server has drifted from
  what the Python client parses, both would be wrong together.
- **No preset save/recall** for generation parameters — not in this story.


frankbria:

One Critical (a false positive) and two Minors (both fixed). Full output below.

### The Critical, verified before acting

> `post()` overwrites the Authorization header with Content-Type. JUCE's `withExtraHeaders` is a setter, not an appender. Every POST loses the API key.

**Not true.** All three JUCE implementations append:

```cpp
// juce_Network_linux.cpp:122, juce_Network_windows.cpp:66, juce_Network_mac.mm
void withExtraHeaders (const String& extraHeaders)
{
    if (! headers.endsWithChar ('\n') && headers.isNotEmpty())
        headers << "\r\n";

    headers << extraHeaders;   // appends
}
```

No change made. **But the reviewer's supporting point was right and worth acting on**: it noted the bug would go unnoticed because every test runs with `auth=absent`. That coverage gap was real, so there are now tests asserting a POST and a poll each carry `Authorization` alongside `Content-Type`. The claim is settled permanently either way.

### The two Minors — both fixed

| Finding | Action |
| --- | --- |
| `downloadAudio` doesn't verify Content-Length, so a connection dropped mid-transfer moves a truncated file into place and reports success | **Fixed.** Compares against Content-Length when the server sends one and fails with got-vs-expected. Verified by mutation: removing the check fails the new test. This one mattered — a truncated WAV still opens. |
| Stub server double-counts an exact path match (once as path, once as route fragment), so `getRequestCountFor` returns inflated numbers | **Fixed.** Skips the fragment pass when it equals the exact path. Test-only, but a lying helper is how a future test gets written wrong. |

### Sections it found clean

Thread-safety and lifetime in `GenerationManager` (it independently walked the static-`runGeneration` / `RunContext`-by-value / `shared_ptr` stop-flag argument and the `PluginProcessor` member ordering), the poll loop, `AceStepClient` envelope and status handling, `GenerationRequest` serialisation including the seed-`0` case, and the panel's `refresh()` triggers.

### What the review did NOT catch — my own mutation pass did

I mutated away the run-id staleness guard in `applyState` and **no test failed**, including one I had just written that claimed to cover it. Rather than patch the test, I worked out that **the guard is currently unreachable** (`currentRun` only advances in `start()`; `start()` refuses while `isBusy()`; `isBusy()` stays true until the worker's terminal state is dispatched, after which it posts nothing more).

So the misleading test is deleted and the guard is documented as deliberate defence-in-depth that becomes live if `start()` ever pre-empts a running job. Details in the PR body.

---

Now I have the complete diff. Here is my analysis:

---

## 1. Thread-safety and lifetime in GenerationManager

**Clean.** The architecture is sound.

- `runGeneration` is a **static** function taking `RunContext` **by value**. Every piece of data the worker needs (request, url, key, runId, shared_ptr stop flag, WeakReference, raw queue pointer) is captured at enqueue time. The worker never dereferences the manager.
- `applyState`/`applyClips` hop to the message thread via `callOnMessageThread`, then check both `owner.get() == nullptr` (WeakReference invalidated by destructor) and `runId == self->currentRun` (superseded by a newer start). Double gating is correct.
- The destructor sets `activeControl->stopped = true` (atomic store) and calls `masterReference.clear()`, invalidating all WeakReferences. A worker that outlives its manager will either see the stop flag at its next check, or its posted message will be silently dropped.
- PluginProcessor member order (`backgroundQueue` → `properties` → `connectionManager` → `generationManager`) gives destruction order: generationManager first, backgroundQueue last. The queue outlives the manager, so the worker's `context.queue->isStopping()` check is always valid, and the queue's teardown joins the worker thread.
- `clips.clear()` in `start()` is safe: any pending `applyClips` from a previous run is guarded by the `runId` check.
- **No defect found.**

## 2. The poll loop

**Clean.**

- The 100ms-sliced sleep (`waited += 100; Thread::sleep(100)`) correctly bounds cancellation latency to ~100ms. `shouldStop()` checks both `control->stopped` (atomic) and `queue->isStopping()`.
- Cannot spin forever: teardown always sets `stopped = true`, and queue destruction joins the worker.
- A superseded run (second `start()`) is stopped because `start()` sets `activeControl->stopped = true` on the OLD control before creating a new one. The old worker's `shouldStop` lambda captures the old shared_ptr and sees the flag.
- No overall wall-clock timeout — documented as a design decision (demo doc: "per the correction posted on #317, no watchdog is warranted").
- **No defect found.**

## 3. AceStepClient submit/query/download

### **CRITICAL — `post()` overwrites the Authorization header with Content-Type**
**File:** `plugin/source/AceStepClient.cpp`, inside the anonymous `post()` function (~diff line 486–505)
**Scenario:** `configure(stream, endpoint, timeoutMs)` sets `extraHeaders = "Authorization: Bearer ..."`. Then `stream.withExtraHeaders("Content-Type: application/json")` **replaces** it entirely. JUCE's `withExtraHeaders` is a setter, not an appender.

Every `POST /release_task` and `POST /query_result` loses the API key. The connection probe (GET) is unaffected because it never calls `withExtraHeaders` a second time. So the user sees the connection go green, but generation fails with "API key rejected by the server" — a confusing mismatch.

**Why it wasn't caught:** The demo server and all tests run with `auth=absent`, so the lost header is never noticed.

**Fix:** combine headers before the single `withExtraHeaders` call:
```cpp
juce::String headers = "Content-Type: application/json";
if (endpoint.key.isNotEmpty())
    headers += "\r\nAuthorization: Bearer " + endpoint.key;
stream.withExtraHeaders (headers);
```

### **Minor — `downloadAudio` doesn't verify Content-Length**
**File:** `plugin/source/AceStepClient.cpp`, `downloadAudio()` (~diff line 689–776)

If the HTTP connection drops mid-transfer, the loop exits (`read() <= 0`), `temp.getSize()` is nonzero, and the truncated file is moved into place and reported as success. The code has the `.part`-file-and-move pattern to handle cancel/error, but not partial reads.

**Fix:** after the loop, if `stream.getTotalLength() > 0`, assert `temp.getSize() >= stream.getTotalLength()`.

### Envelope unwrapping, JSON-string result, status int/string, URL building, .part cleanup
All correct. The `unwrapEnvelope` function handles bare payloads and wrapped `{"data":...}`. The JSON-string `result` field is parsed via `rawResult.isString() ? juce::JSON::parse(...) : rawResult`. Status handles both int (0/1/2) and string ("completed"/"failed"/"pending"). URL building correctly prepends `baseUrl` for relative paths. `.part` cleanup covers cancel, empty response, and move failure.

## 4. GenerationRequest serialisation

**Clean.**

- Auto/Any/Random correctly omitted: `bpm > 0`, `key.isNotEmpty()`, `seed >= 0` guards match the sentinel values (`bpm = -1`, `key = ""`, `seed = -1`). Empty lyrics/language/instrumental/model are also omitted.
- Seed 0 preserved: `seed >= 0` is true for 0, so it's sent. The panel's `buildRequest()` uses `getLargeIntValue()` which returns 0 for "0", and the payload check `seed >= 0` passes.
- Cover mode: `task_type` and `src_audio_path` only sent when `mode == cover`. Text-to-music omits `task_type` (matching the Python client).
- No fields sent wrong.
- **No defect found.**

## 5. GenerationPanel

**Clean.**

- `refresh()` is called on every input that affects validity: `promptEditor.onTextChange`, `durationEditor.onTextChange`, `modeSelector.onChange`, plus `changeListenerCallback` from both the generation manager and connection manager, plus `timerCallback` at 2Hz when busy. Controls that don't affect validity (BPM, seed, key, quality, language, instrumental) correctly omit `onTextChange`/`onChange` — they don't need to re-evaluate Generate's enabled state.
- Controls cannot re-enable mid-run: `refresh()` checks `generation.isBusy()` which reads `state` on the message thread. State transitions only happen via `applyState` on the message thread. No race.
- `buildRequest()` reads directly from controls on the message thread. The click handler and `buildRequest()` both run on the message thread, so no drift.
- **No defect found.**

## 6. Tests

### **Minor — Stub server `pathCounts` double-counts exact matches**
**File:** `plugin/tests/StubAceStepServer.h` (~diff line 3276–3304)

The server increments `pathCounts[exactPath]` for every request, then also increments `pathCounts[fragment]` for matching routes. For a request to `/release_task` with a route for `/release_task`, the count becomes 2 instead of 1. Doesn't break any test because the "second start refused" test compares before/after counts (both inflated equally), and the "never submitted" test uses `>= 1`. But `getRequestCountFor` returns incorrect absolute numbers.

**Fix:** only count fragment matches, not the exact path — or don't double-count.

### Test quality

No vacuous or dangerously flaky tests found. The `expect(true, "survived ...")` in the lifetime test is a standard crash-detection pattern — its value is in the code that runs before it (destroying managers with a live run), not the assertion itself. The pump-based tests use generous timeouts (20–30s) against a loopback server that responds instantly, so they should be stable. The `parseTaskStatus` tests cover the pure-function layer that would catch envelope/JSON-string bugs even if the integration tests were mocked.

---

**Summary:** One critical defect (POST requests silently drop the Authorization header due to `withExtraHeaders` replacing rather than appending), one minor defect (download doesn't verify Content-Length against bytes read), and one minor test infrastructure issue (stub double-counting). The thread-safety architecture, poll loop, request serialisation, and panel sync are all correct.

</details>

Issue #319, opened 2026-07-23
US-23.5: Local Cache and File Management


frankbria:
## User Story

**As a** musician, **I want** my generated clips stored locally and managed efficiently, **so that** I can access past generations without regenerating.

## Description

The plugin maintains a local file cache for all generated clips. The cache is organized, browsable, and configurable. This keeps the plugin lightweight (~20MB binary) while audio data lives on disk.

## Functional Requirements

- Default cache directory: `~/ACEStepPlugin/cache/`
- Configurable cache path via plugin settings
- Cache organized by date and generation ID
- Cache browser in the plugin UI showing past generations with metadata (prompt, date, duration)
- Delete individual cached clips or clear entire cache
- Total cache size displayed in settings
- Cache is independent of the web platform — no account required for local-only usage

## Acceptance Criteria

- [ ] Generated clips appear in the cache directory after generation
- [ ] Cache browser lists clips with correct metadata
- [ ] Deleting a cached clip removes it from disk and the browser
- [ ] Custom cache path is respected after configuration change
- [ ] Cache size is reported accurately in settings

## Dependencies

**Blocked by:** US-23.4 (prior story in this stage). Part of Stage 23: VST3 Plugin — Core, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 23, US-23.5

PR #373, opened 2026-07-30
feat(plugin): US-23.4 Results Panel and Clip Insertion (#318)

### LINKED ISSUE #318: US-23.4: Results Panel and Clip Insertion
## User Story

**As a** musician, **I want** to preview generated clips and insert them into my DAW timeline, **so that** I can use AI-generated audio alongside my existing tracks.

## Description

After generation completes, the results panel shows waveform previews of the 2 generated clips. The musician can play/stop each clip, then insert it at the DAW playhead position or send it to a new track.

## Functional Requirements

- Waveform visualization for each of the 2 generated clips
- Play/Stop buttons per clip (audio routed through the DAW's audio engine)
- "Insert to Track" button — places the selected clip at the DAW playhead position on the plugin's track
- "Send to New Track" button — creates a new audio track in the DAW containing the clip
- Generation history: scrollable list of all generations in this session, with the ability to re-insert any past result
- Generated clips cached locally at `~/ACEStepPlugin/cache/` (configurable)
- Cache is browsable from the history panel

## Acceptance Criteria

- [ ] Both generated clips display waveform previews
- [ ] Play/Stop works for each clip within the DAW
- [ ] "Insert to Track" places audio at the correct playhead position
- [ ] History panel shows all generations from the current session
- [ ] Clips are persisted in the local cache directory and survive plugin close/reopen

## Dependencies

**Blocked by:** US-23.3 (prior story in this stage). Part of Stage 23: VST3 Plugin — Core, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 23, US-23.4

frankbria:
## US-23.4: Results Panel and Clip Insertion

Closes #318. Fourth story of Stage 23, on US-23.3 (#317, merged `d00a5a1`).

Generated clips get waveforms, playback through the host's output, and a drag handle
onto the timeline.

### Two acceptance criteria were amended — approved before implementation

The issue asked for **"Insert to Track places audio at the correct playhead position"**
and **"Send to New Track creates a new audio track in the DAW"**.

**VST3 provides no API for either.** A plugin can *read* the host transport
(`juce::AudioPlayHead`) but cannot create tracks or place regions on the timeline.
Verified against JUCE 8.0.9. ARA exists in JUCE but is for editing audio the host
already has, not introducing new material, and only in ARA-capable hosts.

Approved approach: **drag-and-drop export**, which is what commercial generative
plugins do and works in essentially any DAW.

| Original AC | Amended |
| --- | --- |
| Insert to Track places audio at the playhead | Clips are **draggable** onto any track; the panel **displays** the host playhead so a drop can be lined up |
| Send to New Track creates a new track | **Dropped** — dropping on empty space creates the track, and that is the DAW's business |

I'll update the issue text to match on merge. The other three criteria are unchanged
and met as written.

### What's here

| Path | Purpose |
| --- | --- |
| `source/ClipPlayer.*` | Preview playback mixed into `processBlock` |
| `source/ResultsPanel.*` | Clip rows with waveforms, play/stop, drag-out, history, playhead readout |

### The part that deserves scrutiny: audio in `processBlock`

This is the first story that puts audio in the callback, so the discipline is
explicit:

- `load()` opens and reads the file on the **message** thread and prepares the source
  there. `prepare()` sizes everything the audio thread touches.
- `addTo()` only pulls from an already-prepared source into an already-allocated
  buffer.
- The clip swap takes a `juce::SpinLock`, and the audio side uses **`tryEnter`**: if a
  load is mid-swap the block is skipped rather than blocked. A dropped preview buffer
  is inaudible; a priority inversion in the host's callback is not.
- Playback goes through the plugin's own output, **not** a second `AudioDeviceManager`
  opened behind the host's back — which is what the story means by "routed through
  the DAW's audio engine".

There's a test that hammers `load()` from another thread while `processBlock` runs,
and one that hands the processor a block larger than `prepareToPlay` promised.

### Verification

- **122 test groups** (18 new), zero warnings.
- **pluginval strictness 10**: exit 0, 25/25. That matters more than usual here —
  pluginval exercises the audio callback under a host.
- **VST3 6612KB** (up from 4948KB: `juce_audio_formats` and the thumbnail machinery
  are now linked). Still well inside the 25MB gate.
- **Cross-family review** (opencode / GLM-5-turbo, pre-PR): real-time safety,
  lifetime, `rebuildRows`, `toggle`, the playhead readout and the tests all came back
  clean, with its own walk of the lock discipline. Its one finding — repeated drag
  starts — I had already found and fixed independently.
- **Mutation check** — 5 mutations, **4 caught**:

| Mutation | Result |
| --- | --- |
| Oversized-block clamp removed | **SEGFAULT inside the test named for it** |
| `rebuildRows` always rebuilds | caught |
| History not accumulated | caught |
| Drag payload emptied | caught |
| Mono upmix removed | **survived** — see below |

### The mutation that survived

Removing the explicit mono→stereo upmix failed no test, including "a mono clip
reaches both output channels". That test passes because `prepare()` always sizes the
scratch buffer to two channels and `AudioFormatReaderSource` duplicates a mono file
across them itself — **my branch could never run**.

Unlike the run-id guard on #317 (kept as documented defence-in-depth), this one has
no plausible future trigger: I control the scratch sizing. So it is **deleted**, not
documented. The test stays, because "a mono clip reaches both channels" is the
behaviour that matters regardless of who performs the duplication.

### Also in here

- **The editor default size went from 720×520 to 760×720**, and the minimum height
  up with it. The US-23.1 default left Results squeezed to a single line once it had
  waveforms in it. Found by the demo, not by a test.

### Known limitations

- **AC 5 ("clips persist and survive close/reopen") is only half met by this story.**
  Clips persist on disk and are re-read, but there is no cache *browser*, no
  configurable path, no delete and no size readout — all of which the issue itself
  lists under US-23.5 (#319). What this adds is session history within a running
  plugin; reopening shows an empty panel until #319 loads past generations from disk.
  Flagged as **PARTIAL** in the demo rather than claimed.
- **The clip directory is still provisional** (`~/.config/AutoMusicStudio/clips/run-N/`).
  #319 owns making it configurable. Note the issue's suggested `~/ACEStepPlugin/cache/`
  would drop a *visible* directory in `$HOME` — the same problem fixed in US-23.2, so
  I would not adopt it as written.
- **A real OS drag is not exercised end to end.** It cannot be driven from a unit
  test; what is asserted is the exact payload handed to the drag service, and that it
  is a real existing file. Dropping into an actual DAW is part of the manual Reaper
  check in `plugin/README.md`.
- **History is unbounded** within a session. Fine for a session's worth of
  generations; #319's cache management is where a bound belongs.
- **Playback is preview-only**, not synchronised to the host transport. The story
  asks for preview, not for the plugin to act as a sampler.
- **A clip the format manager cannot read shows "loading waveform..." forever**
  rather than an error. Unreachable today — generated clips are always WAV.


frankbria:

**No Critical or Major defects.** One finding, which I had already found and fixed independently before reading the review.

### Sections it walked and found clean

| Area | Verdict |
| --- | --- |
| Real-time safety in `ClipPlayer::addTo` | Clean — it independently verified the `tryEnter` discipline, that the old source stays alive until `setSource(nullptr)` releases it, that no allocation happens, and that the oversized-block clamp holds |
| Lifetime | Clean — `ClipPlayer` on the processor outliving the editor is the right shape; `AudioThumbnail` teardown correct |
| `rebuildRows` | Clean — comparison correct, `OwnedArray` prevents leaks |
| `toggle` start/stop/switch | Clean for all call patterns |
| Playhead readout | Clean |
| Tests | "All tests are substantive, the concurrency test genuinely exercises the race, no flaky patterns" |

### The one finding

> `ClipRow::mouseDrag` calls `performExternalDragDropOfFiles()` on every drag event. `mouseDrag` fires continuously while the mouse is held down and moving — not once per gesture.

Correct, and already fixed in `4c7e4a3` — I suspected it while writing the code and specifically asked the reviewer to check it. Guarded with a per-gesture flag cleared on `mouseUp`, plus a 6px threshold so a shaky click does not become a drag. (The review called it Major in its detail section and Minor in its summary; either way it is fixed.)

### What the review did not catch — the mutation pass did

5 mutations, 4 caught. The headline: **removing the oversized-block clamp segfaulted inside the test named for it** — which is exactly what a real buffer overrun in `processBlock` would do to a host.

The survivor is the interesting one. Removing the explicit mono→stereo upmix failed **no** test, including "a mono clip reaches both output channels". That test passes because `prepare()` always sizes the scratch buffer to two channels and `AudioFormatReaderSource` duplicates a mono file across them itself — **my branch could never run**.

Deleted rather than documented (`0842148`). Unlike the run-id guard on #317, this one has no plausible future trigger since I control the scratch sizing. The test stays, because the behaviour it asserts is right regardless of who performs the duplication.

---

Let me read the rest of the diff and the relevant existing source files to verify context.Now let me check the JUCE documentation for `performExternalDragDropOfFiles` to verify the repeated-drag concern, and look at existing tests to understand the test harness pattern:Now I have a thorough understanding of the diff and the existing codebase. Here's the review:

---

## PR Review: US-23.4 Results Panel and Clip Insertion

### 1. REAL-TIME SAFETY — `ClipPlayer::addTo` (audio thread path)

**Clean.** This is well-designed:

- `ScopedTryLockType` on `sourceLock` — correct, never blocks.
- `scratch` buffer pre-allocated in `prepare()` — no allocation in the callback.
- `jmin(numSamples, scratch.getNumSamples())` gracefully truncates if the host exceeds the promised block size — safe, just silent for the excess samples.
- `transport.getNextAudioBlock()` reads from an already-prepared `AudioFormatReaderSource` — it does I/O from a memory buffer (the whole file was read by the reader on load), no disk access on the audio thread.
- The mono-to-stereo upmix at the end is correct and necessary.

**One concern — race between `load()` and `addTo()`:**

`load()` (message thread) does:
```cpp
const juce::SpinLock::ScopedLockType lock (sourceLock);
transport.setSource(nullptr);
readerSource = std::move(newSource);
transport.setSource(readerSource.get(), ...);
```

`addTo()` (audio thread) does:
```cpp
const juce::SpinLock::ScopedTryLockType lock (sourceLock);
if (!lock.isLocked() || readerSource == nullptr) return;
```

**This is correct.** If `load()` holds the lock, `addTo()` bails and returns silence for one block. The only dangerous scenario would be if `readerSource` was being *deleted* while the audio thread reads it, but the unique_ptr swap happens under the lock and `addTo()` checks `readerSource == nullptr` after acquiring the lock. The old source is kept alive until `setSource(nullptr)` releases it. No defect here.

### 2. LIFETIME

**Clean.** `ClipPlayer` lives on `PluginProcessor`, `ResultsPanel` on `PluginEditor`. `ClipRow` holds `ClipPlayer&` (not owning). When the editor is destroyed mid-playback, `ClipPlayer` survives on the processor — the test "closing the window while a clip plays does not stop it or crash" confirms this design.

`AudioThumbnail` — `ClipRow` calls `thumbnail.removeChangeListener(this)` in its destructor, and `AudioThumbnail` manages its own background thread via `AudioThumbnailCache`. The `thumbnailCache` is on the `ResultsPanel` (stack member, destroyed with it). `AudioThumbnail` objects are owned by `ClipRow`s in the `OwnedArray<ClipRow> clipRows` — destroyed before `thumbnailCache`. This is correct: JUCE thumbnails detach from the cache on destruction.

`ClipRow::file` is a `juce::File` (value type — path string). If the file is deleted from disk, `mouseDrag` checks `file.existsAsFile()` before dragging, and playback simply ends when the source runs out. No dangling pointer possible.

### 3. RESULTS PANEL `rebuildRows` — MEMORY CORRECTNESS

**Clean.** The comparison logic (line 891 of diff):
```cpp
auto sameAsShown = clips.size() == clipRows.size();
if (sameAsShown) { for (...) compare file paths }
if (!sameAsShown) { clipRows.clear(); rebuild }
```

`clipRows` is an `OwnedArray<ClipRow>` — `clear()` deletes all elements. No leak. The comparison is by `juce::File` equality (path comparison), which is correct for this use case.

**History is unbounded** — `history.add(clips[i])` grows forever. However, `juce::File` is a small value type (~32 bytes), and a "session" is typically short. This is a **Minor** concern at most, not a defect. The demo doc acknowledges this is US-23.5 territory.

### 4. DRAG — REPEATED `performExternalDragDropOfFiles` CALLS

**Major defect.** `ClipRow::mouseDrag` calls `juce::DragAndDropContainer::performExternalDragDropOfFiles()` on every drag event. `mouseDrag` fires continuously while the mouse is held down and moving — not once per gesture.

**What actually happens:** Looking at JUCE's implementation, `performExternalDragDropOfFiles` on Linux uses X11 drag-and-drop. JUCE internally tracks whether a drag is already in progress and won't start a second one. On macOS it goes through `NSPasteboard`/`NSDraggingItem`. On Windows it's OLE drag. All three OS-level drag mechanisms are inherently one-shot — once started, the OS takes over the mouse until the drop or cancel.

**Verdict:** This is safe in practice because the OS drag-and-drop system won't start a second drag. The redundant calls are wasted cycles but not a bug. **Downgrade to Minor** — the intent would be clearer with a guard, but it works correctly on all platforms.

### 5. `ClipPlayer::toggle` LOGIC

```cpp
bool ClipPlayer::toggle(const juce::File& file) {
    if (isPlaying() && getCurrentFile() == file) { stop(); return true; }
    if (getCurrentFile() != file && !load(file)) return false;
    play();
    return true;
}
```

**One subtle issue (Minor):** `play()` does:
```cpp
if (transport.hasStreamFinished() || transport.getCurrentPosition() >= transport.getLengthInSeconds())
    transport.setPosition(0.0);
transport.start();
```

But between `load(file)` and `play()`, there's no lock held. `load()` sets position to 0, but another thread could theoretically call `stop()` between them (though in practice `toggle` is only called from the UI). Not a real race in practice, and the code is correct for all call patterns.

**Clean overall.**

### 6. `playhead` READOUT

```cpp
if (auto* playHead = processor.getPlayHead())
    if (const auto position = playHead->getPosition())
        if (const auto seconds = position->getTimeInSeconds())
```

The three nested optional-pointer checks are correct. `getTimeInSeconds()` returns `std::optional<double>`. The conversion to MM:SS.mmm at line 873–875 has a minor display issue: if `*seconds` is negative (rewound past zero in some hosts), `(int) *seconds` truncates toward zero, giving wrong display. **Negligible** — JUCE hosts report non-negative time in practice.

### 7. TESTS

**All substantive.** Key observations:

- **Concurrency test ("processBlock stays safe while clips are swapped")** — This genuinely exercises the race. It spins a loader thread doing `load()` + `play()` in a tight loop while the main thread calls `processBlock` 400 times. The `std::atomic<bool> stop` synchronisation is correct. The test is not vacuous — it would catch a deadlock or crash from a blocking lock.

- **"playing survives prepareToPlay being called again"** — Correctly tests a real scenario (sample rate change). The `prepare()` method resizes `scratch` and re-prepares `transport` while potentially playing. The assertion `expect(true, ...)` is weak — it only checks no crash, not that audio continues correctly. **Minor** — the test does prove the crash-free guarantee which is the important part.

- **"a host block larger than promised does not overrun"** — Good. It prepares with 256 samples, then calls `processBlock` with 2048. Asserts audio in the first 256, and survival. **Correct.**

- **No flaky tests observed.** The `pumpUntil` with timeouts is the right pattern for async thumbnail loading. The 10-second timeout for waveform loading is generous.

### 8. `GenerationManager::setClipsForTesting`

Not a defect, but worth noting: this method is on a `final` class that's production code. It directly mutates `clips`, `state`, and `statusMessage` then broadcasts. It's clearly marked as a test seam with naming that prevents accidental production use. Acceptable.

---

## Summary

| # | File | Line (diff) | Severity | Finding |
|---|---|---|---|---|
| 1 | — | — | **Clean** | Real-time safety in `ClipPlayer::addTo` — correct tryLock, no allocation, oversized-block safe, mono upmix correct |
| 2 | — | — | **Clean** | Lifetime: ClipPlayer on processor survives editor, AudioThumbnail teardown correct, file deletion handled |
| 3 | — | — | **Clean** | `rebuildRows` comparison correct, `OwnedArray` prevents leaks; history is unbounded but acknowledged for US-23.5 |
| 4 | `ResultsPanel.cpp` | ~122 | **Minor** | `mouseDrag` calls `performExternalDragDropOfFiles` on every drag event; redundant but safe because OS drag is one-shot and JUCE/OS won't start a second. A `isDragging` guard would be tidier. |
| 5 | — | — | **Clean** | `toggle` logic correct for all call patterns |
| 6 | — | — | **Clean** | Playhead readout correct; negative time edge case negligible |
| 7 | — | — | **Clean** | All tests are substantive, the concurrency test genuinely exercises the race, no flaky patterns |

**No Critical or Major defects. One Minor finding (redundant drag starts, harmless).** The real-time audio path is well-designed and the lock discipline is correct. Ready to merge.

</details>

frankbria:
## Review round: one real finding fixed, one Major verified as a false positive

### Fixed — disk I/O in the audio callback (`e52afb8`)

> `load()` opens the file and prepares the source, but doesn't read the audio data
> into memory. The data is still read lazily, block by block, during playback.

Correct, and a fair catch on my own framing: `ClipPlayer.h` claimed `processBlock`
"only ever pulls from an already-prepared source", which was overstating it. An
unbuffered read in the host's callback is exactly the class of stall the
SpinLock/`tryEnter` design exists to avoid, so it was inconsistent to guard one and
not the other.

Now wrapped in a `juce::BufferingAudioSource` fed by an owned `TimeSliceThread`, with
teardown ordered so the buffering source releases before the reader it points at.

### Not fixed — the Major finding is a false positive

> `prepare()` sizes `scratch` with `avoidReallocating = true`, so a re-prepare with a
> **larger** block size never grows it; every subsequent preview block is then clamped
> to the original (smaller) size and preview plays at the wrong speed with gaps.

I applied this fix first and verified afterwards, which was the wrong order — the
verification says the finding is wrong. `juce_AudioSampleBuffer.h:432-444`:

```cpp
else   // keepExistingContent == false, which is our call
{
    if (avoidReallocating && allocatedBytes >= newTotalBytes)
    {
        if (clearExtraSpace || isClear)
            allocatedData.clear (newTotalBytes);
    }
    else
    {
        allocatedBytes = newTotalBytes;
        allocatedData.allocate (newTotalBytes, clearExtraSpace || isClear);
        ...
    }
    ...
}
...
size = newNumSamples;     // unconditional
```

`avoidReallocating` only skips the *reallocation* when the existing allocation is
already large enough. Growing past it still reallocates, and `size` — what
`getNumSamples()` returns, and therefore what the `jmin` clamp uses — is updated
either way.

Confirmed by mutation: restoring `avoidReallocating = true` failed **no** test,
because there was no defect to fail. The flag is back as it was, with a comment
citing the line so the next reader doesn't re-derive this.

### Kept the test the finding prompted

The old re-prepare test went 512 → **256** (smaller), so it structurally could not
have covered growth even if the bug had been real. It now re-prepares in both
directions and asserts the back half of a grown block is not silent. Worth having
regardless of who was right.

PR #374, opened 2026-07-30
feat(plugin): US-23.5 Local Cache and File Management (#319)

### LINKED ISSUE #319: US-23.5: Local Cache and File Management

frankbria:
## US-23.5: Local Cache and File Management

Closes #319. **Last story of Stage 23**, on US-23.4 (#318, merged).

Generated clips become a browsable, configurable, deletable cache that survives
closing the plugin.

### What's here

| Path | Purpose |
| --- | --- |
| `source/ClipCache.*` | Resolve the directory, enumerate runs with metadata, report size, delete one or all |
| `source/ResultsPanel.*` | "Past generations" browser: list, editable cache path, Delete, Clear cache, total size |
| `source/GenerationManager.*` | Writes a `meta.json` sidecar per run; run directories are timestamped |

### Acceptance criteria

| # | Criterion | Evidence |
| --- | --- | --- |
| 1 | Clips appear in the cache directory after generation | Two real generations on disk in the *configured* directory |
| 2 | Browser lists clips with correct metadata | Screenshot after **closing and reopening** the plugin: prompt, date, duration, clip count, size, newest first |
| 3 | Deleting removes from disk and the browser | Asserted at both the cache and panel level; deleting the generation that is playing stops it first |
| 4 | Custom cache path respected after configuration change | Asserted including across a fresh cache over the same settings, and typed through the panel |
| 5 | Cache size reported accurately | Compared against a real byte count, not a fixture constant |

This also completes **US-23.4's AC 5**, which that story could only mark PARTIAL.

### Decisions worth knowing

- **The default is NOT `~/ACEStepPlugin/cache/`** as the issue specifies. JUCE
  resolves a bare folder name under `$HOME` on Linux, so that drops a *visible*
  directory into the user's home — the same bug fixed for the settings file in
  US-23.2. Default is the platform cache tree (`~/.cache/AutoMusicStudio/clips`),
  which is also where regenerable data belongs. Still configurable.
- **Metadata is a `meta.json` sidecar**, because a WAV header has no prompt. A run
  whose sidecar is missing or corrupt **still lists**, labelled
  `(no description recorded)` — losing a whole generation because its metadata went
  missing would be worse than showing it without a prompt.
- **Run directories are timestamped**, not `run-N`. The run counter restarts with the
  plugin, so a bare index would have had the first generation of a new session
  overwrite the first of the last one.
- **The browser lives in the Results panel**, not a fourth top-level panel — it is
  the same list of clips, just historical.

### The data-loss path this PR closes

`deleteEntry`'s only protection was `isAChildOf(getDirectory())`. But **the cache root
is a path the user types.** Point it at `~/Music` and every album folder is a child of
it — so "Clear cache" would have recursively deleted the user's music library.
Containment is no protection when the container itself is user-chosen.

Deletion now *also* requires the directory to be named the way this plugin names its
runs. Anything else still lists — browsing a folder read-only is harmless — but cannot
be deleted. Covered by a test that aims the cache at a mock music library and asserts
both `deleteEntry` and `clearAll` refuse, and that the audio survives.

Found by reasoning through the guard, not by a test failing.

### The bug the demo found: the test suite was deleting the user's real cache

The first attempt at the demo generated twice, then showed **"0 generations - 0 bytes"**
on reopen. The clips were genuinely gone.

The generation tests' cleanup helper called `deleteRecursively()` on
`ClipCache::getDefaultDirectory()` — the **user's real cache**. It had been doing that
since US-23.3 wrote its first clips; nothing failed because nothing was checking. Three
`ResultsPanel` tests were writing there too.

All of them now use throwaway directories. Verified with a **canary** rather than by
inspection: a marker file in the real cache and a real settings file both survive a
full suite run.

### Verification

- **143 test groups** (25 new), zero warnings.
- **pluginval strictness 10**: exit 0, 25/25. VST3 6756KB.
- **Mutation check** — 4 mutations on the cache, 2 caught immediately, and the two
  survivors were the interesting part:

| Mutation | Result |
| --- | --- |
| Delete guard removed | caught |
| Size reports zero | caught |
| Newest-first sort disabled | **survived** — test weakness, fixed (below) |
| Runs without a sidecar dropped | **survived** — my mutant was ineffective (below) |

- **Cross-family review** (opencode / GLM-5-turbo, pre-PR): `ClipCache` correctness,
  threading, and the `ListBox` row-indexing all clean. Two Majors and three Minors —
  see below.

### The mutation that revealed a weak test

Disabling the newest-first sort failed **no** test. With only two runs, directory
iteration order — which is unspecified — can hand them back already sorted by luck, so
the test was passing on chance rather than on the sort.

Now five runs, asserting the real invariant (non-increasing creation time across the
whole list) plus the exact expected order. Accidental pass drops from a coin flip to
1-in-120. Verified: the same mutation now fails it with 4 distinct assertions.

The other survivor was **my own bad mutant** — the early return I inserted still left
the clip list populated, so the entry was listed anyway. Not a coverage gap; the
behaviour is covered by "a run without a sidecar still lists".

### Review findings

| Finding | Action |
| --- | --- |
| **Major** — `deleteEntry` follows symlinks | **False positive**, verified: `juce::File::deleteRecursively` takes `bool followSymlinks = false` (`juce_File.h:508`) and we use the default |
| **Major** — `ResultsPanelTests` write to the user's real cache | **Already fixed** in `c4dc1b0`, found independently via the demo |
| **Minor** — a typo'd cache path was persisted before validation | **Fixed** — refuses a path that is a file or cannot be created, and restores the previous one |
| **Minor** — run directory collision if the plugin restarts inside the same second | **Fixed** — takes the next free suffix |
| **Minor** — the sort test depends on filesystem timestamp resolution | **Not applicable** as written: entries carry `createdAtMs` from the sidecar (millisecond, monotonic), and the fixture writes metadata. Only sidecar-less runs fall back to the directory timestamp |

### Known limitations

- **The cache is unbounded.** Nothing evicts old generations; the user clears them. A
  size cap or age-based eviction is the obvious next step; the issue asks for delete
  and clear, which is what is here.
- **No confirmation on "Clear cache"** — it deletes every generation immediately. A
  modal in a plugin window is its own problem in hosts, and the issue does not ask for
  one, but it is a sharp edge worth naming.
- **The browser lists runs, not individual clips.** Deleting removes a whole
  generation, since that is the unit the plugin creates and the unit the metadata
  describes. The issue's "delete individual cached clips" would leave a generation
  that no longer matches its own metadata.
- **Runs from US-23.3/23.4 predate the sidecar** and list as `(no description
  recorded)` rather than being hidden.
- **Changing the cache path does not move existing clips.** They stay put and the
  browser shows the new location. Moving files the user did not ask to move seemed
  worse than the surprise.
- **`listEntries`/`getTotalSizeInBytes` walk the filesystem on the message thread.**
  Fine for a session's worth of generations; a very large cache on a slow filesystem
  would be felt. The review agreed this is not a practical risk at expected sizes.


frankbria:

Two Majors and three Minors. One Major was a false positive, the other I had already fixed independently; two Minors fixed, one not applicable.

| Finding | Action |
| --- | --- |
| **Major** — `deleteEntry` follows symlinks | **False positive**, verified before acting: `juce::File::deleteRecursively` takes `bool followSymlinks = false` (`juce_File.h:508`) and we use the default, so a symlink inside the cache is unlinked rather than followed |
| **Major** — `ResultsPanelTests` write to the user's real cache | **Already fixed** in `c4dc1b0` — found independently while chasing a demo that showed `0 generations` after a test run had silently deleted them |
| **Minor** — a typo'd cache path was persisted before validation | **Fixed** — refuses a path that is a file or cannot be created, says so, and restores the previous one |
| **Minor** — run directory collision if the plugin restarts inside the same second | **Fixed** — takes the next free suffix instead of overwriting |
| **Minor** — the sort test depends on filesystem timestamp resolution | **Not applicable as written** — entries carry `createdAtMs` from the sidecar (millisecond, monotonic) and the fixture writes metadata; only sidecar-less runs fall back to the directory timestamp |

### Sections it found clean

`ClipCache` directory resolution, enumeration and sort, size accounting, corrupt-sidecar degradation, `writeMetadata` JSON round-tripping; threading; and — usefully — it independently checked every `ListBox` row access against `cacheEntries` mutating underneath it and confirmed the bounds guards and `updateContent()` selection clamping.

### The thing neither of us found by review — the mutation pass did

Disabling the newest-first sort comparator failed **no** test. With only two runs, directory iteration order is unspecified and can hand them back already sorted by luck, so the test was passing on chance rather than on the sort. Now five runs asserting the real invariant; accidental pass drops from a coin flip to 1-in-120, and the same mutation now fails it with 4 distinct assertions.

A second mutation in that pass survived because **my mutant was badly constructed**, not because of a coverage gap — the early return I inserted still left the clip list populated, so the entry was listed anyway. Recording that rather than quietly dropping it.

### And the one I found by reading rather than testing

`deleteEntry`'s only protection was `isAChildOf(getDirectory())`. The cache root is a path the **user types** — aim it at `~/Music` and every album folder is a child of it, so "Clear cache" would have recursively deleted their music library. Containment is no protection when the container itself is user-chosen. Deletion now also requires the directory to be named the way this plugin names its runs; anything else lists read-only.

---

## Code Review: US-23.4/23.5 — Results Panel and Clip Cache

### 1. ClipCache correctness

**Clean.** Directory resolution (configured vs default via `cachePathKey`), `listEntries` enumeration + newest-first sort, `sizeOfDirectory` recursive accounting, `readEntry` handling of missing/corrupt `meta.json` (gracefully degrades to no-metadata rather than hiding the run), and `writeMetadata` JSON round-tripping (quotes, backslash, newline covered by test) are all correct.

### 2. Safety of `deleteEntry` / `clearAll`

**Major — `deleteEntry` follows symlinks** (`ClipCache.cpp:~183`)

`isAChildOf(getDirectory())` checks the **logical** path. On Linux/macOS, if a user (or any process) creates a symlink *inside* the cache that targets an external directory containing `.wav` files:

```
~/.cache/AutoMusicStudio/clips/evil → /home/user/important-project-with-wavs/
```

Then `entry.directory.isDirectory()` returns true (follows the symlink), `isAChildOf` returns true (the symlink path itself is inside the cache), and `deleteRecursively()` traverses into the target and deletes its contents. This is the classic `rmdir`-via-symlink class of attack, though here the threat is accidental (a symlink created by another tool or script inside the cache dir) rather than malicious.

`clearAll()` inherits the same risk since it calls `deleteEntry()` per entry, and `listEntries()` follows symlinks in its directory iterator.

**Fix:** Before `deleteRecursively()`, add a real-path check:

```cpp
// After the isAChildOf check:
#if JUCE_LINUX || JUCE_MAC || JUCE_BSD
if (entry.directory.getLinkedTarget() != entry.directory)
    return false;  // refuse to follow symlinks
#endif
```

Or more portably: `juce::File::getNativePath()` doesn't resolve symlinks, but you can compare `getFullPathName()` against a fully-resolved version. A simpler portable approach: `entry.directory != entry.directory.getAbsoluteFile()` won't work either since JUCE normalizes but doesn't resolve symlinks. The platform-specific `getLinkedTarget()` is the cleanest.

**The `isAChildOf` check is otherwise sufficient** for the other scenarios you asked about:
- **`..` components**: JUCE normalizes paths in `getFullPathName()`, so `/cache/../etc` resolves to `/etc` and `isAChildOf` rejects it.
- **Path change between listing and deleting**: `deleteEntry` calls `getDirectory()` fresh, so stale entries from a prior path are correctly rejected.
- **Cache directory itself as an entry**: `listEntries()` enumerates *subdirectories* only; the root can't be an entry. Even if constructed manually, `isAChildOf(parent, parent)` returns false in JUCE (the paths are equal, not one strictly inside the other).

### 3. User-typed cache path data loss

**No direct data loss from the path mechanism.** `isAChildOf` correctly bounds deletes to whatever directory is configured. The user *can* point the cache at `/home/user/documents` and then `Clear cache` would delete subdirectories containing `.wav` files — but that's user-directed action on a path the user chose, not a silent escape.

**Minor — bad path persisted without validation** (`ResultsPanel.cpp:~160`): `commitCachePath()` calls `cache.setDirectory(wanted)` before any validity check. A typo'd path gets saved. `hasProblem()` catches it on display, but the bad value persists across restarts. Fix: check `hasProblem()` before calling `setDirectory`, or validate the path is a writable directory first.

### 4. Threading

**Clean.**

- `listEntries()`/`getTotalSizeInBytes()` are synchronous on the message thread. For typical usage (tens of runs, each a few MB), this is <10ms on local storage. A network-mounted cache could be slower, but that's an edge case and not an audio-thread concern.
- The audio thread (`processBlock`) only calls `ClipPlayer::addTo()`, which uses `ScopedTryLockType` (non-blocking spinlock try). Never allocates, never blocks.
- `ClipPlayer::getCurrentFile()` and `isPlaying()` take the spinlock with `ScopedLockType` (blocking), but they're only called from message-thread contexts (`ClipRow::timerCallback`, `deleteSelectedEntry`). The audio thread never calls them.

### 5. ResultsPanel integration

**Clean.** All ListBox row access is guarded with `isPositiveAndBelow(row, cacheEntries.size())` in both `paintListBoxItem` and `listBoxItemDoubleClicked`. `deleteSelectedEntry` re-checks bounds after `getSelectedRow()`. JUCE's `ListBox::updateContent()` clamps the selection, so a stale index after `refreshCache()` is handled before anyone reads it. `cacheEntries` is only mutated on the message thread (in `refreshCache`), and paint callbacks are also message-thread, so no concurrent mutation.

### 6. GenerationManager timestamp collision

**Minor — run directory name collision after restart** (`GenerationManager.cpp:~107`)

```cpp
const auto directory = context.clipDirectory.getChildFile(
    juce::Time::getCurrentTime().formatted("%Y%m%d-%H%M%S")
        + "-run" + juce::String(context.runId));
```

`currentRun` resets to 0 on every plugin instantiation. If a generation completes in the same second as a prior run (from a previous plugin session) with the same `runId`, the directory name collides. `createDirectory()` silently succeeds (directory already exists), and clips from both sessions mix in the same folder.

Probability is low (requires restart + completion in the same second + same run ID), but non-zero, especially with two plugin instances sharing a cache.

**Fix:** Use millisecond-resolution timestamp (`"%Y%m%d-%H%M%S-fff"`) or check for existence and append a suffix.

### 7. Tests

**Major — ResultsPanelTests write to the user's real cache** (`ResultsPanelTests.cpp:~196, ~232, ~267`)

Three tests create run directories directly in `ClipCache::getDefaultDirectory()`:

- *"AC: the browser lists past generations with their metadata"* (line ~196)
- *"AC: deleting from the browser removes it from disk and the list"* (line ~232)
- *"deleting the generation that is playing stops it first"* (line ~267)

Each calls `run.deleteRecursively()` in cleanup, but if the test crashes before that line, garbage is left in the user's real `~/.cache/AutoMusicStudio/clips/`. This is exactly the bug the demo found and fixed in `GenerationManagerTests` — the same `ScopedClipCleanup` pattern (temp dir + `cachePathKey` override) was applied there and in `GenerationPanelTests` but was **not** applied in `ResultsPanelTests`.

The root cause: `ResultsPanelTests::Harness::makeSettings` does not set `ClipCache::cachePathKey`, so the cache defaults to the real one.

**Fix:** Set `cachePathKey` in the harness settings to a temp directory (same pattern as the other two test files), and create the test runs there instead of in `getDefaultDirectory()`.

**Minor — timestamp-dependent sort test** (`ClipCacheTests.cpp:~138`): *"runs come back newest first"* uses `Thread::sleep(25)` between creating five runs and relies on second-resolution filesystem timestamps. On FAT32, SMB, or other coarse-timestamp filesystems, all five could get the same timestamp and the sort would be unstable. Fix: use `juce::Time::getCurrentTime().toMilliseconds()` to write the `meta.json` `createdAtMs` field directly with a guaranteed-increasing value, then assert on that instead of `entry.created`.

All other tests are sound — the `ScopedCache`, `ScopedClips`, and `ScopedWav` fixtures use random temp directories, the concurrent-load test genuinely exercises the lock contention, and the cache-tests' `ScopedClipCleanup` in the other files correctly redirects to temp dirs. The `expect(true, "survived...")` idiom in the threading test is intentional (it's a crash-or-pass test; the assertion is the label in test output, not a check).

</details>

PR #386, opened 2026-08-02
feat(plugin): US-24.1 DAW Tempo Sync and Tempo-Matched Insertion (#320)

### LINKED ISSUE #320: US-24.1: DAW Tempo and Key Sync
## User Story

**As a** musician, **I want** the plugin to automatically detect my DAW's BPM and key, **so that** generated music matches my project without manual configuration.

## Description

The plugin reads the host DAW's tempo and key signature and auto-populates the generation fields. Generated audio is time-stretched or pitch-shifted if needed to match the DAW project exactly. This removes friction and ensures every generation is immediately usable in context.

## Functional Requirements

- Read host BPM via JUCE's `getPlayHead()->getPosition()` API
- Read host key signature if available
- Auto-populate BPM and Key fields in the generation panel (overridable by the musician)
- "Sync" indicator shows when fields are auto-populated vs. manually set
- If DAW tempo changes during a session, the fields update accordingly
- Generated audio is time-stretched to match host BPM if the generation result differs slightly

## Acceptance Criteria

- [ ] Opening the plugin in a 120 BPM project auto-fills the BPM field with 120
- [ ] Changing the DAW tempo updates the plugin's BPM field
- [ ] Generated clip at 118 BPM is time-stretched to 120 BPM on insertion
- [ ] Manual BPM override disables auto-sync with a visual indicator

## Dependencies

**Blocked by:** Stage 23. First story of Stage 24.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 24, US-24.1

frankbria:
Implements **US-24.1 — DAW Tempo and Key Sync** (#320).

The generation panel now follows the host's tempo, and a clip generated at a different
BPM is stretched to the project tempo when it is dragged into the timeline.

---

| Criterion | Evidence |
| --- | --- |
| Opening the plugin in a 120 BPM project auto-fills BPM with 120 | `GenerationPanel / AC: opening the plugin in a 120 BPM project auto-fills BPM with 120` — asserts the field reads `"120"` **at construction** (not after a timer tick), that the indicator names the tempo, and that `buildRequest().bpm == 120` so it actually reaches the server |
| Changing the DAW tempo updates the plugin's BPM field | `GenerationPanel / AC: changing the DAW tempo updates the plugin's BPM field` — host moves 120 → 90, field follows |
| Generated clip at 118 BPM is time-stretched to 120 BPM on insertion | `ResultsPanel / AC: a 118 BPM clip is dropped into a 120 BPM project already stretched` — the drag payload is a *different, real* file at the expected path and `44100/rate` samples long, read back through an independent `AudioFormatReader`. Plus the end-to-end measurement below |
| Manual BPM override disables auto-sync with a visual indicator | `GenerationPanel / AC: a manual BPM override disables auto-sync, with a visual indicator` — typing 100 clears the sync flag, changes the indicator to say "manual", and a subsequent host move to 140 **cannot** overwrite the field |

### End-to-end measurement of the stretch, by a tool that is not our code

A real 12s / 118 BPM track (220 Hz tone + a click on every beat) was stretched through
the shipped `TimeStretch::processFile` and measured with **librosa**:

```
source (118 BPM)        12.000s   tempo  117.45 BPM   tone  220.72 Hz
stretched -> 120 BPM    11.800s   tempo  120.19 BPM   tone  220.72 Hz

length ratio     1.01695   (expected 1.01695)
tempo measured   120.19 BPM   (target 120)
pitch drift      +0.0 cents   (a resample would be +29.1)
```

Full suite: **all plugin unit tests pass**; VST3 links at 6.7 MB against CI's 25 MB cap.

---

## Notable decisions

**WSOLA, not a resample.** `juce::LagrangeInterpolator` would have been a third of the
code, but resampling changes pitch with speed: 118→120 is a 1.7% rate change and so a
29-cent detune. Audio that arrives out of tune with the project defeats the purpose of
syncing to it, so `TimeStretch` is waveform-similarity overlap-add — pitch-transparent,
no new dependency. The measurement above is the proof.

**One reader of the play head.** `getPlayHead()` is an audio-thread API.
`ResultsPanel::timerCallback` was already calling it from the *message* thread for its
playhead readout — a live data race on `main`. Rather than adding a second racy reader,
`HostSync` samples it once in `processBlock` and publishes lock-free, and the results
panel is switched over. **This PR removes a race rather than adding one.**

**Bug found and fixed while wiring this up.** Tempo-matched copies were first written
beside the clip. `ClipCache::readEntry` lists `*.wav` per run directory
non-recursively, so they were counted as *extra clips of that generation* — a two-clip
run listed four, and would have restored four from the cache. They now live in a
`tempo-match/` child directory, which that scan skips but the size accounting (which
recurses) and `deleteEntry` (which is `deleteRecursively`) both still cover. Regression
test: `ResultsPanel / tempo-matched copies do not show up as extra clips in the cache
browser`, verified by mutation — reverting to sibling naming makes it fail with
`Expected value: 2, Actual value: 4`.

**A real race in the sync loop, caught by a test.** `TextEditor::setText(…, true)`
delivers `onTextChange` *asynchronously* (it posts a command message), so between a
user's first keystroke and the callback that clears the sync flag there was a window
where the 2 Hz tick could overwrite what they had just typed. `applyHostTempo` now
declines while the BPM editor has keyboard focus.

---

## Known limitations

1. **Host key signature is not synced, and cannot be.**
   `juce::AudioPlayHead::PositionInfo` carries tempo, time signature, PPQ and the
   transport flags — and no key signature. The only key-signature channel in JUCE is
   ARA (`kARAContentTypeKeySignatures`), a separately licensed Celemony SDK that a
   handful of hosts implement. It is unreachable through VST3, AU or Standalone.
   The issue's functional requirements say "read host key signature **if available**";
   it is not. The **Key** field stays manual and the indicator never implies otherwise.
   **All four acceptance criteria on #320 are BPM-only and all are met.**

2. **Clip tempo is the *requested* BPM, not analysis of the returned audio.** A
   generation that left BPM on *Auto* has no known tempo, so nothing is stretched —
   guessing would be worse than leaving it alone. Beat-detecting a finished clip is a
   separate story.

3. **Not verified in a real DAW.** No DAW is installed in this environment, and no
   unit test can make a real host publish a real tempo. The README's manual check now
   has an explicit step for it. Everything below the host boundary is covered by tests
   driving a `juce::AudioPlayHead` double.

4. Third-party review: opencode (GLM) cross-family pass; findings triaged in a comment
   below.

Closes #320

### DIFF
--- a/plugin/CMakeLists.txt
+++ b/plugin/CMakeLists.txt
@@ -44,9 +44,11 @@ set(ACEMUSIC_SOURCES
     source/GenerationManager.cpp
     source/GenerationPanel.cpp
     source/GenerationRequest.cpp
+    source/HostSync.cpp
     source/PluginProcessor.cpp
     source/PluginEditor.cpp
-    source/ResultsPanel.cpp)
+    source/ResultsPanel.cpp
+    source/TimeStretch.cpp)
 
 # ---------------------------------------------------------------------------
 # Plugin
@@ -111,8 +113,10 @@ target_sources(AceMusicPluginTests PRIVATE
     tests/GenerationManagerTests.cpp
     tests/GenerationPanelTests.cpp
     tests/GenerationRequestTests.cpp
+    tests/HostSyncTests.cpp
     tests/PluginProcessorTests.cpp
-    tests/ResultsPanelTests.cpp)
+    tests/ResultsPanelTests.cpp
+    tests/TimeStretchTests.cpp)
 
 target_include_directories(AceMusicPluginTests PRIVATE source tests)
 
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -2,17 +2,16 @@
 
 JUCE-based plugin that brings ACE-Step generation into a DAW.
 
-Stage 23 progress:
+Progress:
 
 | Story | State |
 | --- | --- |
 | US-23.1 — JUCE project, cross-platform build, off-audio-thread work queue | done |
 | US-23.2 — Connection panel: server URL, API key, status, model list | done |
-| US-23.3 — Generation panel | not started |
-| US-23.4 — Results panel and clip insertion | not started |
-| US-23.5 — Local cache and file management | not started |
-
-The Generation and Results panels are still outlined placeholders in the UI.
+| US-23.3 — Generation panel | done |
+| US-23.4 — Results panel and clip insertion | done |
+| US-23.5 — Local cache and file management | done |
+| US-24.1 — DAW tempo sync and tempo-matched insertion | done |
 
 | Format | Windows | macOS | Linux |
 | --- | --- | --- | --- |
@@ -88,8 +87,11 @@ the VST3 in Reaper once and confirm the UI renders and audio passes through:
    `%COMMONPROGRAMFILES%\VST3` on Windows).
 2. Reaper → Options → Preferences → Plug-ins → VST → *Re-scan*.
 3. Add it to a track with audio on it, open the UI, confirm the three panels
-   render (Connection is live; Generation and Results are placeholders) and the
-   audio is unchanged.
+   render and the audio is unchanged.
+4. Set the project tempo to 120 and confirm the **BPM** field reads 120 and the
+   indicator says `Sync: host 120 BPM`; change the project tempo and watch it
+   follow. This is the one part of US-24.1 that no unit test can cover, because
+   it needs a real host publishing a real tempo.
 
 ## Connecting to ACE-Step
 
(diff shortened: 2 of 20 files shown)


frankbria:
## Cross-family review triage

Two independent reviewers were run against this branch: **codex** (`codex review --base main`)
and **opencode/GLM**. Both converged on the same single finding, which was real.

### [P2] WSOLA left a silent tail when slowing a clip down — **fixed** in aacb553

> When `rate < 1.0` the output buffer is longer than the source, but the loop stops as
> soon as the analysis reaches the final usable source frame, so the remainder of the
> destination stays zero-filled.

**Verified before fixing**, by writing a test that measures the trailing silence directly
rather than asserting on loop shape. The effect was real but much smaller than described
at ordinary tempo pairings, and only became audible at extremes:

| Stretch | Before | After |
| --- | --- | --- |
| 120 → 100 BPM | 0.0 ms | 0.0 ms |
| 120 → 70 BPM | 7.5 ms | 1.5 ms |
| 128 → 118 BPM | 0.0 ms | 3.4 ms |
| 120 → 60 BPM | 35.8 ms | 6.2 ms |
| 200 → 60 BPM | **98.2 ms** | 0.0 ms |

**Root cause**: `findBestStart` clamped candidates to `numSamples - frameSize`, requiring
room for a whole 2048-sample frame. A candidate only needs room for the 256-sample
*correlation* — the per-frame copy length already shortens on its own near the end. With
the clamp relaxed and the early `break` dropped, the tail becomes a taper instead of a gap.

**A guard I added and then removed, on purpose.** My first fix capped the supported rate at
half-to-double speed and declined anything outside it. Once the loop was actually fixed, I
measured 200→60 and 60→200 and both came back at 0.0 ms with pitch intact — so the limit
had no evidence behind it and is not shipped. Better no guard than a guard I can't defend.

### Notes on the reviews themselves

- opencode spotted a transient working-tree state (the `minRate`/`maxRate` guard) that was
  mid-edit while it read the files, and reported against it. Not a defect in the branch.
- Neither reviewer flagged the real-time-safety, `SafePointer` lifetime, or `processFile`
  temp-file paths they were specifically pointed at.

No other Critical or Major findings from either reviewer.

PR #387, opened 2026-08-02
feat(plugin): US-24.2 Selection-Aware Generation (#321)

### LINKED ISSUE #321: US-24.2: Selection-Aware Generation
## User Story

**As a** musician, **I want** the plugin to use my DAW's time selection to set generation duration and insertion point, **so that** I can generate audio that fits exactly where I need it.

## Description

When the musician has a time selection in the DAW (e.g., bars 5–13), the plugin reads it, auto-sets the duration to match the selection length, and offers a "Generate & Insert" button that places the result precisely at the selection start.

## Functional Requirements

- Detect DAW time selection (start and end position) via host transport API
- Auto-set duration field to match selection length in seconds
- "Generate & Insert" button: generates and places result at the selection start on the plugin's track
- Selection info displayed in the generation panel (e.g., "Selection: bars 5–13, 16.0s")
- Works with no selection — falls back to manual duration and playhead insertion
- Duration field shows both the auto-detected value and allows manual override

## Acceptance Criteria

- [ ] Selecting bars 5–13 in the DAW auto-sets duration to the correct number of seconds
- [ ] "Generate & Insert" places audio starting at bar 5
- [ ] With no selection, the plugin behaves as in Stage 23 (manual duration, playhead insertion)
- [ ] Selection info is displayed correctly in the generation panel

## Dependencies

**Blocked by:** US-24.1 (prior story in this stage). Part of Stage 24: VST3 Plugin — Advanced, which is blocked by Stage 23.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 24, US-24.2

frankbria:
Implements **US-24.2 — Selection-Aware Generation** (#321).

The plugin follows the host's loop/cycle range: the generation panel displays it and the
**Duration** field is sized to it, so a generation fits the gap it was meant for.

---

| Criterion | Evidence |
| --- | --- |
| Selecting bars 5–13 auto-sets duration to the correct seconds | `GenerationPanel / AC: selecting bars 5-13 auto-sets the duration to 16 seconds` — the story's own example. Field reads `"16"` at construction and `buildRequest().durationSeconds == 16` |
| ~~"Generate & Insert" places audio starting at bar 5~~ | **Amended** per #318 → each clip row is tagged with the bar to drop it at (`Clip 1 -> 120 BPM @ bar 5`). See below |
| With no selection, behaves as in Stage 23 | `GenerationPanel / AC: with no selection the panel behaves exactly as it did in Stage 23` — duration stays at the 60 default and provably does not drift across ticks |
| Selection info displayed correctly | `GenerationPanel / AC: the selection is displayed in the generation panel` — asserts bars **and** length in the label |

`HostSync` bar arithmetic is covered for **4/4, 3/4, 6/8 and 7/8**, not just four-four.

## AC 2 — applying the decision already taken on #318

VST3 gives a plugin no way to write the host's arrangement. This is the identical
constraint raised on **#318** and resolved there: *"Insert to Track" became
drag-to-timeline, and the playhead-position criterion became "the plugin reports the
position so the user can align the drop."* I applied that settled decision rather than
re-asking.

**I did not add a "Generate & Insert" button.** With duration driven by the selection, the
existing Generate button already produces a correctly sized clip; a second button that
cannot actually place anything would promise a capability that does not exist. One honest
button beats two, one of which lies.

## What is actually being read

The **loop / cycle locators** — `AudioPlayHead::getLoopPoints()`, which JUCE fills from
`Vst::ProcessContext::kCycleValid` (`juce_audio_plugin_client_VST3.cpp:3120`). There is no
VST3 API for an arbitrary time selection. Reaper links the two by default; Cubase and Logic
use cycle markers. The README says this plainly rather than implying more.

## Design notes

**Duration reuses the BPM rule from US-24.1 exactly** — the host drives the field until the
user types, and clearing it hands control back. One mechanism to understand, not two.

**Absent host data is a rendered state, not a default.** A loop range with no tempo behind
it has no knowable length, so the panel shows the bars and says `no host tempo` rather than
claiming `0.0s`. Pinned by its own test.

**One writer for the clip caption.** The tempo tag (US-24.1) and the new bar tag both live
in `ClipRow::updateNameLabel()`, so they cannot overwrite each other's half.

`HostSync::formatBar` was extracted at its third call site, not before.

---

## Known limitations

1. Placement at the selection start is not possible (above). The clip is *sized* to the
   selection and the start bar is *reported*.
2. It follows the loop/cycle range. In a DAW where the time selection is not linked to it,
   moving the time selection alone will not update the plugin.
3. Bar numbers assume the reported time signature and a bar-1 origin. A host reporting no
   time signature shows the length only.
4. **Not verified in a real DAW** — none is installed here, and no unit test can make a
   real host publish a real loop range. `plugin/README.md` carries an explicit manual-check
   step. Everything below the host boundary is covered by tests driving a
   `juce::AudioPlayHead` double.

Closes #321

### DIFF
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -12,6 +12,7 @@ Progress:
 | US-23.4 — Results panel and clip insertion | done |
 | US-23.5 — Local cache and file management | done |
 | US-24.1 — DAW tempo sync and tempo-matched insertion | done |
+| US-24.2 — Selection-aware generation | done |
 
 | Format | Windows | macOS | Linux |
 | --- | --- | --- | --- |
@@ -92,6 +93,9 @@ the VST3 in Reaper once and confirm the UI renders and audio passes through:
    indicator says `Sync: host 120 BPM`; change the project tempo and watch it
    follow. This is the one part of US-24.1 that no unit test can cover, because
    it needs a real host publishing a real tempo.
+5. Set the cycle locators over bars 5–13 and confirm the panel reads
+   `Selection: bars 5-13, 16.0s` with **Duration** at 16. Same reason: it needs a
+   real host publishing a real loop range.
 
 ## Connecting to ACE-Step
 
@@ -157,6 +161,36 @@ JUCE is [ARA](https://www.celemony.com/en/service1/about-celemony/technologies),
 a separately licensed SDK that a handful of hosts implement. There is no way to
 read the project key through VST3, AU or Standalone, so the **Key** field is
 always manual. The indicator deliberately never claims otherwise.
+
+### Selection
+
+The plugin follows the host's **loop / cycle range** and sizes generation to it. Set
+the cycle locators over bars 5–13 in a 120 BPM 4/4 project and the panel reads
+`Selection: bars 5-13, 16.0s`, with the **Duration** field set to 16.
+
+Duration follows the same rule as BPM: it tracks the host until you type your own
+value, and clearing the field hands it back.
+
+**It is the loop range, not an arbitrary time selection.** VST3 has no API for the
+latter — `AudioPlayHead::PositionInfo::getLoopPoints()` is what exists, and JUCE fills
+it from `Vst::ProcessContext::kCycleValid`. In Reaper the loop range is linked to the
+time selection by default (Options → Loop points linked to time selection); in Cubase
+and Logic these are the cycle markers. In a DAW where the two are not linked, moving
+the time selection alone will not move the plugin's.
+
+With no loop range set, the panel reads `Selection: none` and Duration keeps its
+default of 60 — the Stage-23 behaviour, unchanged.
+
+A loop range with no host tempo behind it shows the bars but no length, rather than
+claiming `0.0s`.
+
+### Where to drop it
+
+The results panel tags each clip with the bar a drop should line up with —
+`Clip 1  ->  120 BPM  @ bar 5`. The plugin cannot place the audio there itself: VST3
+gives it no way to write the host's arrangement. This is the same resolution agreed on
+[#318](https://github.com/frankbria/auto-music-studio/issues/318) for the playhead
+readout — the plugin reports the position, and the drop stays yours.
 
 ### Tempo-matched insertion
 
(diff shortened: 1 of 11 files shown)


frankbria:

**Three P2 findings, all real, all fixed in 111583e.** Each was confirmed by a test that
failed before the fix — and two turned out to be worse than reported.

### 1. `buildRequest` ignored the host when a field was empty

> Clearing a manually entered duration and clicking Generate before the next 2 Hz tick
> submits 60 seconds instead of the selected loop length.

**Confirmed**, and it applies to BPM too: `Expected 16, Actual 60` and `Expected 120,
Actual -1`.

The suggested fix — call `applyHostSelection()` from the change handler — does **not**
close it. `TextEditor` delivers `onTextChange` *asynchronously* (it posts a command
message), so no work inside that callback can help a Generate that lands first.

Fixed at the root instead: an empty field now means "follow the host" **inside
`buildRequest`**, falling back to Auto/60 only when there is no host value to follow. The
request no longer depends on a display side-effect having already run.

### 2. The readouts were 16 px wide, not 112 px

> At 720 px this split gives the selection label only about 112px.

**Confirmed and worse.** At the editor's *minimum* size — which a user can drag to — both
labels got **16 px**:

```
the sync indicator is clipped: "Sync: host 120 BPM" needs 102px, has 16px
the selection readout is clipped: "Selection: bars 5-13, 16.0s" needs 138px, has 16px
```

They now have a row to themselves rather than competing with three fixed-width controls
and two toggles.

### 3. The clip caption had 48 px for 147 px of text

> `ClipRow::resized()` still gives that label only 52px.

**Confirmed** — and this one has been broken since **US-24.1 (#386) shipped**, not just in
this PR. The tempo-match tag that story added has been invisible the whole time; this PR's
bar tag simply made it longer. My fault, and it would have kept shipping unnoticed.

The label now scales with the row width, and the separators are compact.

### Guarding against a repeat

All three are pinned by tests that measure the rendered string against the space the label
actually has, via `GlyphArrangement::getStringWidth`, at the editor's **minimum** size:

```
expect ((float) label.getWidth() >= needed, "... is clipped: ...")
```

A future caption cannot silently outgrow its box.

PR #388, opened 2026-08-02
feat(plugin): US-24.3 MIDI Input and Sidechain Audio (#322)

### LINKED ISSUE #322: US-24.3: MIDI Input and Sidechain Audio
## User Story

**As a** musician, **I want** to feed MIDI or audio from my DAW into the plugin as creative input, **so that** I can use my own performances and existing tracks as seeds for AI generation.

## Description

The plugin accepts MIDI input from the DAW for "Complete" generation mode (turning a MIDI sketch into full audio) and sidechain audio input for Cover/Repaint modes (restyling an existing DAW track). These inputs deeply integrate AI generation into the production workflow.

## Functional Requirements

- MIDI input mode: plugin receives MIDI from the DAW (played or recorded)
- Captured MIDI is used as melodic input for the "Complete" generation mode
- "Record MIDI" toggle captures incoming MIDI to a buffer for submission
- Sidechain audio input: receive audio from a sidechain bus/send
- "Capture Sidechain" records sidechain audio for use as reference in Cover or Repaint modes
- Mode selector includes: Text to Music, Cover (sidechain), Complete (MIDI), Repaint (sidechain + time range)
- Clear indicators showing when MIDI or sidechain audio has been captured and is ready for generation

## Acceptance Criteria

- [ ] Playing MIDI into the plugin captures it and enables "Complete" mode generation
- [ ] Sidechain audio from another track is captured and usable as Cover reference
- [ ] Generated audio from MIDI input reflects the melodic content of the MIDI performance
- [ ] Cover mode with sidechain input produces a restyled version of the reference audio
- [ ] Mode selector correctly shows available modes based on captured inputs

## Dependencies

**Blocked by:** US-24.2 (prior story in this stage). Part of Stage 24: VST3 Plugin — Advanced, which is blocked by Stage 23.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 24, US-24.3

frankbria:
Implements **US-24.3 — MIDI Input and Sidechain Audio** (#322).

The plugin now accepts MIDI and a sidechain input and can seed a generation from either.

---

## The finding that shaped this story: ACE-Step has no MIDI input

I checked the server rather than assuming. Its task types are `text2music`, `cover`,
`repaint`, `extract`, `lego`, `complete`, `mashup` — and **`complete`, the mode this story
wants MIDI to drive, takes `src_audio_path`.** There is not one reference to MIDI anywhere
in the ACE-Step source, and the platform's own `midi_client.py` runs the other way entirely
(audio → MIDI, via basic-pitch).

So *Record MIDI* captures your performance and the plugin **renders it to a plain tone**,
which is submitted as the melodic reference. That does what the user story asks for
musically — a MIDI sketch becomes full audio — but the model conditions on rendered audio,
not on note data, and the README says so plainly.

---

| Criterion | Outcome |
| --- | --- |
| Playing MIDI captures it and enables Complete mode | **Met** — `Capture / AC: playing MIDI into the plugin captures it` and `GenerationPanel / AC: modes are only offered once their input has been captured` |
| Sidechain audio captured and usable as Cover reference | **Met** — round-trips through `SidechainCapture`, and Cover/Repaint unlock only once a take exists |
| Generated audio reflects the melodic content of the MIDI | **Met — verified against a live ACE-Step server.** See below |
| Cover with sidechain produces a restyled version | **NOT VERIFIED.** See below |
| Mode selector shows available modes based on captured inputs | **Met** — a mode with no input is greyed out, and Cover/Repaint stay unavailable when no sidechain is routed at all |

### Live verification — Complete mode ✅

A melody was rendered through the **real `MidiCapture`** (not a fixture) and measured by
librosa as `C4 C4 G4 G4 A4 A4 G4`. Submitting it as `task_type=complete`:

```
sketch     3.52s  top pitch classes: G, A, C, G#
generated  5.12s  top pitch classes: G, A, C, G#

chroma cosine similarity:                            0.847
mean similarity under rotation (unrelated baseline):  0.579
```

Identical top four pitch classes, and similarity well clear of the unrelated baseline.

### Live verification — Cover mode ❌ not verified

The integration is correct: the server **accepts** the task and its log confirms it
consumes the reference —

```
[generate_music] Processing source audio...
[generate_music] cover task: using params.caption='heavy distorted electric guitar, rock'
```

— but the output's harmonic content was **indistinguishable from unrelated**, at both
turbo and standard quality:

| Inference steps | similarity | unrelated baseline |
| --- | --- | --- |
| 8 | 0.623 | 0.620 |
| 60 | 0.647 | 0.637 |

I am reporting this as unverified rather than claiming the criterion. Plausible causes I
did not chase: ACE-Step's cover may condition on style/timbre rather than harmony; the
reference I used is a bare synthesised melody rather than a real track; or parameters like
`style_influence` control adherence. **What is proven is that the plugin submits the mode
correctly and the server uses the file.** What is not proven is the musical claim.

---

## Design notes

**Nothing is synthesised on the audio thread.** `processBlock` only pushes note events into
a lock-free FIFO; draining, rendering and writing happen on the message thread when the
take is disarmed. Far less machinery than a live `juce::Synthesiser`, and the audio
callback still allocates nothing. Events past the FIFO's capacity are **dropped and
counted**, and the indicator reports the count rather than passing the take off as clean.

**The sidechain bus is disabled by default**, so a host with no sidechain send loads the
plugin exactly as before and existing sessions do not change shape.
`isBusesLayoutSupported` accepts it absent, mono or stereo — that is what pluginval fuzzes,
so all three have tests.

**The regression that mattered most** has its own test: audio still passes through
bit-identical with MIDI *and* a sidechain being captured.

`AudioIo::writeWav` was extracted at its third caller, not before.

---

## Known limitations

1. **MIDI reaches the model as rendered audio, not note data** — ACE-Step has no MIDI
   conditioning (above).
2. **Cover's musical behaviour is unverified** (above).
3. Sidechain capture is bounded to `SidechainCapture::maxSeconds`; it stops at the bound
   rather than wrapping, and says so.
4. `src_audio_path` is a **server-side** path. Fine for the local ACE-Step the plugin
   targets; a remote server needs an upload endpoint (US-24.5 territory).
5. **Not verified in a real DAW** — no DAW is installed here. The README carries a manual
   check for MIDI routing and sidechain sends.

Closes #322

### DIFF
--- a/plugin/CMakeLists.txt
+++ b/plugin/CMakeLists.txt
@@ -37,6 +37,7 @@ set(ACEMUSIC_SOURCES
     source/AceStepClient.cpp
     source/BackgroundTaskQueue.cpp
     source/ClipCache.cpp
+    source/AudioIo.cpp
     source/ClipPlayer.cpp
     source/ConnectionManager.cpp
     source/ConnectionPanel.cpp
@@ -45,9 +46,11 @@ set(ACEMUSIC_SOURCES
     source/GenerationPanel.cpp
     source/GenerationRequest.cpp
     source/HostSync.cpp
+    source/MidiCapture.cpp
     source/PluginProcessor.cpp
     source/PluginEditor.cpp
     source/ResultsPanel.cpp
+    source/SidechainCapture.cpp
     source/TimeStretch.cpp)
 
 # ---------------------------------------------------------------------------
@@ -64,7 +67,7 @@ juce_add_plugin(AceMusicPlugin
     PLUGIN_MANUFACTURER_CODE    Amst
     PLUGIN_CODE                 Aces
     IS_SYNTH                    FALSE
-    NEEDS_MIDI_INPUT            FALSE
+    NEEDS_MIDI_INPUT            TRUE
     NEEDS_MIDI_OUTPUT           FALSE
     IS_MIDI_EFFECT              FALSE
     EDITOR_WANTS_KEYBOARD_FOCUS TRUE
@@ -104,6 +107,7 @@ target_sources(AceMusicPluginTests PRIVATE
     tests/TestMain.cpp
     tests/AceStepClientTests.cpp
     tests/BackgroundTaskQueueTests.cpp
+    tests/CaptureTests.cpp
     tests/ClipCacheTests.cpp
     tests/ClipPlayerTests.cpp
     tests/ConnectionManagerTests.cpp
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -13,6 +13,7 @@ Progress:
 | US-23.5 — Local cache and file management | done |
 | US-24.1 — DAW tempo sync and tempo-matched insertion | done |
 | US-24.2 — Selection-aware generation | done |
+| US-24.3 — MIDI input and sidechain audio | done |
 
 | Format | Windows | macOS | Linux |
 | --- | --- | --- | --- |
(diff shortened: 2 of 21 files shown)


frankbria:

**Two P1 findings. Both real, both blocking — between them the feature this PR adds was
unusable from the UI.** Fixed in 4c98884, and both verified by mutation.

### 1. Complete and Repaint submitted as text-to-music

`buildRequest()` still mapped the selector with `getSelectedId() == 2 ? cover : textToMusic`.
Correct while the list had two entries; this PR added Complete and Repaint as ids 3 and 4,
so choosing either **silently built a text-to-music request**, skipped `attachSourceAudio()`
and discarded the capture entirely.

Now indexed off `GenerationRequest::allModes()` — the same list the selector is built from,
so the two cannot drift apart again.

### 2. Generate could never enable for any source mode

`refresh()` validates `buildRequest()`; `findProblem()` rejects a source mode with an empty
`sourceAudioPath`; and the path was only set inside the click handler that the disabled
button could never run. A perfect deadlock.

The capture's path is now known *before* it is written: `buildRequest` names it as soon as a
capture exists, and `attachSourceAudio` writes to that same path at click time.
`getCaptureFileFor` / `hasCaptureFor` are the single source of both answers, and
`isModeAvailable` defers to them instead of repeating the rule.

### Why my tests missed both

They checked the availability *gate* (`isModeAvailable`) rather than what the panel actually
*builds*. A mode can be offered and still produce the wrong request — which is exactly what
happened.

The new tests drive the whole path: select each mode, assert the request carries it; capture,
assert the request becomes submittable and names the right file. Reverting either fix makes
them fail:

```
!!! selecting "Complete" built a request for "Text to Music"
!!! selecting "Repaint" built a request for "Text to Music"
!!! the request names no source, so Generate can never enable
!!! Cover stayed unsubmittable after a capture: Cover mode needs a source audio file
```

PR #389, opened 2026-08-02
feat(plugin): US-24.4 Lego Mode and Layer-by-Layer Generation (#323)

### LINKED ISSUE #323: US-24.4: Lego Mode and Layer-by-Layer Generation
## User Story

**As a** musician, **I want** to build up a song layer by layer, generating one instrument or part at a time, **so that** I have fine-grained control over the arrangement.

## Description

Lego mode enables iterative layer-by-layer generation. The musician generates a drum track, then a bass line that fits the drums, then a melody on top, each on a separate DAW track. Each layer is generated with awareness of what exists already.

## Functional Requirements

- "Lego Mode" toggle in the generation panel
- In Lego mode, existing audio on the plugin track (or sidechain) is sent as context for the next generation
- Each generated layer is placed on a separate new track
- Layer order is tracked: the musician can regenerate any layer while keeping others
- Prompt for each layer specifies the instrument or part (e.g., "add a funky bass line")
- Context audio is mixed down and sent as reference to the API

## Acceptance Criteria

- [ ] Generating a drum layer, then a bass layer, produces two tracks where the bass fits the drums
- [ ] Each layer is placed on its own track in the DAW
- [ ] Regenerating one layer does not affect other layers
- [ ] Context audio is audibly reflected in the generated layer (musical coherence)

## Dependencies

**Blocked by:** US-24.3 (prior story in this stage). Part of Stage 24: VST3 Plugin — Advanced, which is blocked by Stage 23.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 24, US-24.4

frankbria:
Implements **US-24.4 — Lego Mode and Layer-by-Layer Generation** (#323).

Pick **Lego** and a **Layer** track, describe the part, Generate. Each finished layer joins
the stack, and the next generation receives the mix of the layers so far as its context.

---

| Criterion | Outcome |
| --- | --- |
| Drums then bass produces two layers where the bass fits the drums | **Met — verified live.** See below |
| ~~Each layer is placed on its own track in the DAW~~ | **Amended** per #318 — each layer is a draggable clip |
| Regenerating one layer does not affect the others | **Met** — `LegoStack / AC: regenerating one layer leaves the others untouched` |
| Context audio is audibly reflected in the generated layer | **Met — verified live.** See below |

### Live verification: drums, then bass in context

```
layer1 drums  16.0s  tempo 117.45 BPM  centroid 2219 Hz
layer2 bass   16.0s  tempo 117.45 BPM  centroid 1888 Hz

distinct renders                        yes
tempo agreement                         0.00 BPM
onset-envelope correlation (aligned)   +0.844
same under random time shifts (null)   -0.002
```

The second layer's rhythmic events land **on** the first's, and shifting either in time
destroys the correlation completely. Tempo matches exactly. The bass also sits lower in the
spectrum than the drums, as a bass part should.

That is a real result, and worth contrasting with US-24.3's Cover mode, which I reported as
**unverified** because its numbers came out at chance. The same method distinguishes them.

---

## What the server actually does, and a correction

`lego` is a genuine ACE-Step task type — *"generate a specific instrument track in
context"* — steered by an **`instruction`** string naming the track, not by your caption:

```
Generate the BASS track based on the audio context:
```

**I got this wrong first.** I read `acestep/ui/gradio/api/api_routes.py`, found no
`instruction` handling anywhere in it, and concluded track steering was CLI-only. A live
submission proved otherwise — the running server uses a different route and used the
instruction exactly as sent. Recording it because the code reading was confidently wrong
and only the empirical check caught it.

The track list is ACE-Step's own `TRACK_NAMES`, so a layer always names something the model
was trained on. Your prompt still describes *what kind* of part you want; the two compose.

---

## Design notes

**The first layer is deliberately not a `lego` task.** There is nothing to build on, so no
`task_type` and no context are sent and it goes out as ordinary text-to-music. This is the
one thing about the mode that is easy to get wrong, so it has its own test.

**Regenerating a layer excludes that layer from its own context** — a replacement bass
should fit the drums it sits under, not the bass it is replacing. That is the whole reason
`LegoStack` is a type rather than a `juce::Array` in the panel. Layers can also be muted out
of the context without being deleted.

**A layer whose file has gone missing degrades the mix rather than failing it**, so one bad
clip does not take down a whole build.

---

## Known limitations

1. **The plugin cannot place layers on separate tracks** — VST3 has no arrangement write.
   Same limit as #318 and #321; each layer is a draggable clip.
2. **The context is the plugin's own layers, not the DAW timeline.** A plugin cannot read
   other tracks. To bring real DAW audio in, capture it on the sidechain (US-24.3) — the
   two compose.
3. Not verified in a real DAW (none installed here); the README carries a manual check.

Closes #323

### DIFF
--- a/plugin/CMakeLists.txt
+++ b/plugin/CMakeLists.txt
@@ -46,6 +46,7 @@ set(ACEMUSIC_SOURCES
     source/GenerationPanel.cpp
     source/GenerationRequest.cpp
     source/HostSync.cpp
+    source/LegoStack.cpp
     source/MidiCapture.cpp
     source/PluginProcessor.cpp
     source/PluginEditor.cpp
@@ -117,6 +118,7 @@ target_sources(AceMusicPluginTests PRIVATE
     tests/GenerationManagerTests.cpp
     tests/GenerationPanelTests.cpp
     tests/GenerationRequestTests.cpp
+    tests/LegoStackTests.cpp
     tests/HostSyncTests.cpp
     tests/PluginProcessorTests.cpp
     tests/ResultsPanelTests.cpp
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -14,6 +14,7 @@ Progress:
 | US-24.1 — DAW tempo sync and tempo-matched insertion | done |
 | US-24.2 — Selection-aware generation | done |
 | US-24.3 — MIDI input and sidechain audio | done |
+| US-24.4 — Lego mode, layer by layer | done |
 
 | Format | Windows | macOS | Linux |
 | --- | --- | --- | --- |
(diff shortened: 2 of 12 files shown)


frankbria:

**Two P1 findings. Both real, and between them Lego mode did nothing at all.** Fixed in
82c940b, both verified by mutation.

### 1. Lego could not be selected

`isModeAvailable` ran Lego through `needsSourceAudio`/`hasCaptureFor` like every other
non-text mode, so with no MIDI or sidechain capture the dropdown entry was **greyed out —
from an empty stack, which is where every build starts.** Lego needs a *track*, not a
capture, and its first layer has nothing to build on by design.

### 2. Nothing in the plugin ever called `addLegoLayer`

Only tests did. A finished generation was never recorded, so the stack stayed at **zero
layers** and every generation went out as another first-layer text-to-music request with no
context — precisely the thing the feature exists to avoid.

The panel already listens to `GenerationManager`, so a completed run is now collected
there. Two details that matter:

- The track and prompt are captured when the run **starts**, not when it lands — by then
  the user may have moved the selector on to the next part they intend to add.
- The pending flag is cleared **before** adding, because `addLegoLayer` calls `refresh()`,
  which lands back in the collector. Otherwise the layer would be re-added on every refresh
  for as long as the run read complete.

### This is the second story running where I tested pieces and missed wiring

US-24.3 had the same shape: the parts were covered, the production path was not, and the
feature was unusable while every test passed. I have changed what these tests assert.

The new tests drive the real path — click Generate, let the run publish a clip, assert the
stack grew, assert it grew **only once**, and assert the next request actually carries
context. Reverting either fix fails them:

```
!!! Lego was unavailable with an empty stack, so it can never be started
!!! the Lego entry is greyed out in the dropdown
!!! the finished layer was never added to the stack
!!! the layer recorded the wrong track -- Expected value: drums, Actual value:
```

PR #391, opened 2026-08-02
feat(api): US-25.1 Voice Model Training Backend (#325)

### LINKED ISSUE #325: US-25.1: Voice Model Training Backend
## User Story

**As a** musician, **I want** to upload reference audio files and train a custom voice model, **so that** generated songs can use my unique vocal style.

## Description

The backend accepts 2–10 reference audio files, validates their quality and consistency, and queues a LoRA fine-tuning job. The resulting voice model weights are stored in the user's private voice library. Training consumes premium credits (10 credits).

## Functional Requirements

- POST `/api/v1/voice-models/train` accepts: 2–10 audio files (WAV, FLAC, MP3 at 16kHz+ sample rate), model name, optional description
- Validation: minimum 2 files, maximum 10, audio quality check (sample rate, duration, noise level), consistency check (similar vocal characteristics across files)
- Training job queued with estimated time (typically minutes on GPU)
- Credit validation: 10 credits deducted before training begins; insufficient credits returns 402
- On completion, LoRA weights file stored and associated with the user's account
- On failure, credits are refunded and error details are returned

## Acceptance Criteria

- [ ] Uploading 3 valid audio files and triggering training returns a job ID
- [ ] Validation rejects files below 16kHz sample rate with a clear error message
- [ ] Training job completes and produces a usable voice model
- [ ] 10 credits are deducted on submission; refunded on failure
- [ ] Uploading 1 file or 11 files returns a validation error

## Dependencies

**Blocked by:** Stage 21. First story of Stage 25.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 25, US-25.1

frankbria:
Implements **US-25.1 — Voice Model Training Backend** (#325).

Upload 2–10 reference recordings, train a LoRA voice model on ACE-Step, and store the
adapter against the musician's account.

---

## Acceptance criteria

| Criterion | Outcome |
| --- | --- |
| Uploading 3 valid files and triggering training returns a job ID | **Met** |
| Validation rejects files below 16 kHz with a clear error message | **Met** — the message names the file: `thin.wav: sample rate is 8000 Hz, below the 16000 Hz minimum.` |
| Training job completes and produces a usable voice model | **Met — verified against a live ACE-Step**, see below |
| 10 credits deducted on submission; refunded on failure | **Met** — plus a ledger row for both, so the balance reconciles |
| Uploading 1 file or 11 files returns a validation error | **Met** |

### AC 3, verified for real

A 3-reference run against a live ACE-Step 1.5:

```
preprocess   3/3 samples -> tensors on disk
training     11,010,048 parameters, loss 1.647
export       43 MB adapter
load back    {"lora_loaded": true, "active_adapter": "adapter"}
```

The adapter **loads back into the model**, which is the difference between "the job
completed" and "the voice is usable" — a distinction the code makes too: `is_usable`
requires `ready` **and** weights.

---

## The ACE-Step integration was written twice

The first version invented its payloads and its status vocabulary, and **the stub tests
passed because the stub encoded the same guess.** Probing the live server found four
things wrong, none of them guessable:

1. The sequence is **scan → label → preprocess**, not preprocess directly. Unlabelled
   samples are silently skipped and preprocess returns `No labeled samples to
   preprocess` with a task id that immediately 404s.
2. `/v1/training/start` takes `tensor_dir` + `lora_output_dir`, not a dataset name.
3. Training status is a **boolean** `is_training` with a human `status` string
   (`"Idle"`), not `completed`/`failed` — **and it reads false both before a run starts
   and after it ends**, so a poller without a start grace period reports success for a
   run that never began, and keeps the credits.
4. Export writes a **directory**; `/v1/lora/load` wants its `adapter/` child and rejects
   the root.

Plus a deployment constraint: ACE-Step's path safety rejects anything outside its working
directory, and its training endpoints take **paths, not uploads** — so the platform and
ACE-Step must share a filesystem.

The stub now asserts on the real payload fields and on label-before-preprocess ordering,
so a wrong request fails the test instead of agreeing with it.

## Three findings from the pre-PR review, two of them P1

**The references were never copied to where ACE-Step scans.** Uploads live in the
platform's storage backend; ACE-Step scans a directory. Every real job would have found
zero samples, failed, and refunded.

**My live verification missed this because I had copied the files there by hand.** Proving
ACE-Step's sequence works said nothing about the handoff between the two systems — which
was the only part that was mine. Worth recording, because it is the same shape as the
mistake the probe was meant to fix.

**Refunding on `CancelledError` broke requeue.** `JobProcessor` re-raises it *without*
marking the job, deliberately leaving it `processing` so startup requeues it as stale. The
refund therefore either gave credits back for a run that then succeeded, or refunded the
same charge twice when the retry failed, since refunds are not idempotent. A cancelled run
now returns to `queued` untouched.

**No `CreditTransaction` row was written.** `/users/me/credits` builds its history from the
ledger. Now recorded on charge *and* refund, with a test asserting the ledger reconciles to
a zero net.

---

## Design notes

**Validation order is the behaviour.** Count → format/sample-rate → duration/level →
consistency → *then* credits. A rejected upload costs nothing, and there is a test that
asserts the balance is untouched after a rejection.

**Every rejection names the file** — "invalid file" is useless when ten were uploaded.

**Refunds use `credits_charged`, not the current price**, so a change to the cost table
cannot alter what someone already paid gets back. Tested by changing the price between
charge and failure.

---

## Known limitations

1. **The consistency check is spectral, not speaker verification.** It compares spectral
   centroids, which catches the mistake that actually happens — a drum loop among the
   vocal takes — and will **not** distinguish two people with similar timbre. The
   threshold is deliberately loose: a false reject costs a usable model, which is worse
   than training on one odd take. Doing it properly needs speaker embeddings.
2. **Platform and ACE-Step must share a filesystem** (above). A split deployment needs a
   shared volume; there is no upload endpoint to fall back on.
3. Progress reporting and notifications are US-25.2; this story records the phase and
   ACE-Step's own step/epoch/loss on the job but pushes nothing.
4. **Loading a LoRA into ACE-Step is not cleanly reversible** — `/v1/lora/unload` failed
   with *"Base decoder backup not found"* after my verification run, and a restart was
   needed to clear it. Not something this code does, but worth knowing.

Closes #325

### DIFF
--- a/docs/voice-models.md
+++ b/docs/voice-models.md
@@ -1,0 +1,79 @@
+# Custom voice models (Stage 25)
+
+Train a voice from your own recordings and use it in generation.
+
+## Training a voice
+
+`POST /api/v1/voice-models/train` — multipart: 2–10 reference recordings, a name, and an
+optional description.
+
+| Check | Rule |
+| --- | --- |
+| File count | 2–10 |
+| Format | anything `soundfile` decodes (WAV, FLAC, OGG) |
+| Sample rate | ≥ 16 kHz |
+| Duration | ≥ 1 s per reference |
+| Level | not silent |
+| Consistency | no obvious odd-one-out (see the caveat below) |
+
+**Every rejection names the file**, because "invalid file" is useless when ten were
+uploaded: `thin.wav: sample rate is 8000 Hz, below the 16000 Hz minimum.`
+
+**Validation runs before the charge.** A rejected upload costs nothing. Training costs
+**10 credits**, deducted only once every check that can reject the request has passed, and
+**refunded on any failure** — using the amount actually charged, so a later price change
+cannot alter what you get back.
+
+### The consistency check is not speaker verification
+
+It compares spectral centroids. That catches the mistake that actually happens — a drum
+loop or a backing track mixed in among the vocal takes — and it will **not** distinguish
+two different people with similar timbre. The threshold is deliberately loose: a false
+reject costs you a usable model, which is worse than training on one slightly odd take.
+
+Doing this properly needs speaker embeddings, which is its own piece of work.
+
+## How training actually runs
+
+The worker drives ACE-Step's LoRA endpoints:
+
+```
+POST /v1/dataset/scan               → find the references
+PUT  /v1/dataset/sample/{i}         → label each one
+POST /v1/dataset/preprocess_async   → build tensors
+POST /v1/training/start             → fine-tune
+GET  /v1/training/status            → poll
+POST /v1/training/export            → export the adapter
+```
+
+Four things about that sequence are worth knowing, because none is guessable and each was
+wrong in the first implementation:
+
+1. **Scan and label are not optional.** Preprocessing works on the server's *current*
+   dataset and silently skips unlabelled samples — calling it directly returns
+   `No labeled samples to preprocess` and produces nothing.
+2. **Training status is a boolean.** `/v1/training/status` reports `is_training` with a
+   human `status` string ("Idle"), not `completed`/`failed`. `is_training` reads false
+   both *before* a run starts and after it ends, so the poller has a start grace period —
+   without it, a run that never began reports success and keeps the credits.
+3. **Export writes a directory.** The loadable PEFT adapter is the `adapter/` child of the
+   export path; pointing `/v1/lora/load` at the root is rejected.
+4. **Paths must resolve under ACE-Step's working directory** — see
+   `acestep/training/path_safety.py`. The training endpoints take *paths, not uploads*, so
+   **the platform and ACE-Step must share a filesystem.** A split deployment needs a
+   shared volume; there is no upload endpoint to fall back on.
+
+## Verified end to end
+
+A real 3-reference run against a local ACE-Step 1.5:
+
+```
+preprocess   3/3 samples → tensors on disk
+training     11,010,048 parameters, loss 1.647
+export       43 MB adapter
+load back    {"lora_loaded": true, "active_adapter": "adapter"}
+```
+
+The exported adapter loads back into the model, which is the difference between "the job
+completed" and "the voice is usable" — a distinction the code makes too: a model is only
+`is_usable` when it is `ready` **and** holds weights.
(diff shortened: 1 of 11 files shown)


frankbria:
## Cross-family review triage (codex), run before this PR was opened

**Three findings, all real, two P1.** Fixed in 4679fcf — and then genuinely fixed in
c5e277c, see the postscript.

### 1. [P1] The references were never copied to where ACE-Step scans

Uploads live in the platform's storage backend; ACE-Step scans a directory on its own
filesystem. The worker asked it to scan a path **nothing ever wrote to**, so every real job
would find zero samples, fail, and refund.

**My live verification missed this because I had copied the files there by hand.** Proving
ACE-Step's sequence works said nothing about the handoff between the two systems — which
was the only part that was mine. The worker now downloads each reference into the shared
training directory, and the root is threaded through the processor as configuration.

### 2. [P1] Refunding on `CancelledError` broke requeue

`JobProcessor` re-raises `CancelledError` *without* marking the job, deliberately leaving
it `processing` so startup requeues it as stale. The refund therefore either gave credits
back for a run that then succeeded, or refunded the same charge **twice** when the retry
also failed, since refunds are not idempotent. A cancelled run now returns to `queued`
untouched.

### 3. [P2] No `CreditTransaction` row was written

`/users/me/credits` builds its history from the ledger, so the balance dropped with nothing
to explain it — unlike every other billed endpoint. Recorded on charge *and* refund, with a
test asserting the ledger reconciles to a **zero net** after a refund.

---

## Postscript: the first fix commit shipped neither P1 fix

Worth recording, because the commit message claimed both.

The edits were multi-line `str.replace` calls that **did not match**, because `black` had
already reformatted those call sites onto single lines — and a non-matching replace returns
the original silently. So `training_root` was never passed to `_train()` (a `TypeError` on
every real invocation) and `except BaseException` was never narrowed.

**Only the integration suite caught it**; the CI-safe tests do not reach that path. Fixed
properly in c5e277c, verified by `grep` rather than by intent, with the edit asserting its
own match.

Final: **20/20 integration tests pass**, 14 CI-safe tests pass.

PR #395, opened 2026-08-04
feat: US-25.4 Voice Selection in Creation Modes

### LINKED ISSUE #328: US-25.4: Voice Selection in Creation Modes
## User Story

**As a** musician, **I want** to select a custom voice model when creating songs in any mode, **so that** all my music can feature my unique vocal style.

## Description

Custom voice models appear as an option in all creation forms — Simple, Advanced, Cover, Add Vocal, and Extend. An "Add Voice" modal lets the musician browse their voice library and attach a voice to the generation.

## Functional Requirements

- "Add Voice" button in the creation form opens a voice selection modal
- Modal lists all trained voice models with name, preview button, and select button
- Selected voice is shown as a pill/badge in the creation form
- Voice selection is available in: Simple mode, Advanced mode, Cover workflow, Add Vocal workflow, Extend workflow
- Voice parameter sent to the generation API alongside other parameters
- "Remove Voice" option to clear the selection and use the default model voice
- Voice selector also available in the VST3 plugin (if authenticated with the platform)

## Acceptance Criteria

- [ ] Voice selection modal appears and lists all trained voices
- [ ] Selecting a voice and generating produces audio with the custom vocal timbre
- [ ] Voice selection works in all five creation modes
- [ ] Removing voice selection reverts to the default model voice
- [ ] Generation without a custom voice works as before (no regressions)

## Dependencies

**Blocked by:** US-25.3 (prior story in this stage). Part of Stage 25: Custom Voice Models, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 25, US-25.4

frankbria:
## US-25.4: Voice Selection in Creation Modes

Closes #328.

A musician can now attach one of their trained voices to a generation, in every creation
mode, and the generation actually sings in it.

### The constraint that shaped this

ACE-Step's LoRA is **global handler state**, not a per-task parameter — `/release_task`
has no `lora_path` field, so whichever adapter is loaded conditions every task the host
runs. Loading a voice for one job would therefore have put it into anyone else's job
running alongside it.

`src/acemusic/api/tasks/voice_adapter.py` owns that state. Generations that want the
adapter already loaded run **concurrently, exactly as before**; only a *change* is
exclusive — it waits for the in-flight generations to drain, swaps the adapter, and lets
everyone back in. New arrivals queue behind a waiting switch even when the current adapter
suits them, so a steady stream of one voice cannot starve another. A deployment that never
trains a voice never talks to `/v1/lora` at all.

### Backend

- `voice_model_id` on `POST /generate` and on the `cover` / `add-vocal` / `extend` clip
  endpoints, validated against the caller's library **before any credit is deducted**:
  404 for unknown-or-someone-else's (deliberately indistinguishable), 409 for a voice that
  exists but has no usable weights yet.
- A voice pins the job to **local** compute — the adapter is a file on the local ACE-Step
  host, and a remote backend has no way to load it.
- The worker re-resolves the voice at run time rather than trusting enqueue time: a voice
  deleted in between fails the job instead of quietly delivering the base model's voice.
- `GET /voice-models/{id}/preview` streams the first stored reference recording, owner-scoped.

### Web

- The Add Voice modal reads the real library instead of `MOCK_VOICES`, offers only voices
  that are `ready`, and fetches a reference recording on demand for preview.
- The selection reaches the API in all five modes: Simple and Advanced through the existing
  `AudioInputs` chip, and Cover / Add Vocal / Extend through a new `VoiceField` in their
  modals. Removing the chip reverts to the model's own voice.

### Tests

- `tests/test_voice_adapter.py` (11) — load/unload/no-op, retry after a failed call, that
  two different voices are never live at once, and that one voice does not cost the
  throughput plain generation had.
- `tests/test_voice_selection_api.py` (20) — the five modes, the 404/409 rules with credit
  balances asserted, preview scoping, and worker resolution.
- Web: the modal against a real library, and the payload wiring (plus removal) in each mode.

Three ACE-Step test doubles gained the `base_url` the real client already carried; without
it they failed their jobs on an `AttributeError` rather than exercising the path under test.

### Demo

`docs/demos/us-25.4-voice-selection.md` — the modal in a browser against a real seeded
library, the LoRA traffic the ACE-Step host actually receives, and jobs read back out of
MongoDB for each of the five modes.

### Known limitations

- **The VST3 plugin voice selector is not included.** The plugin's `PlatformClient` only
  browses and uploads clips — it generates straight against ACE-Step rather than through
  the platform — so a voice selector there is its own piece of work. Filed as #396.
- **The adapter state is tracked per worker process, for one ACE-Step host.** Several
  workers against one host would each believe their own answer, and another client on the
  same host (the CLI, the plugin) is not covered by this coordination. Named adapters plus
  shared state is the upgrade path; the module says so.
- **No live GPU verification.** No ACE-Step host is available here, so "produces audio with
  the custom vocal timbre" is verified structurally — the adapter is loaded before the task
  is submitted and held until it completes — not by listening to the output.


frankbria:
## Cross-family review

`opencode` (GLM) failed with a server-side `UnknownError` on both attempts, so this is
`codex review --base main` — a different model family from the author, which is the point
of the gate.

### Finding: [P2] a queued voice switch could be starved — CONFIRMED, fixed in e53e733

> When a generation for adapter A is running and a request for adapter B is queued, any
> later request for A reaches this loop with `_matches(weights_path)` true and immediately
> increments `_users`, even though `_switchers > 0`.

Correct, and my "new arrivals queue behind them" comment was wrong about its own code. The
fast path did guard on `_switchers == 0`, but the *wait loop's* early exit did not: an
arrival whose adapter already matched fell straight through it. Under a steady stream of
plain generations, `_users` never reaches 0 and the queued voice waits forever.

Fixed with an adapter epoch. A waiter may only skip its own switch when the epoch changed
**after it started waiting** — i.e. someone else really did swap to what it wanted — rather
than because the adapter happened to match on arrival.

`tests/test_voice_adapter.py::test_a_queued_voice_is_not_starved_by_steady_base_traffic`
reproduces it: reverting the source fix alone fails it with `- voice / + plain-0`.

No other findings were reported.

frankbria:
## Demo (Phase 11)

`docs/demos/us-25.4-voice-selection.md` — every acceptance criterion mapped to outcome evidence.

| Acceptance criterion | Verdict |
|---|---|
| Modal appears and lists all trained voices | ✅ browser screenshot against a real seeded library; the still-training voice is absent, another user's voice is not in `GET /voice-models` |
| Selecting a voice produces the custom vocal timbre | ⚠️ **structural only** — `POST /v1/lora/load` is issued before the task is submitted and held until it completes, and no other voice can be live in that window. There is no ACE-Step GPU on this machine, so the rendered audio was not listened to. |
| Works in all five creation modes | ✅ Simple, Advanced, Cover, Add Vocal, Extend each persist `voice_model_id` on the job (read back out of MongoDB) |
| Removing the selection reverts to the default voice | ✅ same form, chip removed, next job carries `voice=None` |
| Generation without a custom voice works as before | ✅ five plain generations → 0 LoRA calls; 1773 Python + 1590 web tests pass, plus 131 integration |

The one criterion this machine cannot close is the listen test; it is called out in the demo and in the PR body rather than papered over.

PR #397, opened 2026-08-04
fix: US-26.0 dependency and supply-chain security hardening

### LINKED ISSUE #394: US-26.0: Dependency and Supply-Chain Security Hardening
## User Story

**As a** platform operator, **I want** the web application's dependency tree free of known high-severity vulnerabilities and the CI supply chain pinned to immutable references, **so that** payment and credit flows are not built on top of a frontend with live XSS, arbitrary-file-read, and SSRF-adjacent exposure.

## Description

A supply-chain triage of the twelve open Dependabot PRs (#353, #375–#385) on 2026-08-03 merged all of them, including `next` 16.2.10 → 16.2.11 which patched four verified high-severity advisories (CVE-2026-64641/64642/64645/64649). Merging those PRs closed the *known-and-patched* gap, but a fresh `npm ci && npm audit` against merged `main` still reports **11 vulnerabilities (8 high, 3 moderate)** in `web/`, none of which Dependabot will open a PR for because they are transitive or require a direct `next` upgrade it classifies as breaking.

This is a prerequisite for Stage 26, not part of it. Stage 26 introduces subscription tiers, payment integration, and credit top-up purchase (US-26.1–26.5). Shipping real payment flows onto a frontend carrying a live PostCSS XSS / arbitrary-file-read chain is the wrong ordering — the hardening lands first.

Two contributing gaps make this recurrent rather than one-off:

1. **`web/` is not covered by CI.** The required status checks on `main` are `ci (3.11)` and `ci (3.12)` — Python only. Frontend typecheck, lint, test, build, and `npm audit` never run in CI, so a green PR says nothing about the frontend. This is why the audit findings accumulated unnoticed.
2. **The advisory set is live and moving.** Two `npm audit` runs roughly an hour apart on the *same lockfile* returned 8 and then 11 vulnerabilities as new advisories were published. A point-in-time manual audit cannot hold this line; it needs a gate.

### Remediate the auto-fixable transitive advisories

Eight resolve without a direct-dependency major bump (`npm audit fix` territory), all transitive:

| Package | Severity | Issue |
|---|---|---|
| `brace-expansion` | high | DoS via unbounded expansion / bypass of the CVE-2026-14257 mitigation |
| `fast-uri` | high | Host confusion via literal backslash authority delimiter; failed IDN canonicalization |
| `ip-address` | high | (advisory via transitive chain) |
| `js-yaml` | high | YAML merge-key chains force quadratic CPU consumption |
| `undici` | high | (advisory via transitive chain) |
| `hono` | moderate | — |
| `@hono/node-server` | moderate | Path traversal in `serve-static` on Windows via encoded backslash (`%5C`) |
| `@modelcontextprotocol/sdk` | moderate | Inherits `@hono/node-server` |

### Resolve the `next` → `postcss` / `sharp` chain

Three high-severity advisories are reachable only through a `next` upgrade to `16.3.0`, which `npm audit` flags as breaking from the currently pinned `16.2.11`:

- **`postcss`** (build-time) — XSS via unescaped `</style>` in CSS stringify output; arbitrary file read and information disclosure via attacker-controlled `sourceMappingURL`; path traversal in previous-source-map auto-loading; plus an incomplete-fix follow-up to `GHSA-6g55-p6wh-862q`.
- **`sharp`** (runtime, image optimization) — inherited libvips CVE-2026-33327, CVE-2026-33328, CVE-2026-35590, CVE-2026-35591.

Evaluate whether `next@16.3.0` is a genuine breaking change for this app or only npm's conservative classification, then either upgrade or document a justified override with the residual risk stated. Do not silently `--force`.

### Pin GitHub Actions to commit SHAs

`.github/workflows/claude-code-review.yml` uses `anthropics/claude-code-action@v1.0.179`, a **mutable version tag**, and it executes with `secrets.CLAUDE_CODE_OAUTH_TOKEN`. A tag repoint changes what runs against that secret with no diff in this repo. The same applies to `actions/cache@v6`, `actions/upload-artifact@v7`, and `actions/checkout@v7` in `.github/workflows/plugin.yml` — floating major tags by design.

### Close the CI coverage gap

Add the `web/` checks to CI so this cannot silently regress. Scope here is the security gate specifically; broader frontend CI coverage is noted under Out of Scope.

## Acceptance Criteria

- [ ] `cd web && npm audit --audit-level=high` exits clean, or every remaining finding has a documented, time-bounded exception with a named residual risk
- [ ] The `next` / `postcss` / `sharp` chain is resolved — either upgraded past `16.3.0` or explicitly deferred with written justification and a follow-up issue
- [ ] `cd web && npm run typecheck && npm run lint && npm test && npm run build` all pass after the dependency changes
- [ ] `anthropics/claude-code-action` is pinned to a full commit SHA with the human-readable version in a trailing comment
- [ ] First-party `actions/*` uses in `.github/workflows/plugin.yml` are pinned to commit SHAs
- [ ] CI runs `npm audit --audit-level=high` for `web/` and fails the build on new high-severity findings
- [ ] Dependabot is configured to keep SHA pins updated (`github-actions` ecosystem) so pinning does not become staleness
- [ ] `plugin.yml` still passes on Linux / macOS / Windows after SHA pinning

## Out of Scope

- Full frontend CI coverage (typecheck / lint / test / build as required checks on `main`) — related and worth doing, but a separate change with its own branch-protection implications. Only the `npm audit` security gate is in scope here.
- The Python dependency tree. `pip-audit` / `osv-scanner` coverage for the API is not assessed in this issue.

## Dependencies

**Blocked by:** Nothing. All Stage 22–25 work is merged; only US-25.4 remains open in Stage 25 and does not touch this.

**Blocks:** Stage 26 (US-26.1 Credit Tracking, US-26.2 Subscription Tiers, US-26.3 Payment Integration, US-26.4 Credit Top-Up Purchase, US-26.5 Usage Dashboard). Payment and credit UI should not ship on a frontend with unresolved high-severity XSS and arbitrary-file-read advisories.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 26 (prerequisite, numbered `.0` per the `US-20.0` precedent)
- Dependabot triage comments on #353, #375, #376, #377, #378, #379, #380, #381, #382, #383, #384, #385
- Next.js advisories patched by #377: `GHSA-m99w-x7hq-7vfj`, `GHSA-6gpp-xcg3-4w24`, `GHSA-p9j2-gv94-2wf4`, `GHSA-89xv-2m56-2m9x`
- Audit captured against `main` @ `c227f9b` on 2026-08-03 with a clean `npm ci`

frankbria:
## US-26.0: Dependency and Supply-Chain Security Hardening

Closes #394.

`npm audit --audit-level=high` in `web/` goes from **11 vulnerabilities (8 high, 3 moderate)
→ 0**, the workflows that hold secrets stop trusting mutable tags, and a CI gate keeps it
that way.

### The eight transitive advisories

`npm audit fix` resolved all of them with no direct-dependency change — `brace-expansion`,
`fast-uri`, `ip-address`, `js-yaml`, `undici`, `hono`, `@hono/node-server`,
`@modelcontextprotocol/sdk`. Lockfile only.

### The `next` → `postcss` / `sharp` chain

The issue asked me to decide whether `next@16.3.0` is genuinely breaking or just npm's
conservative classification, and **not** to silently `--force`. It is the latter:

- It is a **minor** bump inside the same major. `npm audit` calls it "outside the stated
  dependency range" only because `package.json` pins the exact version `16.2.11`.
- **Peer requirements do not change**: `react ^18.2.0 || ^19.0.0`, and this repo is on
  19.2.7. React does not move.
- It is the **only** path. `16.2.12` still pins `postcss@8.4.31`; `16.3.0` takes `postcss`
  to `8.5.23` and `sharp` to `^0.35.3`, which is what actually closes the advisories.

Verified rather than asserted — after the upgrade, `npm run typecheck`, `npm run lint`,
1590 tests, and `npm run build` all pass.

So this is an upgrade, not a documented deferral. No override, no residual risk to carry.

### SHA pinning

All five action uses in the affected workflows now pin a full commit SHA with the version in
a trailing comment, matching the convention `glm-review.yml` already used:

| Workflow | Action | Pin |
|---|---|---|
| `claude-code-review.yml` | `anthropics/claude-code-action` | `b76a077` # v1.0.179 |
| `claude-code-review.yml` | `actions/checkout` | `3d3c42e` # v7.0.1 |
| `plugin.yml` | `actions/checkout` | `3d3c42e` # v7.0.1 |
| `plugin.yml` | `actions/cache` | `55cc834` # v6.1.0 |
| `plugin.yml` | `actions/upload-artifact` | `043fb46` # v7.0.1 |

`claude-code-action@v1.0.179` resolves to an **annotated tag object**, not a commit, so it
had to be dereferenced — pinning the tag SHA would not have worked. Every pin was then
re-verified against the API: each SHA is a real commit on that repo *and* matches the tag
named in its comment. (I initially wrote one SHA from memory; it happened to be right, which
is exactly why the verification pass exists.)

`ci.yml`'s own `actions/checkout` is pinned too, since the audit job made it a two-`uses`
file. `docker-publish.yml` still floats — see below.

### The CI gate

`web/` had **no CI at all**, which is how eleven advisories accumulated unnoticed. A new
`web-audit` job runs `npm audit --audit-level=high` against `web/`.

Its own job rather than a step in `ci`: it needs Node not Python, and it must not run once
per Python version. It reads the lockfile, so it skips a full `npm install`.

### The lockfile had to be repaired (found by `codex review`)

The first lockfile here was written by local **npm 11.6.2**, which drops the nested bundled
`@emnapi` entries this repo's **npm 10**-lineage lockfile carries. `npm ci`, the build, the
tests and `npm audit` all stayed green — but `npm ls --package-lock-only --all` went from
**exit 0 on main to `ELSPROBLEMS` on the branch**. A dependency graph that no longer
validates is not something to ship in a supply-chain PR.

Regenerated with npm 10; it exits 0 again. That check is now a CI step, since `npm ci`
cannot catch this class on its own.

My first repair attempt (`rm package-lock.json && npm install`) produced a valid lockfile
and a **177-package diff** — every caret range re-resolved to newest, which would have
bypassed the 7d/14d Dependabot cooldown this repo sets deliberately. Re-derived minimally
from main instead: **48 version changes**, each traceable to one of the eight advisories or
to the `next`/`sharp` upgrade, plus two new `sharp` platform targets.

### Dependabot

No change needed — `.github/dependabot.yml` **already** configures the `github-actions`
ecosystem at `/`, which is what keeps SHA pins from becoming staleness. Verified rather than
assumed.

### Known limitations

- **The `web-audit` job is not a *required* check.** Making it one is a branch-protection
  change on `main`, which I can't make from a PR. It runs and reports on every PR; until
  someone marks it required, it is advisory. Worth doing right after merge.
- **`docker-publish.yml` still uses floating tags** (`docker/*@v4`/`v6`/`v7`, `actions/checkout@v7`),
  and `ci.yml` still floats `actions/setup-python@v6.3.0` / `astral-sh/setup-uv@v8.3.2`.
  The issue scoped pinning to `claude-code-review.yml` and `plugin.yml`, so I did not widen
  it — but the same tag-repoint risk applies, and `docker-publish.yml` pushes images. Worth
  its own issue.
- **The advisory database is live.** A newly published advisory can fail a PR that changed
  nothing. That is the gate working as intended, not a flake — but it will surprise someone.
- **The Python dependency tree is unassessed** (`pip-audit` / `osv-scanner`), explicitly out
  of scope here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
           --health-retries 5
 
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v6.3.0
@@ -56,3 +56,30 @@ jobs:
         run: uv run pytest -m integration
         env:
           ACEMUSIC_TEST_MONGODB_URL: mongodb://localhost:27017
+
+  # US-26.0: web/ had no CI at all, which is why eleven npm advisories (eight high)
+  # accumulated unnoticed. This is the security gate specifically — typecheck, lint,
+  # test and build are still local-only, deliberately (they carry branch-protection
+  # implications this change does not take on).
+  #
+  # Its own job rather than a step in `ci`: it needs Node, not Python, and it must not
+  # run once per Python version.
+  web-audit:
+    name: web npm audit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Set up Node
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: "22"
+          cache: npm
+          cache-dependency-path: web/package-lock.json
+
+      # `npm audit` reads the lockfile and does not need node_modules, so this skips
+      # a full install. The advisory database is queried live, so a newly published
+      # advisory can fail a PR that changed nothing — that is the point of a gate.
+      - name: Audit web dependencies
+        working-directory: web
+        run: npm audit --audit-level=high
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,3 +83,11 @@ jobs:
       - name: Audit web dependencies
         working-directory: web
         run: npm audit --audit-level=high
+
+      # A lockfile can be internally inconsistent while `npm ci` still succeeds — an
+      # npm version that drops bundled nested entries leaves declared dependencies
+      # with no corresponding package entry. This PR shipped exactly that for one
+      # commit, and only a reviewer caught it.
+      - name: Check the web lockfile is internally consistent
+        working-directory: web
+        run: npm ls --package-lock-only --all > /dev/null
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -33,7 +33,7 @@ jobs:
         if: |
           github.event.pull_request.changed_files >= 5 ||
           steps.calc.outputs.total >= 20
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 1
 
@@ -43,7 +43,7 @@ jobs:
           github.event.pull_request.changed_files >= 5 ||
           steps.calc.outputs.total >= 20
         id: claude-review
-        uses: anthropics/claude-code-action@v1.0.179
+        uses: anthropics/claude-code-action@b76a0776ae74036e77cd11018083743453d7ad35 # v1.0.179
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           prompt: |
--- a/.github/workflows/plugin.yml
+++ b/.github/workflows/plugin.yml
@@ -49,7 +49,7 @@ jobs:
         shell: bash
 
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Install Linux build dependencies
         if: runner.os == 'Linux'
@@ -68,7 +68,7 @@ jobs:
 
       # The JUCE shallow clone is the slowest part of a cold configure.
       - name: Cache JUCE
-        uses: actions/cache@v6
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: plugin/build/_deps
           key: juce-${{ runner.os }}-${{ hashFiles('plugin/CMakeLists.txt') }}
@@ -140,7 +140,7 @@ jobs:
 
       - name: Upload VST3
         if: always()
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: acemusic-vst3-${{ runner.os }}
           path: plugin/build/AceMusicPlugin_artefacts/**/VST3/
--- a/docs/demos/us-26.0-supply-chain.md
+++ b/docs/demos/us-26.0-supply-chain.md
@@ -1,0 +1,346 @@
+# US-26.0: Dependency and Supply-Chain Security Hardening
+
+*2026-08-04T07:44:53Z*
+
+Stage 26 introduces subscription tiers, payment integration and credit top-up purchase. Shipping real payment flows onto a frontend carrying a live PostCSS XSS / arbitrary-file-read chain is the wrong ordering, so the hardening lands first.
+
+Five things to show:
+
+1. `npm audit --audit-level=high` exits clean
+2. the `next` / `postcss` / `sharp` chain is genuinely resolved, not deferred
+3. typecheck, lint, test and build still pass on the upgraded tree
+4. every action is pinned to an immutable commit SHA
+5. CI now fails on a new high-severity finding, so this cannot silently regress
+
+## Before: what `main` carries
+
+Run against `main`'s lockfile, extracted into a scratch directory so the working tree is untouched.
+
+```bash
+SC=/tmp/claude-1000/-home-frankbria-projects-auto-music-studio/e9e3c62d-a729-4893-b40e-b671581d939c/scratchpad; rm -rf $SC/beforeweb && mkdir -p $SC/beforeweb && cd /home/frankbria/projects/auto-music-studio && git show main:web/package-lock.json > $SC/beforeweb/package-lock.json && git show main:web/package.json > $SC/beforeweb/package.json && cd $SC/beforeweb && npm audit --audit-level=high --json 2>/dev/null | python3 -c "
+import json,sys
+d=json.load(sys.stdin)
+for n,v in sorted(d[chr(118)+chr(117)+chr(108)+chr(110)+chr(101)+chr(114)+chr(97)+chr(98)+chr(105)+chr(108)+chr(105)+chr(116)+chr(105)+chr(101)+chr(115)].items()):
+    fa=v[chr(102)+chr(105)+chr(120)+chr(65)+chr(118)+chr(97)+chr(105)+chr(108)+chr(97)+chr(98)+chr(108)+chr(101)]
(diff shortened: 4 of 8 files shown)


frankbria:
## Cross-family review (pre-PR)

`opencode` (GLM) is still returning server-side `UnknownError`s, so this is
`codex review --base main`.

### Finding: [P2] lockfile internally inconsistent — CONFIRMED, fixed in 5a6da29

> `cd web && npm ls --package-lock-only --all` now exits with `ELSPROBLEMS` because
> `@tailwindcss/oxide-wasm32-wasi` still declares `@emnapi/core` and `@emnapi/runtime`,
> but the corresponding nested package entries were removed.

Real, and mine. Verified both sides rather than taking it on trust — extracted main's
lockfile into a scratch directory and ran the same command: **main exits 0, my branch
exited 1** with seven `missing: … required by …` errors.

Root cause: my local **npm 11.6.2** drops the nested bundled `@emnapi` entries under
`oxide-wasm32-wasi`; this lockfile's lineage is **npm 10**, which preserves them.
Regenerated with npm 10 → `npm ls --package-lock-only --all` exits 0 again.

`npm ci` and the build had been passing throughout, which is exactly why this needed a
reviewer to catch — and why it is now a CI step (`0b35ff4`) rather than a lesson.

### A second problem the fix surfaced

My first repair attempt was `rm package-lock.json && npm install`. That produced a valid
lockfile but re-resolved every caret range: **177 packages moved**, far outside this
issue, and it would have bypassed the 7d/14d Dependabot cooldown this repo sets
deliberately for supply-chain safety — the opposite of what a hardening PR should do.

Re-derived minimally from main's lockfile instead. Now **48 version changes**, each one
traceable to one of the eight advisories or to the `next`/`sharp` upgrade, plus two new
`sharp` platform targets.

Re-verified after the repair: `npm ci`, typecheck, lint, 1590 tests, production build all
pass; `npm audit --audit-level=high` reports 0.

No other findings were reported.

frankbria:
## Demo (Phase 11)

`docs/demos/us-26.0-supply-chain.md` — every acceptance criterion mapped to evidence, all ✅.

Two things I made a point of proving rather than asserting:

**The `next@16.3.0` decision.** The issue asked me to judge whether it is genuinely breaking. Demo shows the three facts that decide it: identical peer requirements on both sides (`^18.2.0 || ^19.0.0`, repo is on 19.2.7), a minor bump inside the same major, and `16.2.12` still shipping `postcss@8.4.31` — so `16.3.0` is the *only* version that closes the advisory. npm being conservative, not a real break.

**The gate actually fails.** A gate nobody has watched fail is not a gate. Both steps run against a deliberately broken throwaway lockfile (never the real one): `npm audit --audit-level=high` → exit 1, and the consistency check with a package entry deleted → `ELSPROBLEMS`, exit 1.

Also in the demo: the before/after audit (11 → 0), the minimal 48-package diff with every entry traceable to an advisory or the `next`/`sharp` upgrade, all five SHA pins verified against the API, and this PR's own checks green — including the plugin build on Linux/macOS/Windows, which is what proves the pins work.

One thing the demo is explicit about: **`web npm audit` is not a *required* check.** That's branch protection on `main`, which a PR can't set. It runs and reports on every PR but can't block a merge until someone marks it required — worth doing right after this lands.

frankbria:
## On the `eslint-config-next` skew

The GLM review carried this forward across passes as a non-defect. I checked it rather than accepting that label, and agree — but for a reason worth writing down.

`next` is now `16.3.0`; `eslint-config-next` is `16.2.10`. My change widened that from a patch skew to a minor one. Three facts decide it:

- **The skew is pre-existing, not introduced here.** `main` was already `next@16.2.11` / `eslint-config-next@16.2.10`. These are separate packages that Dependabot updates independently, each under this repo's 7d/14d cooldown, so they drift by design.
- **It is a devDependency** — lint rules only, no runtime or build surface.
- **Lint passes** (`npm run lint` → exit 0), which is the thing the skew could plausibly break.

So I'm deliberately *not* bumping it. Adding a devDependency bump that no advisory requires would put unrelated churn into a security PR and step around the cooldown policy — which is the exact mistake I already made and corrected once here, when my first lockfile repair moved 177 packages.

Dependabot will align it on its own schedule. Flagging it in the PR rather than silently leaving it.

PR #398, opened 2026-08-04
feat: US-26.1 credit tracking, refund-on-failure, and the visible balance

### LINKED ISSUE #329: US-26.1: Credit Tracking and Deduction
## User Story

**As a** musician, **I want** my credits tracked accurately per action, **so that** I know what each generation costs and how many credits I have left.

## Description

Every generative action has a defined credit cost. The system deducts credits atomically before processing begins and displays the remaining balance. If the musician has insufficient credits, the action is denied with a clear explanation.

## Functional Requirements

- Credit costs per action: generation (1), extend (1), cover (1), mashup (2), stems (1), MIDI (1), remaster (0.5), mastering (2–5), LoRA training (10), video (5–10)
- Credits are deducted atomically before the job begins (no double-deduction on retry)
- GET `/api/v1/credits/balance` returns current credit balance
- Credit balance displayed in the app header/sidebar at all times
- Insufficient credits returns 402 with: required credits, current balance, and upgrade prompt
- Credit refund on job failure (automatic)

## Acceptance Criteria

- [ ] Generating a song deducts exactly 1 credit
- [ ] Mashup deducts exactly 2 credits
- [ ] Attempting an action with insufficient credits returns a 402 with clear messaging
- [ ] Failed jobs result in automatic credit refund
- [ ] Credit balance is visible at all times and updates in real time after each action

## Dependencies

**Blocked by:** Stage 21. First story of Stage 26.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 26, US-26.1

frankbria:
## US-26.1: Credit Tracking and Deduction

Closes #329.

Three of the five acceptance criteria were already satisfied by earlier stories (atomic
deduction, per-action costs, the 402). The other two were not, and neither was a small
addition.

### AC4 exposed two real defects

**The processor never refunded a failed job.** It marked the job `FAILED` and stopped.
Only three narrow paths gave anything back — full-song unperformed sections, mastering
pre-flight rejection, voice-training failure — so an ordinary failed generation simply
kept the credit.

**`refund_credits()` wrote no ledger row.** It moved the balance and nothing else, so most
refunds were invisible in `/users/me/credits` and the ledger could not answer *"how much
has this job already been paid back?"*.

Fixing the second is what makes the first safe. `refund_credits` now ledgers the movement,
so `amount_owed_for_job = charges − refunds`, and `refund_failed_job` pays only the
remainder. That is **idempotent by construction** rather than by a flag: its own refund is
a ledger row, so a second call finds nothing owed, and the three handlers that already
refund partial work cannot be paid twice.

`_mark_failed` is the single point every job type converges on, so this needed no
per-handler wiring — and it deliberately does not refund if the status write itself failed.

### The refund that must *not* be ledgered

The request paths deduct → create the job → record the charge **last**. So when their
compensating `except` runs, the charge was never recorded, and writing a refund row would
leave a lone positive movement: a credit the user appears to have been granted from
nowhere.

Those call `reverse_unrecorded_charge` — a separate, self-documenting function rather than
a `ledger=False` flag, because a flag gets passed wrong eventually and this is money.

### AC1: three actions that were free now cost credits

`stems` and `midi` now cost 1 credit each and `remaster` 0.5. Both routers had documented
them as *"non-generative local CPU work, so no credits are deducted"* — that reasoning was
not wrong; the story simply makes a different pricing decision.

No migration concern: the product is not live, so there is no installed base to carry
across the change.

Two boundaries keep it fair, both tested:
- a **cached** result costs nothing — the credit buys the separation, not the lookup, so
  polling the endpoint cannot bill repeatedly;
- a request **rejected for format** is refused before the charge.

Crop and speed stay free.

### AC5: the balance, everywhere

The web app never called the credits endpoint at all. This adds the whole path: a BFF
proxy, a client, a root-layout context, and the sidebar display — a context for the same
reason notifications has one, since the balance renders on every page while the actions
that spend credits live all over the app.

`refresh()` fires from `use-generation` and `use-clip-edit` the moment a job is accepted
(the charge lands at enqueue) and on a 402, where the server has just reported a balance we
may not have.

### Also here

- `GET /api/v1/credits/balance`, deliberately separate from `/users/me/credits`: the
  sidebar polls it on every page and does not need fifty ledger rows each time.
- One 402 shape for every billed action, carrying `balance`, `required`, a readable
  `message` and `upgrade_url`. An error that only says "no" leaves the musician stuck.
- `charge_and_create()` replaces the deduct/create/compensate/ledger dance that was
  hand-written in five places and would have been written three more times here.

### A mutation-found gap, worth naming

All 20 pricing tests passed first run. Mutating anyway: **deleting the compensating
reversal from `charge_and_create` killed zero tests** — the existing coverage for
"charged ⇒ a job exists or the credit is returned" drove the hand-rolled router copies, not
the new helper three endpoints had just started depending on. Six tests later the same
mutant fails two of them. Dropping the clamp on `amount_owed_for_job` and dropping the
`FAILED`-status guard were also checked; both die.

### Tests

- `tests/test_credit_refunds.py` (25) — ledgering, `amount_owed_for_job` including the
  over-refund clamp, refund-exactly-once against each partial-refund shape, the processor
  end-to-end (failure refunds, success does not), and `charge_and_create`'s failure paths
  including `CancelledError`.
- `tests/test_credit_pricing_api.py` (20) — the three new costs, their ledger rows, the
  cached-is-free and rejected-is-free boundaries, crop/speed still free, the 402 shape, and
  the balance endpoint.
- Web (17 new) — the client, the 402 parser, and the sidebar display.

No existing assertion was weakened. The four pre-existing tests that broke were asserting
the old invisible-refund behaviour.

### Known limitations

- **`upgrade_url` points at `/settings/billing`, which does not do anything yet** — US-26.4
  (credit top-up purchase) is what makes it useful. Served from the API rather than
  hardcoded in the client so it can move without a frontend release.
- **The other five `charge_and_create`-shaped call sites still hand-roll the pattern**
  (generation, iterative, videos, mastering, batch mastering). Adopting the helper there is
  a behaviour-preserving refactor I deliberately kept out of a money-behaviour PR.
- **Refunds are best-effort at the edges.** A refund that itself fails is logged, never
  masking the original job failure — consistent with the existing handlers.


frankbria:
## Cross-family review (pre-PR)

`opencode` (GLM) is still failing server-side, so this is `codex review --base main`.
**Three findings, all real, all fixed in 5a90a31.**

### [P1] The refund guard test could no longer reach its guard — CONFIRMED

> `refund_credits()` … now raises `TypeError` before the function body runs, so the
> existing CI guard … no longer verifies the `ValueError` path and currently fails.

Exactly right, and worse than it sounds: **two tests were failing in the default
(non-integration) suite** and I had not seen them. I ran `-m integration` over every file
I touched, plus the full unit suite *before* these changes — never the unit suite after.
That was a red CI waiting to happen, caught by review rather than by me.

The deeper point is the one worth keeping: a guard test that can no longer reach its guard
still *looks* like coverage. Making a parameter required silently disarms every existing
call, including the ones in tests.

Fixed, and `reverse_unrecorded_charge` — which had no such guard test at all — now has one.

### [P2] The balance went stale after a refund — CONFIRMED (both)

> the new worker refund path restores the balance on the backend but this hook's failed
> polling branch does not call `refreshCredits()`.

Correct. This PR is what *introduced* the refund-on-failure, so it also introduced the
staleness: a musician whose charged job then failed kept seeing the deduction for work they
were no longer charged for, until something unrelated triggered a refresh. Both
`use-generation` and `use-clip-edit` now refresh on the failure branch.

The refresher is reached through a ref, so `poll()` is not re-created every time the credits
context renders — a memoised callback with a new dependency would otherwise restart the poll
loop.

`use-clip-edit.test.tsx` covers it; reverting just the failure-branch refresh fails it.

### Verification after the fixes

1775 unit tests, 152 integration across the touched areas, 1609 web tests, typecheck, lint,
black and ruff all clean.

frankbria:
## Demo (Phase 11)

`docs/demos/us-26.1-credit-tracking.md` — all five acceptance criteria ✅, against the real app, real MongoDB, real tokens and balances. ACE-Step appears once, stubbed, to make a job fail on purpose so the refund can be watched.

| Criterion | Verdict |
|---|---|
| Song deducts exactly 1 credit | ✅ shown with the full cost table |
| Mashup deducts exactly 2 | ✅ |
| Insufficient credits → 402 with clear messaging | ✅ `required`, `balance`, readable `message`, `upgrade_url`; nothing queued |
| Failed jobs are automatically refunded | ✅ real worker failure, both movements ledgered, refunding again pays nothing |
| Balance visible at all times, updating after each action | ✅ sidebar of a real session, collapsed **and** expanded |

Two things I made a point of showing rather than asserting:

**What keeps pricing a previously-free action fair.** A cached result costs nothing (the credit buys the separation, not the lookup — otherwise polling would bill repeatedly), and a format-rejected request is refused before the charge.

**The mutation check, not just the passing suite.** Deleting the compensating reversal from `charge_and_create` originally killed *zero* tests; it now kills two, and the demo runs that mutant live. A suite that passes first time is exactly when to check it bites.

The demo's "What is not here" is explicit about the thing that needs a human decision: **stems, MIDI and remaster were free by documented design and now cost credits.** Nobody has decided how existing users hear about that, and it shouldn't be discovered from a balance.

frankbria:
## Carried-forward review items — three fixed in 43056ab

The review thread listed four items as "still open". Two of them were defects **this PR introduced**, so they should not have been carried any further:

**`CreditTransaction` had no `job_id` index.** `amount_owed_for_job()` sums every row for one job, and this PR made it run on *every job failure* via `refund_failed_job` — against an append-only ledger, the one collection that only ever grows. That is a collection scan per failure. Indexed.

**The mastering comment asserted the opposite of the code.** It still said the fallback-exhaustion path "fails without a refund", which stopped being true the moment `_mark_failed` began refunding unconditionally. A comment that contradicts the code is worse than no comment.

**`parseInsufficientCredits` had no production caller** — and this one matters more than "dead code". The hooks still went through `editing.ts`'s older `extractCredits`, which drops both `message` and `upgrade_url`. So the richer 402 this PR added never reached the UI, which is the *entire point* of AC3's upgrade prompt. `extractCredits` now parses the shared shape (falling back for endpoints that have not adopted it), and `use-clip-edit` shows the server's sentence rather than re-wording it.

Worth noting how the last one surfaced: three test doubles needed the widened result type, and `vi.fn().mockResolvedValue` is untyped — so one of them rendered a **blank error alert** rather than failing to compile. Caught by the suite, not the compiler.

The fourth item — the 402 shape still being split across `generation.py`, `iterative.py`, `mastering.py` and `videos.py` — stays deferred and is named in the PR's Known Limitations. Those four still hand-roll the older dict; adopting `charge_and_create` there is a behaviour-preserving refactor I am keeping out of a money-behaviour PR.

PR #399, opened 2026-08-05
feat: US-26.2 subscription tiers, backend feature gating, and monthly credit reset

### LINKED ISSUE #330: US-26.2: Subscription Tiers and Feature Gating
## User Story

**As a** musician, **I want** to understand what my subscription tier includes and see clear prompts to upgrade when I hit a limit, **so that** I can make informed decisions about my plan.

## Description

Two tiers exist: Free (50 credits/month, limited features) and Pro (500 credits/month, all features). Pro-only features show a Pro badge and an upgrade prompt for Free users. Feature gating is enforced both on the frontend (UI indicators) and backend (API authorization).

## Functional Requirements

- Free tier: 50 credits/month, MP3 download only, Studio view-only, no stems/MIDI export, no mastering, no distribution, VST3 preview-only, 720p watermarked video, no custom voice models
- Pro tier: 500 credits/month, all formats, full Studio editing, stems/MIDI, mastering, distribution, full VST3, 1080p/4K video, custom voice models, priority queue
- Pro badge displayed on locked features in the UI
- Clicking a locked feature shows an upgrade modal explaining the benefit
- Backend enforces tier restrictions — Free users hitting Pro endpoints receive 403 with upgrade guidance
- Credits reset monthly on the subscription anniversary date

## Acceptance Criteria

- [ ] Free user sees Pro badges on restricted features
- [ ] Free user clicking a Pro feature sees an upgrade modal (not an error)
- [ ] Pro user can access all features without restriction
- [ ] API returns 403 for Free users attempting Pro-only actions
- [ ] Credits reset to tier allocation on monthly anniversary

## Dependencies

**Blocked by:** US-26.1 (prior story in this stage). Part of Stage 26: Credits & Subscription System, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 26, US-26.2

frankbria:
## US-26.2: Subscription Tiers and Feature Gating

Closes #330.

AC1 (Pro badges) and AC3 (Pro users unrestricted) were already satisfied by US-17.2. The
other three were not, and two of them had **no implementation at all**.

### AC4: there was no backend enforcement — none

Not one endpoint checked the subscription tier. A free account that called
`POST /clips/{id}/stems` directly got stems; the same for mastering, DAW export, voice
training and distribution. The Pro badges in the UI stopped an honest user and nobody else.

`services/tiers.py` is now the single table of what each tier unlocks, and every Pro-only
action declares it:

| Capability | Gated at |
|---|---|
| `stems`, `midi` | the extraction endpoints |
| `mastering` | single **and** batch |
| `distribution` | SoundCloud upload |
| `studio_editing` | Studio mixdown, DAW export |
| `voice_models` | voice training, **and** generating with a custom voice |

Three decisions worth stating:

- **Gated per action, not per router.** The free tier is "Studio *view-only*" and "no stems
  export", not "cannot see your own data". Reads stay open, so a downgraded account can
  still list and delete voices it trained.
- **The tier is read from the database, not the token claims.** Claims outlive a plan
  change by up to the token TTL — which would let a downgrade keep Pro, and make someone
  who has just paid wait for it to expire. Tested in both directions.
- **The table fails closed.** Anything not exactly `"pro"` is free (matching the frontend's
  rule), and the free tier is defined by what it *lacks*, so a newly added capability is
  locked by default rather than leaking. There is a test that fails if one is ever added
  outside the gated set.

Three gates depend on a request **argument** rather than the endpoint — a custom voice, a
video resolution, an export format — which a router dependency cannot see. So
`require_capability` (the dependency) delegates to `require_tier_capability` (imperative):
one implementation, two call shapes, one 403 body.

The 403 mirrors the 402's shape — what is locked, what it would do, where to go — so a
client presents "you cannot do this yet" the same way whether the obstacle is money or plan.

### AC2: clicking a locked feature did nothing

The story asks for an upgrade modal, "not an error". What it replaced was neither: locked
menu items rendered `disabled`, so the click never fired. **Silence is the worst of the
three** — it reads as a broken menu rather than a boundary.

Locked items stay clickable now, and the interception moved up into `useSongActions` so
every entry point behaves the same rather than each menu re-deciding. The prompt leads with
the feature that was actually reached for ("Mastering is a Pro feature"), because it should
answer *why can't I do this* before it sells anything.

`lib/tiers.ts` keys its copy by the same capability strings the API's 403 uses, so a lock
predicted client-side and one discovered server-side render identically — and an unknown
capability falls back to usable copy rather than crashing a dropdown.

### AC5: credits reset monthly

Applied **lazily on read** rather than by a scheduler: a cron that misses its window leaves
someone short until the next one, and there is no window to miss if the answer is derived
from the dates every time it is asked for. Wired into both `/credits/balance` and
`/users/me/credits` so the two cannot disagree.

Anchored to the signup anniversary rather than the calendar month, so nobody gains or loses
a partial period by signing up on the 3rd. Three things it deliberately does **not** do:

- **claw back** — a balance above the allocation is left alone, so a top-up purchase
  (US-26.4) survives the reset;
- **pay for missed periods** — six months away returns one month's allowance, not six;
- **pay accounts that predate the field** — `credits_reset_at is None` means "never run",
  which is backfilled rather than treated as infinitely overdue. Otherwise every existing
  account collects a windfall on next read.

The date arithmetic is the risky part, so `next_reset_due` is pure and tested directly:
Jan 31 clamps to Feb 28 (29 in a leap year) and **returns to the 31st in March** rather
than sticking at 28 forever.

### Tests

- `tests/test_tiers.py` (47) — the table itself, including every way of being not-quite-Pro.
- `tests/test_tier_enforcement_api.py` (19) — every Pro action refused for free and allowed
  for Pro, over the real HTTP surface; the 403 body; that a refusal costs nothing and
  queues nothing; upgrade/downgrade taking effect mid-token; and that the voice gate does
  not leak whether the voice exists.
- `tests/test_credit_reset.py` (13) — the date arithmetic, plus each of the three things the
  reset must not do.
- Web (11 new) — the modal, and the locked-click path through `useSongActions`.

Three existing web tests asserted the old inert behaviour and now assert the new contract.
`ClipCard` needed the tier mocked: `useSongActions` reads it, and the profile fetch was
landing ahead of the call those tests assert on.

### Review and demo (added after the original description)

**Two Major review findings**, both raised by the GLM cross-family pass and the Claude bot:

- `GET /clips/{id}/stream?format=wav` performed the same on-the-fly conversion as
  `/clips/{id}/audio` with no tier check — the refusal was one URL away from being undone.
  Fixed, including the **anonymous** path: `/stream` serves public clips without a login, so
  gating only authenticated callers would have made signing out the way around the gate.
  `require_tier_capability` now takes `user_id: str | None` and treats no account as no plan.
- `apply_monthly_reset` was a read-modify-write: three full-document `save()` calls pushed a
  stale in-memory user back over the document, so a deduction landing between the read and
  the save was silently reverted. The backfill branch was the sharp one — every new account
  takes it on its first read, exactly when it is likeliest to be spending its first credit.
  Now the backfill `$set`s only the anchor while it is still unset, and the reset `$inc`s the
  grant filtered on the anchor it read, so concurrent overdue reads serialise into one grant
  and one ledger row.

**CI was red on `ci (3.12)`: 113 integration tests.** Every one was a pre-existing test of a
feature this story makes Pro-only, running as the default free user. Their user helpers now
create a Pro account; the refusals are tested in `test_tier_enforcement_api.py`. The unit job
was green throughout, which is why this was invisible locally.

**The demo found three more ungated Pro routes** — the same "declared capability, unchecked
API path" class this PR set out to close, on routes nobody had listed:

| Route | Was | Now |
|---|---|---|
| `POST /batch/stems` | 202 for a free account — stems for 50 clips at once, while the single-clip route refused one | 403 `stems` |
| `POST /batch/export` (wav/wav32/flac) | 202 — the batch version of the download just gated | 403 `lossless_export` |
| `POST /clips/{id}/export/daw` | 202 — same bundle as the gated `/studio/export/daw`, different router | 403 `studio_editing` |

Full evidence for all five acceptance criteria, against a real stack and a real browser:
`docs/demos/us-26.2-subscription-tiers.md`.

### Known limitations

- **`high_res_video` and `lossless_export` are now enforced** — `videos.py` gates on the
  requested resolution, and the lossless gate covers `/clips/{id}/audio`,
  `/clips/{id}/stream` (authenticated and anonymous) and `POST /batch/export`. The earlier
  version of this description listed them as declared-but-unenforced; that is no longer true.
- **Two routes were deliberately left ungated** (#403): `POST /releases/{id}/prepare/{target}`
  builds a distribution bundle for *manual* submission, and `POST /clips/{id}/remaster` is a
  local loudness edit that costs credits. Neither has a gated sibling doing the same work, so
  whether they are Pro is a product decision rather than an inconsistency to close in a review
  pass.
- **`/settings/billing` still does not do anything** — US-26.4 is what makes the upgrade
  path real. Every 403 and the modal point at it so there is one place to change.
- **The monthly reset only runs when someone reads their balance.** An account that never
  opens the app is not topped up until it does, which is harmless but worth knowing before
  anyone reports it as a bug.
- **`SongDetail` and `useSongActions` each fetch `/api/users/me`** (#402) — correctness-neutral,
  deferred rather than fixed here since the real answer is a shared tier context.


frankbria:
## Review triage — round 2 (`41f4f26`)

Cross-family review on record: **GLM (z.ai) via the GLM Review workflow**, plus the Claude bot's general pass. Both rounds triaged below.

| # | Severity | Finding | Action |
|---|---|---|---|
| 1 | Major | `apply_monthly_reset` is a non-atomic read-modify-write; a deduction racing the lazy reset is silently reverted (raised twice — GLM inline `credits.py:424`, Claude review §2) | **Fixed** |
| 2 | Major | `GET /clips/{id}/stream?format=wav` re-encodes with no tier check — the same bypass just closed for `/audio` (GLM inline `clips.py:486`, Claude follow-up §1) | **Fixed** |
| 3 | Suggestion | `SongDetail` and `useSongActions` each call `useSubscriptionTier()`, doubling `/api/users/me` per mount (Claude review §3) | **Deferred** — see rebuttal |
| 4 | Critical | *Not from a reviewer*: `ci (3.12)` was red — 113 integration tests failing `403 upgrade_required` | **Fixed** |

### 1. Atomic monthly reset

Three full-document `save()` calls wrote a stale in-memory user back over the document. The backfill branch is the sharp one: every new account takes it on its first read, which is exactly when it is likeliest to be spending its first credit, so a `deduct_credits` landing between the read and the save was reverted and the generation came out free.

Now: the backfill `$set`s only the anchor, and only while it is still unset. The reset `$inc`s the grant filtered on the anchor it read, so two concurrent overdue reads serialise — one grants and writes one ledger row, the other matches nothing and returns a re-read. `$inc` rather than `$set` on the balance for the same reason `deduct_credits` uses it: a charge landing mid-reset composes instead of being overwritten. This also closes the duplicate-`monthly_reset`-row case Claude's §2 raised.

### 2. The `/stream` lossless bypass

Confirmed and fixed. The gate now sits on both conversion paths.

It also covers the **anonymous** path, which the finding raised as an open question: `/stream` serves public and unlisted clips without a login, so gating only authenticated callers would have made *signing out* the way around the gate. `require_tier_capability` now takes `user_id: str | None` and treats no account as no plan. Four tests: the authenticated bypass, the anonymous bypass, that playback in a clip's stored format is still never gated, and Pro.

Both new gate tests were mutation-checked — with the gate disabled, both fail.

### 3. Deferred: the duplicate profile fetch

**Rebuttal:** correctness-neutral (the reviewer's own framing: "not incorrect, just wasteful"), and request efficiency is outside issue #330's acceptance criteria. The honest fix is a shared tier context rather than threading `isFreeTier` through `useSongActions`' several call sites, which is a refactor with its own blast radius and does not belong in a PR this size. Filed as a follow-up.

### 4. The 113 red tests

Not raised by a reviewer — CI was red on `ci (3.12)` and the unit job is green, so this was only visible in the integration run.

Every failure was a **pre-existing test of a feature US-26.2 just made Pro-only**, running as the default free user: stems, midi, mastering (single and batch), distribution, Studio, voice models, and the pricing suite. They assert what an action *does* and what it *costs*; a 403 raised before the work starts tests neither.

Their user helpers now create a Pro account. This is the story's contract change, not a test edited to reach green — issue #330 makes each of these Pro-only, and the refusal is tested where it belongs, in `test_tier_enforcement_api.py`. No assertion was weakened; no test file asserts free-tier behaviour ("free" in `test_credit_pricing_api.py` means free *of charge*).

Verified locally against a real MongoDB: **169 previously-failing integration tests now pass**, plus 48 in the tier-enforcement and credit-reset suites. Web: 1623 tests, `tsc --noEmit`, and eslint all clean.

frankbria:
## Review triage — rounds 3 and 4 (`6c2c7a5`)

Both reviewers converged on the same theme this round: not a missing gate, but a gate that
refuses too much. Both are fixed.

| # | Severity | Finding | Action |
|---|---|---|---|
| 1 | Major | The menu locks `download-wav`/`download-flac` unconditionally, but the API gates lossless export only on a genuine *conversion* — so a free account is refused a download the backend would serve (Claude rounds 3 and 4) | **Fixed** |
| 2 | Suggestion | `POST /batch/export` has no native-format carve-out, unlike `GET /clips/{id}/audio` (Claude round 4 §1) | **Deliberate** — now stated in code and test |
| 3 | — | GLM re-review of `41f4f26` and of `4ff455c`/`2bbb1f8`: **no new defects**; both earlier Major findings verified genuinely closed, all three new gates verified to close their bypass with no over-gate and no hole | Noted |

### 1. The menu refusing what the API allows

Confirmed and fixed. `POST /clips/upload` is not tier-gated, so a free musician can own a
wav they uploaded themselves; `native_format(clip)` is then `"wav"`, the API's gate does not
fire, and `GET /clips/{id}/audio?format=wav` returns the file. The menu locked it anyway and
opened the upgrade modal.

Worth being precise about severity: this was **introduced by this PR**. Before it, downloads
were not gated at all. So it is a regression, not a pre-existing gap — which is why it is
fixed here rather than deferred like the duplicate-fetch item.

The fix puts the lock decision in one place. `isSongActionLocked(action, {isFreeTier,
nativeFormat})` is now used by **both** the click handler (`useSongActions`) and the padlock
(`SongActionsMenu`). Those two had drifted precisely because each decided for itself — the
same shape of bug as the backend's "declared capability, unchecked path", one layer up.
`SongDetail` passes `clip.format` down.

Five unit tests plus two hook tests: locked for a real conversion, not locked for the clip's
own format (case-insensitively), non-download Pro actions still locked regardless of format,
and locked when the format is unknown — fail-closed. Both new carve-out tests were
mutation-checked; with the carve-out disabled, both fail.

### 2. Batch export: deliberate, not an oversight

Not fixed, and now said out loud in the code and in the test that demonstrates it.

One clip has one stored format, so "is this a conversion?" has an answer. A batch is a set of
clips with potentially different formats and a **single** status code for the whole request.
A mixed batch would have to either refuse clips it should serve or serve clips it should
refuse; 403-vs-202 cannot express "some of each". Refusing the lossless request is the honest
simplification, it errs toward over-gating rather than leaking, and the ungated single-clip
download is still there for the wav someone uploaded.

The round-4 note that the code comment "doesn't currently say this is deliberate" was fair —
it does now, and `test_a_free_account_is_refused_a_lossless_batch_export` uses a natively-wav
clip specifically so the asymmetry is visible in the test rather than surprising in
production.

### Verification

- Web: **1630 tests**, `tsc --noEmit`, eslint — all clean.
- API: full unit suite green; `test_tier_enforcement_api.py` 46 passed against a real MongoDB.
- The three earlier CI-red suites and every suite touched by the fixture change remain green.

frankbria:
## Review triage — rounds 5 and 6 (`0db2973`) — final

| # | Severity | Finding | Action |
|---|---|---|---|
| 1 | Major | `ClipCard` intercepted a locked click but never rendered `UpgradeModal`, and still marked items `disabled` — so on that surface the click went **silent**, the exact failure AC2 names | **Fixed** (`b7548b9`) |
| 2 | Suggestion | `ClipCard`'s ⋯ menu badges only *Open in Editor*; its other Pro items render without one | **Deferred** → #404 |

### 1. The clip card

This one is worth naming plainly: **the demo missed it.** AC2 was marked verified on
evidence from song detail alone, and the other entry point was broken — a free musician
clicking a Pro action on any clip card got nothing at all. `ClipCard` dispatched through the
same `useSongActions`, so the interception fired; there was simply no modal for it to open,
and two menus still rendered `disabled`, so most clicks could not fire in the first place.

Fixed, and **re-verified in the browser** on `/create` rather than by test alone:
`data-locked="true"` with no `aria-disabled`, and clicking opens *"Studio editing is a Pro
feature"*. Screenshot and a before/after table are in the demo doc, recorded as a demo
failure rather than folded into "AC2 verified". Mutation-checked: with the modal removed,
the new test fails.

### 2. The ⋯ menu's missing badges — deferred to #404

Verified **pre-existing**: `main` renders `send-mastering` and `export-daw` badge-less too,
so this PR did not introduce it. Behaviour is already correct — `emitMenu` runs the same
`isSongActionLocked`, so those clicks do open the prompt (that is finding 1's fix). What is
missing is the *advance* visual signal.

The cause is that `ClipCard` hardcodes its own menu and its own `DOWNLOAD_ITEMS` instead of
rendering from the shared registry; the file's own comment predicted this drift ("keep those
in sync … by hand until a third consumer justifies a shared renderer"). Fixing it properly
means extracting the shared renderer, which is a refactor rather than a one-liner and does
not belong in a PR this size. Filed as #404 and added to the demo doc's known limitations —
called out explicitly because it is the same shape as finding 1: a criterion verified on one
surface and not the other.

### State at merge

- **CI green** on every job across both Python versions.
- **All five acceptance criteria demonstrated** against a real stack, a real database and a
  real browser — `docs/demos/us-26.2-subscription-tiers.md`.
- **Every Critical/Major finding fixed**, from six review rounds across two model families.
- **Deferred, each with a rebuttal and an issue**: #402 (duplicate profile fetch),
  #403 (release bundling / local remaster — product decisions), #404 (clip-card badges).
- The demo itself found three ungated Pro routes that no reviewer or test caught; review
  found two over-gates the demo did not. Neither pass would have been sufficient alone.

Issue #402, opened 2026-08-06
P2.4 Deduplicate the subscription-tier profile fetch (US-26.2 follow-up)


frankbria:
## Problem

`useSubscriptionTier()` does its own uncached `fetch("/api/users/me")`, so every component that needs the tier issues its own request. Where two of them co-mount, the same profile is fetched twice for one render:

- `components/song/SongDetail.tsx:78` calls `useSubscriptionTier()` for the menu badges, and
- `components/song/SongDetail.tsx:79` calls `useSongActions(clip)`, which calls `useSubscriptionTier()` again internally (`hooks/use-song-actions.ts:60`).

Raised in review on PR #399 and deferred there: it is correctness-neutral ("not incorrect, just wasteful") and request efficiency is outside issue #330's acceptance criteria.

## Why it was not fixed in #399

The obvious local fix — pass `isFreeTier` into `useSongActions` instead of letting it fetch — pushes the problem onto every call site of that hook (`ClipCard`, `SongDetail`, and any future one) and makes the gating easier to get wrong by omission, which is the failure mode US-26.2 exists to close. The right fix is a shared tier source, which is a refactor with its own blast radius and did not belong in that PR.

## Suggested approach

Lift the lookup into a context/provider mounted once (the existing `ModelSelectionProvider` already fetches the same profile, but only on the Create page — worth deciding whether these become one thing), or dedupe in-flight requests inside the hook. If dedupe, share only the **in-flight** promise and clear it on settle: caching the resolved tier indefinitely would mean an upgrade does not take effect until reload, which the API deliberately avoids by reading the tier from the database rather than the token claims.

Whatever the mechanism, keep the current fail-closed default — anything that is not exactly `"pro"` is free, and a failed fetch must never unlock Pro features.

## Acceptance criteria

- [ ] Mounting `SongDetail` issues one `/api/users/me` request, not two
- [ ] A tier change is still reflected without a full page reload
- [ ] A failed or timed-out profile fetch still leaves the user on the free tier

PR #405, opened 2026-08-06
fix: US-26.2 ClipCard ⋯ menu badges every Pro item (#404)

### LINKED ISSUE #404: P2.7 ClipCard's ⋯ menu shows Pro badges on only one of its Pro items (US-26.2 follow-up)
## Problem

`ClipCard`'s flat "⋯" menu renders a Pro badge for exactly one of its Pro-only items —
**Open in Editor**, hardcoded inline. `Send to Mastering`, `Export to DAW`, and the whole
`Download` submenu (`DOWNLOAD_ITEMS`, a local list separate from `lib/song-actions.ts`'s
`SONG_DOWNLOAD_ITEMS`) render as plain items: no badge, no padlock, no `data-locked` — even
though all of them are `proOnly` in the shared registry.

`SongActionsMenu` (song detail) renders the badge generically off `action.proOnly`, so the
two surfaces disagree about which features look locked.

**Behaviour is correct** — `emitMenu` dispatches through `useSongActions`, which runs
`isSongActionLocked`, so a free-tier click still opens the upgrade prompt rather than
running the action. What is missing is the *advance* signal, which is the AC1 property
("free user sees Pro badges on restricted features").

**Pre-existing, not a regression.** Verified against `main`: `send-mastering` and
`export-daw` render badge-less there too. Found during PR #399 review (round 5) and
deliberately not fixed there, since AC1 was already satisfied by US-17.2 on the canonical
surface and #399 was large enough.

## Root cause

`ClipCard` hardcodes its own menu structure and its own `DOWNLOAD_ITEMS` rather than
rendering from `lib/song-actions.ts`. The file's own comment admits it: *"hardcodes its own
open-editor Pro lock / sample Beta badge — keep those in sync with the `pro`/`beta` flags
here by hand until a third consumer justifies a shared renderer."* That third consumer is
effectively here — the drift it predicted has happened.

## Suggested approach

Render the ⋯ menu's Pro badges from the shared registry the way `SongActionsMenu` does,
reusing `isSongActionLocked(action, { isFreeTier, nativeFormat: clip.format })` so the
native-format carve-out (a free musician downloading back the wav they uploaded) applies
consistently. Extracting the shared `ActionItem` renderer is the obvious way, and would
delete the hand-sync comment along with the duplication.

## Acceptance criteria

- [ ] Every `proOnly` item in `ClipCard`'s ⋯ menu shows the Pro badge, with the padlock only when locked for that user
- [ ] Download items respect the native-format carve-out (no padlock on a wav download of a native-wav clip)
- [ ] The badge state on a clip card matches song detail for the same clip and tier
- [ ] A test covers a Pro item other than Open in Editor

frankbria:
Closes #404.

## The problem

`ClipCard` hardcoded its own ⋯ menu items and its own local `DOWNLOAD_ITEMS`, so
exactly one of its Pro-only items — *Open in Editor* — rendered a Pro badge.
*Send to Mastering*, *Export to DAW* and the whole Download submenu are `proOnly`
in `lib/song-actions.ts` but rendered bare: no badge, no padlock, no `data-locked`.

Behaviour was already correct (`emitMenu` dispatches through `useSongActions`, which
runs `isSongActionLocked`, so a free-tier click opened the upgrade prompt). What was
missing is the *advance* signal AC1 asks for, and the two surfaces disagreed about
which features look locked — `SongActionsMenu` renders the badge generically off
`action.proOnly`.

Pre-existing, not a regression from #399.

## The change

Extracts `SongActionsMenu`'s local `ActionItem` into
`components/song/SongActionItem.tsx` and renders **both** of `ClipCard`'s menus from
the registry. The badge, the padlock and the native-format carve-out now come from one
place; the file's own "keep these in sync by hand until a third consumer justifies a
shared renderer" comment is gone, along with the duplication it predicted would drift.

The Sample `beta` flag moved into the registry too — it was the other half of that
hand-sync comment. **Side effect worth naming: song detail now shows the Beta mark on
*Sample from Song* as well.** That is the two surfaces agreeing, which is the point of
the issue, but it is a visible change to a screen the issue did not mention.

Icons stay hidden in the card's compact menus (Delete keeps its own), so the card's
appearance is otherwise unchanged. `onMenuAction` observers still receive the
`ClipMenuAction` id, so `remix-edit` keeps its identity separate from `remix`.

Net: -30 lines of component code.

## Acceptance criteria

| AC | Evidence |
|----|----------|
| Every `proOnly` item in the ⋯ menu shows the Pro badge, padlock only when locked | [demo, AC1](../blob/fix/us-26.2-clipcard-pro-badges/docs/demos/us404-clipcard-pro-badges.md) — *Open in Editor*, *Send to Mastering*, *Export to DAW* all 🔒 Pro; Pro tier sees badges without padlocks |
| Download items respect the native-format carve-out | WAV-native clip → WAV is "Pro" **without** a padlock; MP3-native clip → WAV is 🔒 Pro. Both screenshotted |
| Badge state on a clip card matches song detail for the same clip and tier | Screenshot pair, plus `marks the same actions as song detail for the same clip and tier` — opens both menus and compares badge + `data-locked` for every shared action |
| A test covers a Pro item other than Open in Editor | `badges Send to Mastering as Pro…`, `badges Export to DAW as Pro…`, `badges the Pro download formats…` |

Demo (screenshots + method): `docs/demos/us404-clipcard-pro-badges.md`.

## Verification

- `vitest run` — 206 files, 1637 tests, all passing
- `tsc --noEmit` clean, `eslint` clean, `prettier --check` clean on changed files
- `next build` succeeds
- Demoed in a real browser: `ClipCard` (free/Pro × WAV-native/MP3-native) and
  `SongActionsMenu` mounted on `/labs` against `npm run dev`, driven with
  `agent-browser`. The harness page was reverted; only the screenshots remain.

## Review

Cross-family pre-PR review: **`codex review --base main`** — *"preserves the existing
menu behaviors while aligning Pro/Beta badges across clip and song surfaces … did not
identify any actionable regressions."*

opencode/GLM was the intended primary reviewer but returned `UnknownError: Unexpected
server error` on two attempts (refs `err_27150172`, `err_905f2f35`), so codex is the
cross-family pass here rather than the fallback.

## Known limitations

- The card still takes `isFreeTier` as a prop while `useSongActions` fetches the tier
  itself — the duplicate-fetch issue is #402 and is deliberately untouched here.
- Song detail's Beta mark on *Sample from Song* is new, as described above.


frankbria:
## Cross-family review (post-PR)

**Reviewer:** `codex review --base main`, run against the full branch (both commits).

> The changes consistently move ClipCard and SongActionsMenu rendering to the shared
> action registry, and the targeted tests, lint, and typecheck pass. I did not find any
> discrete correctness issue introduced by this diff.

It independently ran `SongActionsMenu.test.tsx`, `ClipCard.test.tsx`,
`song-actions.test.ts` and `use-song-actions.test.tsx` (86 passing), plus
`npm run lint` and `npm run typecheck`. **No findings — nothing to triage.**

### Why codex and not opencode

opencode/GLM is the intended primary reviewer for this repo. It failed twice with
`UnknownError: Unexpected server error` (refs `err_27150172`, `err_905f2f35`) before
the diff reached the model, so it produced no review at all rather than a clean one.
codex is standing in as the cross-family pass, not as a fallback after a partial
result.

An earlier `codex review --base main` on the first commit (before the parity test and
demo landed) also reported no actionable regressions.

frankbria:
## Review triage

Both bot reviewers reported **no concrete defects** (GLM precision review, Claude review). One Minor suggestion was actionable and is now addressed in 5f33c82.

> `item()` does `findSongAction(...); if (!definition) return null`. […] if a new `ClipMenuAction` variant is ever added to the union without a matching entry in `SONG_ACTION_GROUPS`/`SONG_DOWNLOAD_ITEMS`, the menu item just silently vanishes […] a `console.warn` (or a dev-only invariant) would make a future registry/id drift loud instead of silent.

Agreed on the failure mode. Took a test rather than a `console.warn` — it fails in CI instead of in someone's console, and costs nothing at runtime:

```ts
const CLIP_MENU_ACTIONS: Record<ClipMenuAction, true> = { "remix-edit": true, /* …all 16… */ }

it.each(Object.keys(CLIP_MENU_ACTIONS) as ClipMenuAction[])(
  "resolves %s to a registry entry", …
)
```

The `Record<ClipMenuAction, true>` literal must name every variant to compile, which is what makes the per-id checks exhaustive — adding a variant to the union without listing it is a type error, and listing it without a registry entry fails the test.

**Verified the guard actually bites**, rather than assuming it: deleting the `mashup` key gives

```
lib/song-actions.test.ts(209,9): error TS2741: Property 'mashup' is missing in type
'{ … }' but required in type 'Record<ClipMenuAction, true>'.
```

Suite is now 206 files / 1653 tests, all passing.

No other findings to triage — the rest of both reviews was confirmation (dispatch contract preserved, carve-out intact, React keys unique, type changes non-breaking).

Issue #406, opened 2026-08-07
P2.3 Settle the VST3 SDK (Steinberg) licence before any public plugin build


frankbria:
## Context

Surfaced while resolving #369 (JUCE licensing). #369 settled the **JUCE** arm — JUCE 8
EULA, Starter tier — and that decision is recorded in `plugin/LICENSE.md`. It does not
cover the VST3 SDK, which is a **separate licence from Steinberg** and is not granted by
the JUCE licence at any tier.

Confirmed on the JUCE forum by the JUCE team: *"You will need a separate VST3 license
from Steinberg."*

`plugin/CMakeLists.txt` builds `FORMATS VST3 Standalone` (plus AU on macOS), so this
applies to the format the plugin is named for.

## Why it needs a decision

The Steinberg VST3 SDK is itself dual-licensed:

- **Steinberg VST3 Licensing Agreement** (proprietary) — requires registering as a
  VST3 developer and accepting its terms, including its trademark/logo rules.
- **GPLv3** — permitted, but distributing under it obliges us to offer the plugin's
  complete corresponding source under GPLv3. That reintroduces exactly the obligation
  #369 chose the commercial JUCE arm to avoid, so picking GPLv3 here would make the
  JUCE decision moot for the VST3 build.

Picking the proprietary arm is almost certainly what we want, but it is paperwork with
Steinberg that has not been done, and it is free — so the only cost of settling it is
the time.

## Not urgent, but it is a hard gate on distribution

Same shape as #369: binding only at **distribution** time. Local development and CI
builds are unaffected. AU and Standalone builds are unaffected — this is VST3-only.

## Done when

- [ ] The VST3 arm is chosen (Steinberg agreement vs GPLv3) and recorded in
      `plugin/LICENSE.md` alongside the JUCE decision
- [ ] If Steinberg: registered as a VST3 developer and the agreement accepted
- [ ] The "Undecided" row in the root `README.md` licensing table is filled in
- [ ] If GPLv3: the source-offer obligation is documented, and the conflict with the
      JUCE Starter decision in #369 is explicitly reconciled

## References

- Raised while implementing #369.
- JUCE forum, "Some questions about JUCE Starter": a separate Steinberg VST3 licence is
  required.

frankbria:
## Plan (implementing-issue-plans, 2026-09-17)

[...]

PR #407, opened 2026-08-07
docs: US-369 record the JUCE licensing arm — JUCE 8 EULA, Starter tier

### LINKED ISSUE #369: P2.1 Decide JUCE licensing (GPLv3 vs commercial) before distributing the VST3 plugin

frankbria:
Closes #369.

Records the licensing decision: **JUCE 8 EULA, Starter licence type** — free, no
source-offer obligation, valid while annual revenue is under **$20,000**. Not AGPLv3.
We are at $0, so we qualify today; upgrade to Indie before crossing the threshold.

## The issue's premise was outdated, in our favour

#369 was filed on this reasoning:

> `JUCE_DISPLAY_SPLASH_SCREEN=0` is already set in `plugin/CMakeLists.txt`. Disabling
> the splash screen is **only permitted under a paid licence** […] So the build as it
> stands already assumes the commercial arm without that having been decided.
>
> That last point is the concrete reason this can't sit indefinitely: the current
> configuration is inconsistent with the free arm.

That was true for JUCE 5–7. **JUCE 8 removed the splash-screen requirement from the
Starter tier.** Two independent confirmations:

- The [JUCE 8 EULA](https://juce.com/legal/juce-8-licence/) contains **no splash-screen
  clause at all** — the full agreement text was fetched and searched.
- JUCE's own release announcement: *"JUCE 8 features a much more generous Starter tier,
  with a narrower definition of income and the requirement to display a splash screen
  removed."*

So the build was **already consistent** with the chosen arm. `JUCE_DISPLAY_SPLASH_SCREEN`
keeps its value; what changed is the comment explaining why it is allowed, plus a warning
not to carry it across a JUCE 9 bump (JUCE 9 ships a different EULA).

The "Done when" item *"`JUCE_DISPLAY_SPLASH_SCREEN` is set consistently with that choice
— `0` only if a paid licence is actually held, otherwise removed"* is satisfied by the
first branch, but via the free tier rather than a paid one.

## What the tier actually costs and constrains

| | Starter | Indie | Pro |
| --- | --- | --- | --- |
| Annual revenue or funding limit | **Up to $20,000** | Up to $300,000 | No limit |
| Perpetual price per user | **Free** | $800 | $3,500 |
| Monthly subscription per user | N/A | $40 | $175 |

Three terms recorded because they bite later, not now:

- **Exceeding $20,000 is not a grace period.** EULA 1.2.1 requires either buying the
  appropriate tier **or immediately ceasing development and distribution**. Breach means
  back-fees for the whole period plus audit costs of no less than £1,000 (EULA 3.6).
- **How the limit is counted depends on who holds the licence.** For an individual, only
  framework-derived revenue counts. For a **company**, it is the entity's and all
  affiliates' total revenue **from all sources, whether connected to the framework or
  not, without offsets** (EULA 1.2.1). If this is ever held by a company with other
  income, that income counts toward the $20,000.
- **Licence types cannot be mixed** (EULA 1.13), so the eventual move to Indie is a clean
  switch, not a per-product choice.

## What the JUCE decision does not cover

The **VST3 SDK is licensed separately by Steinberg** — not granted by the JUCE licence at
any tier. `plugin/CMakeLists.txt` builds `FORMATS VST3 Standalone`, so this is live for
the format the plugin is named for. Its GPLv3 arm would reintroduce exactly the
source-offer obligation this decision avoids, so it needs its own answer.

Filed as **#406** and marked **Undecided** in the README table rather than left implicit.
It is the same shape as #369: binding only at distribution, blocks nothing locally.

## Changes

| File | Change |
| --- | --- |
| `plugin/LICENSE.md` | **New.** The decision, the tier table, obligations, the upgrade trigger, and the JUCE-9 caveat |
| `plugin/README.md` | Short Licensing section up top, pointing at it |
| `README.md` | Top-level licensing table — JUCE arm chosen, VST3 arm undecided |
| `plugin/CMakeLists.txt` | Comments only: why the splash setting is permitted, and that the `GIT_TAG` pins which EULA applies |

No behaviour change. `plugin/LICENSE.md` records the JUCE dependency's arm only — it does
**not** declare a licence for this project's own source, which the repo has never
published and which is not this issue's call to make.

## Verification

- `cmake -S . -B build` reconfigures cleanly (the libcurl warning is pre-existing)
- `JUCE_DISPLAY_SPLASH_SCREEN=0` still reaches the compiled target — grepped the
  generated build files — and no comment text leaked into the compile definitions,
  which was the one way a comment inside `set()` could have gone wrong

## Known limitations

- The $20,000 threshold has no automated tripwire; it is a documented human trigger.
- The individual-vs-company counting rule is recorded but not resolved, because it
  depends on whether this ends up held personally or by an entity — a fact this repo
  doesn't know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,3 +435,17 @@ keys follow `{user_id}/{workspace_id}/clips/{clip_id}.{format}`. See
 ## Environment
 
 Copy `.env.example` to `.env` and fill in the required values before running.
+
+## Licensing
+
+This project has not published a licence for its own source. What is recorded here is
+the arm chosen for dependencies that are dual-licensed, where the choice constrains how
+we build and distribute.
+
+| Component | Arm chosen | Obligation it creates |
+| --- | --- | --- |
+| **JUCE 8.0.9** (`plugin/`) | JUCE 8 EULA, **Starter** tier — free, not AGPLv3 | Valid only while annual revenue is under **$20,000**; upgrade to Indie before crossing it. No source-offer obligation. |
+| **VST3 SDK** (`plugin/`) | **Undecided** | A VST3 binary needs its own Steinberg agreement (or the SDK's GPLv3 arm). Tracked in issue #406 — settle it before any public plugin build. |
+
+Details, the exact upgrade trigger, and how the $20,000 is counted for a company versus
+an individual: **[`plugin/LICENSE.md`](plugin/LICENSE.md)**.
--- a/plugin/CMakeLists.txt
+++ b/plugin/CMakeLists.txt
@@ -16,6 +16,9 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 # and no clone happens at all.
 # ---------------------------------------------------------------------------
 include(FetchContent)
+# The tag fixes which licence governs: JUCE 8 EULA, Starter tier (plugin/LICENSE.md).
+# JUCE 9 ships a different EULA, so moving this tag is a licensing change, not just a
+# version bump.
 FetchContent_Declare(JUCE
     GIT_REPOSITORY https://github.com/juce-framework/JUCE.git
     GIT_TAG        8.0.9
@@ -77,6 +80,9 @@ set(ACEMUSIC_JUCE_DEFS
     JUCE_WEB_BROWSER=0
     JUCE_USE_CURL=${ACEMUSIC_JUCE_CURL}
     JUCE_VST3_CAN_REPLACE_VST2=0
+    # Permitted: JUCE 8 removed the splash-screen requirement from the Starter tier,
+    # which is the arm this repo uses. Under JUCE 5-7 this needed a paid licence, so
+    # do not carry this line to a JUCE 9 bump without re-reading plugin/LICENSE.md.
     JUCE_DISPLAY_SPLASH_SCREEN=0
     # Checked by the platform panel, so a curl-less Linux build explains itself
     # instead of failing an https:// request with a socket error.
--- a/plugin/LICENSE.md
+++ b/plugin/LICENSE.md
@@ -4,6 +4,10 @@ The plugin links the **JUCE Framework**, which is dual-licensed. This file recor
 which arm this repository uses, because the two arms impose opposite obligations and
 the build configuration has to match the one we picked.
 
+> **Not legal advice.** This is a record of a decision and the clauses it rests on. The
+> clause numbers are cited so the reasoning can be checked against the agreement itself
+> rather than taken on trust — read the linked EULA before relying on any of it.
+
 ## The decision
 
 **JUCE 8 End User Licence Agreement — Starter licence type. Not AGPLv3.**
@@ -44,7 +48,7 @@ $20,000.** Exceeding the limit is not a grace period: EULA 1.2.1 and the clause 
 **or immediately ceasing development and distribution**. Breach exposes us to back-fees
 for the entire period plus audit costs of no less than £1,000 (EULA 3.6).
 
-Two details worth knowing before that threshold is near:
+Two details worth knowing well before that threshold is reached:
 
 - **How the $20,000 is counted depends on who holds the licence.** For an *individual*,
   it is only revenue arising from their use of the framework. For a *company*, it is the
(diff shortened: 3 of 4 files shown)


frankbria:
## Cross-family review (pre/post-PR)

**Reviewer:** `codex review --base main`.

> The changes are documentation and CMake comments only, with no apparent behavior
> changes or actionable defects introduced relative to the base branch.

**No findings.** opencode/GLM was unavailable this session (`UnknownError: Unexpected
server error` on both attempts during #405 earlier today), so codex is the cross-family
pass here.

### Sources, so the legal claims are checkable rather than asserted

The two load-bearing factual claims were each verified against a primary source, not
recalled:

1. **The JUCE 8 EULA has no splash-screen clause.** The full agreement at
   <https://juce.com/legal/juce-8-licence/> was fetched and searched for "splash" — zero
   occurrences, in a text that extends through clause 11.4, so it was not a truncated
   fetch.
2. **JUCE 8 removed the requirement from Starter.** JUCE's own announcement post: *"JUCE 8
   features a much more generous Starter tier, with a narrower definition of income and
   the requirement to display a splash screen removed."*

Tier table, revenue-counting rule (1.2.1), no-mixing rule (1.13), and the breach terms
(3.6) are quoted from the same EULA text.

**This is not legal advice** — it is a record of a decision and the clauses it rests on.
The clause references are there so the reasoning can be checked against the agreement
rather than taken on trust.

frankbria:
## Review triage

Both reviewers reported **no defects**. GLM specifically checked the one place a real bug could hide — whether the `#` lines added inside `set(ACEMUSIC_JUCE_DEFS …)` leak into compile definitions — and confirmed they do not, matching the grep of the generated build files in the PR body.

Two non-blocking suggestions, both taken:

1. **"Not legal advice" caveat moved into `plugin/LICENSE.md`.** Correct call — it only lived in the PR description, and the file is what a future reader (or someone copying the pattern for another dependency) actually opens. Now a blockquote directly under the intro, pointing at the linked EULA as the authority.
2. **"before that threshold is near" → "well before that threshold is reached."**

Also noting the reviewer's catch that #369's title says "GPLv3" while JUCE's free arm is **AGPLv3** — that was deliberate, not an oversight, and the PR uses AGPLv3 throughout. The issue title stays as filed.

Issue #333, opened 2026-07-23
P1.5 US-26.5: Usage Dashboard


frankbria:
## User Story

**As a** musician, **I want** to see my credit usage history and breakdown by category, **so that** I can understand my usage patterns and plan accordingly.

## Description

A usage dashboard shows remaining credits, usage history over time, and breakdown by action category (generation, editing, mastering, video, etc.).

## Functional Requirements

- Usage dashboard accessible from account settings or sidebar
- Display: credits remaining (monthly + purchased), days until reset, current tier
- Usage chart: daily or weekly credit consumption over the past 30 days
- Category breakdown: pie/bar chart showing credits spent per action type
- Usage history table: date, action type, clip title, credits consumed
- Export usage data as CSV

## Acceptance Criteria

- [ ] Dashboard shows accurate credit balance and reset date
- [ ] Usage chart reflects actual credit consumption over time
- [ ] Category breakdown correctly attributes credits to action types
- [ ] Usage history lists all credit-consuming actions
- [ ] CSV export contains complete usage data

## Dependencies

**Blocked by:** US-26.4 (prior story in this stage). Part of Stage 26: Credits & Subscription System, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 26, US-26.5

Issue #332, opened 2026-07-23
P1.2 US-26.4: Credit Top-Up Purchase


frankbria:
## User Story

**As a** musician, **I want** to buy additional credits when I run out, **so that** I can keep creating without waiting for my monthly reset.

## Description

Both Free and Pro users can purchase credit top-up packs. Purchased credits do not expire and are used after the monthly allocation is exhausted.

## Functional Requirements

- Credit packs available: 50 credits, 100 credits, 250 credits (pricing TBD)
- Purchase flow via payment provider (one-time charge, not recurring)
- Purchased credits are additive and do not expire
- Monthly credits are consumed first; purchased credits are consumed after monthly allocation is depleted
- Purchase history visible in the usage dashboard
- Credit pack purchase available from the upgrade modal and the usage dashboard

## Acceptance Criteria

- [ ] Purchasing a credit pack increases the credit balance by the correct amount
- [ ] Monthly credits are consumed before purchased credits
- [ ] Purchased credits persist across monthly resets (do not expire)
- [ ] Purchase history is visible in the usage dashboard

## Dependencies

**Blocked by:** US-26.3 (prior story in this stage). Part of Stage 26: Credits & Subscription System, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 26, US-26.4

PR #420, opened 2026-08-07
feat: US-26.3 Stripe subscription billing (#331)

### LINKED ISSUE #331: P1.1 US-26.3: Payment Integration
## User Story

**As a** musician, **I want** to subscribe, cancel, and manage my payment method, **so that** I can upgrade to Pro and manage my billing.

## Description

Integrate with a payment provider (Stripe or similar) for subscription management. The musician can subscribe to Pro, update their payment method, cancel their subscription, and switch between tiers.

## Functional Requirements

- Subscribe to Pro: redirects to payment provider's checkout flow
- Payment methods: credit/debit card, with potential for additional methods
- Cancel subscription: takes effect at end of current billing period (no immediate feature loss)
- Upgrade from Free to Pro: immediate access to Pro features, prorated first month
- Downgrade from Pro to Free: takes effect at end of billing period
- Billing history accessible from account settings
- Webhook handler for payment events (successful charge, failed charge, subscription canceled)

## Acceptance Criteria

- [ ] Subscribing to Pro via checkout flow activates Pro features immediately
- [ ] Canceling retains Pro access until the end of the billing period
- [ ] Failed payment triggers a grace period with retry before downgrade
- [ ] Billing history shows all past charges with dates and amounts
- [ ] Webhook correctly processes payment events and updates user tier

## Dependencies

**Blocked by:** US-26.2 (prior story in this stage). Part of Stage 26: Credits & Subscription System, which is blocked by Stage 21.

## References

- User stories: `user-stories/04-layer-4-advanced-integrations.md` — Stage 26, US-26.3

frankbria:
Closes #331.

Stripe subscription billing: checkout, cancellation, dunning, billing history, and the
webhook that drives all of it. Also adds `/settings/billing`, which `UpgradeModal` has
linked to since US-26.2 — **the route did not exist, so every upgrade CTA in the app was
a 404.**

## The one design decision

**Stripe is the system of record; the user document is a read-model of it, written by
webhooks.**

Every tier check in the platform reads `user.subscription_tier`, and that read happens
on nearly every request — the sidebar polls balance and tier on each page. Asking Stripe
for the answer would put a network call in the hottest path in the app and would fail
the page whenever Stripe has a bad day, for information already known.

Two acceptance criteria then fall out rather than needing machinery:

- **AC2 — cancelling keeps Pro to the period end.** Stripe reports a pending
  cancellation as *still `active`* with `cancel_at_period_end` set, and sends
  `customer.subscription.deleted` when the period actually elapses. Nothing here has to
  notice an expiry date: no scheduler, no nightly sweep.
- **AC3 — grace period on a failed charge.** `past_due` maps to Pro while Stripe's retry
  schedule runs; only the terminal states (`unpaid`, `canceled`) downgrade.
  Re-implementing dunning here would be a second retry clock that could disagree with
  the one actually charging the card.

Unknown statuses **fail closed** to free, matching `tiers.normalise` — Stripe can add a
status without asking us, and the safe reading of an unfamiliar one is "not paid".

## Security

The webhook is unauthenticated by necessity (Stripe has no session) and mutates
subscription tiers, so:

- Verification uses the SDK's `construct_event` — constant-time comparison and a
  timestamp tolerance that defeats replay — rather than hand-rolled HMAC. This is the
  one place in the codebase not to be inventive.
- `stripe_webhook_secret` is part of the `stripe_enabled` check. Without it the endpoint
  cannot verify anything, and an unverifiable tier-mutation endpoint should be closed,
  not open.
- Idempotency is a **unique index** on `stripe_event_id`. Stripe's delivery contract is
  at-least-once, so a redelivery is expected traffic, not an error.
- The Stripe webhook is deliberately **not** proxied through Next.js — its signature
  covers the exact bytes sent, and a re-encoding hop would break verification while
  exposing an unauthenticated mutation endpoint on the frontend origin.

## Acceptance criteria

Demoed against a **live API** (uvicorn + real MongoDB) with real HMAC-signed webhooks
posted at the real endpoint. Full transcript: `docs/demos/us-26.3-payment-integration.md`.

| AC | Evidence |
|----|----------|
| 1 — checkout activates Pro immediately | `checkout.session.completed` → `tier` goes `free` → `pro`, `status: active`. Screenshot of the free-tier entry point |
| 2 — cancelling retains Pro to period end | `cancel_at_period_end` event → still `"tier": "pro"`, with `current_period_end` recorded. UI: *"Cancelled — Pro access continues until Aug 29, 2026"* |
| 3 — failed payment → grace period, then downgrade | `invoice.payment_failed` → `pro` / `past_due`; `unpaid` → `free`. Both transitions shown |
| 4 — billing history with dates and amounts | `$12.00 paid` with an invoice link, in the API and on the page |
| 5 — webhook processes events and updates tier | Redelivery → `duplicate`, one row not two. Forged signature → **HTTP 400**, tier unchanged. Unhandled type → 200 `ignored` |

## Thirteen defects found before merge

One by the demo, two by `codex review`, ten by the bot reviewers across seven review
rounds. Every one moved money, blocked payment, or displayed a wrong number. Each is
enumerated in a review-triage comment on this PR.

Three of them were **caused or left open by earlier fixes on this same PR** — guarding
double-billing created a lockout for churned musicians; the out-of-order watermark
covered subscription events but not invoices; the crash-safety reorder made every
mutation need to be idempotent against newer state, and the one handler that grants Pro
wasn't. In payment code each fix changes the state machine, so the next pass has to run
against the new one rather than the original diff.

The last one found — a one-off invoice moving subscription status — is the landmine
**#332 (credit top-ups) would have walked into**, since a top-up is exactly an invoice
with no subscription.

The demo-found defect is described below because it is the one the screenshots show.

## A defect the demo found

The first capture of the billing page showed `customer.subscription.updated` and
`checkout.session.completed` rows in **Billing history**, each with a dash for the
amount. They land in that collection because it doubles as the idempotency guard — but a
billing history is a list of charges, not a log of callbacks.

The history endpoint now filters to rows carrying an amount.
`test_lifecycle_events_are_not_listed_as_charges` covers it, and also asserts all three
events are still *recorded*, since dropping them would break idempotency. Failed
attempts stay visible — they carry an amount and a non-paid status, which is exactly
what someone opens the page to find out.

## Verification

- 1846 unit tests + **55** billing tests + 177 in the touched-area integration set, all passing
- `ruff`, `black`, `tsc --noEmit`, `eslint`, `prettier` clean
- 207 web test files / 1661 tests passing
- Demoed end-to-end as above, in a browser and against the live API

## Known limitations — stated, not papered over

- **No Stripe account exists in this environment.** Everything driven by webhooks — AC2,
  AC3, AC4, AC5 and the tier-flip half of AC1 — is fully verified, because the
  subscription lifecycle is an event stream and an event stream can be replayed. The
  **outbound** calls (creating a Checkout Session and a Portal Session) are implemented
  and covered against the SDK seam, but have never been run against Stripe's real API.
  Point `ACEMUSIC_API_STRIPE_*` at a test-mode account to close that gap; `stripe listen`
  will drive the webhook half too.
- **Proration on a plan change is the portal's job.** `POST /billing/checkout` now
  refuses an account that already has a live subscription (409) — subscription-mode
  Checkout creates a *new* subscription rather than amending the existing one, so
  allowing it would double-bill. The issue's "prorated first month" therefore applies to
  Stripe's own upgrade path, not to a second checkout.
- **A residual race on that guard, accepted knowingly.** Two concurrent `POST /checkout`
  calls can both pass the `AlreadySubscribed` check before either has a subscription on
  file. Raised in review; narrowed rather than closed. The Stripe idempotency key means
  both share one customer, and a duplicate subscription on a single customer is visible
  and refundable in Stripe's dashboard. Closing it properly needs an atomic claim marker
  with a release path for abandoned checkouts, whose failure mode — a stuck marker
  locking someone out of upgrading — is worse than the race it prevents.
- **Currency is whatever the Price is denominated in.** No multi-currency handling; the
  UI formats via `Intl.NumberFormat` from the currency Stripe reports.
- `/settings/billing` is a minimal page — plan, CTA, history. Richer usage reporting is
  #333 (P1.5, US-26.5 Usage Dashboard).


frankbria:
## Cross-family review (pre-PR) — two real bugs found and fixed

**Reviewer:** `codex review --base main`. It did **not** come back clean, and both findings were correct. Fixed in a16466f.

> **[P1] Prevent duplicate Pro subscriptions** — `services/billing.py:147-157`
> If an already-subscribed user reaches `POST /billing/checkout` directly or via stale UI, this Checkout call creates another subscription for the same Stripe customer; `subscription_data` only adds metadata and does not update/prorate an existing subscription. That can double-bill Pro users.

Correct, and the worst kind of bug this PR could ship — the only endpoint in the platform whose failure mode is charging someone money. Subscription-mode Checkout creates a *new* subscription against the customer; it does not modify the existing one. Now refused with **409** pointing at the billing portal, which is where a plan change belongs.

I also removed the comment that claimed `subscription_data` gave existing subscribers "a prorated switch rather than a second subscription" — that was the belief behind the bug, and leaving it would invite the same mistake again.

> **[P2] Apply side effects before marking webhooks duplicate** — `services/billing.py:283-290`
> If `_record_event` inserts the idempotency row and the process crashes or `user.save()` fails before the subscription fields are saved, Stripe's redelivery will hit the unique event id and return `duplicate`, permanently skipping the entitlement change.

Also correct: a musician who paid and stayed free, unrecoverable without manual intervention. **The state change now goes first.** That ordering is safe precisely because every state change here is idempotent — setting the tier to pro twice is setting it once — while the ledger insert is the thing that must happen at most once.

Codex suggested a transaction as the alternative. I went with the reorder instead: transactions need a replica set, which a single-node deployment does not have, and the reorder is the standard at-least-once-plus-idempotent-handler pattern rather than a workaround.

### Regression tests

- `test_an_existing_subscriber_cannot_start_a_second_one` — 409, not a second subscription
- `test_a_lapsed_subscriber_may_subscribe_again` — the guard must not strand someone whose subscription genuinely ended
- `test_a_redelivery_repairs_a_half_applied_event` — plants an event row with no corresponding state change (exactly the crash window), then delivers the event and asserts the tier *does* move and the ledger still holds one row

38 billing tests passing, up from 33.

frankbria:
## Review triage — five more fixes (ba2c2af)

Both bot reviewers found real defects. Every one of them costs someone money or shows them a wrong number, so all five are fixed rather than deferred.

### GLM — major: `customer.subscription.created` can downgrade a paid subscriber

Confirmed, and worse than the single event type suggests. **Stripe does not guarantee event ordering**, and each delivery carries its own event id — so the idempotency guard structurally cannot catch a late *older* snapshot. `subscription.created` (status `incomplete` while a 3DS card confirms) landing after Pro was granted was one instance; a redelivered `past_due` snapshot after recovery is another.

Fixed twice over:
- `customer.subscription.created` is no longer handled at all. `checkout.session.completed` grants access and `updated`/`deleted` carry every later transition, so it only ever arrived with a status too early to mean anything.
- A general guard: subscription snapshots older than the last one applied are dropped (`subscription_synced_at`, from the event's own `created` stamp). Removing the one messenger would have left the class of bug in place.

### GLM — minor: sync SDK calls block the event loop

Confirmed. All three outbound calls now go through `asyncio.to_thread`.

### claude-review — `ensure_customer` TOCTOU creates two Stripe customers

Confirmed, and the consequence is the bad one: the losing request checks out against an orphaned customer, `_user_for_customer` can't map the resulting webhook back, and the event is dropped as `unknown_customer` — **a musician who paid and stayed free**.

Fixed with a Stripe **idempotency key** rather than a lock or a unique index: both concurrent calls then return the *same* customer, which is precisely what the key exists for. A DB-side unique index would not have helped — the race is one user, two requests.

The reviewer was also right that the docstring claimed a guarantee the code didn't provide. It does now.

### claude-review — `amount_paid or amount_due` mishandles a $0 charge

Confirmed. `0 or x` is `x`, so a 100%-coupon month displayed the pre-discount figure as though it had been charged. Explicit `None` check.

### claude-review — `invoice.paid` and `invoice.payment_succeeded` double-record

Confirmed. Stripe fires both for the same successful subscription invoice with different event ids, so event-id idempotency doesn't dedupe them and the charge appeared twice. Only `invoice.paid` is handled now.

### On the TOCTOU in `create_checkout_session`

claude-review also flagged that the `AlreadySubscribed` guard has its own race (two tabs both passing the check before either has a subscription on file) and asked for it to be a conscious call rather than an oversight. **Conscious call: left as residual risk, narrowed.** The idempotency-key fix means both racers share one customer, and Stripe's own dashboard surfaces a duplicate subscription on one customer, which is recoverable. Closing it properly needs an atomic claim marker with a release path for abandoned checkouts — a state machine whose failure mode (a stuck marker locking someone out of upgrading) is worse than the race it prevents. Noted in the PR body as a known limitation.

### Tests

43 billing tests, up from 38. New regressions cover each fix: a stale snapshot not downgrading, a *newer* snapshot still applying (the guard must not freeze the read-model), `subscription.created` being ignored, a zero-amount invoice, and one row for a doubly-signalled payment.

The ordering guard also surfaced a latent trap — MongoDB returns naive datetimes, so comparing against a fresh aware one raises `TypeError`. `credits.apply_monthly_reset` already normalises the same way, so I matched it instead of adding a second convention.

frankbria:
## Review triage — two more (adc3801)

Both correct, both fixed. 46 billing tests, up from 43.

### GLM — major: a churned musician could never resubscribe

This is the sharpest finding on the PR. `invoice.payment_failed` set `subscription_status = "past_due"` unconditionally, and `past_due` is in `PAID_STATUSES` — so a late failure for a **long-dead** subscription flipped a free-tier account to a paid-looking status, and the `AlreadySubscribed` guard then refused them checkout **forever**, waiting on a subscription event they would never generate again. A customer trying to give us money, silently blocked, with no path out.

Note the shape: this bug was *created* by the fix for codex's double-billing finding two commits ago. Guarding a payment path introduced a new way to lock someone out of paying. Worth naming rather than quietly patching.

Fixed at both ends, because either alone leaves a hole:
- **The guard keys on the effective tier**, not the mirrored status. The tier is the entitlement that may not be paid for twice; the status is a mirror that can be wrong.
- **A payment failure for a non-subscriber no longer writes a paid-looking status at all.** The event is still recorded in billing history — a failed charge is a fact — but it does not resurrect `past_due` on a free account.

GLM suggested only the first; the second removes the bad state rather than teaching one caller to tolerate it.

### claude-review — the ordering watermark was not armed until the first subscription event

Correct, and it left unguarded exactly the window that matters. `_handle_checkout_completed` granted Pro without stamping `subscription_synced_at`, and `handle_event` did not even pass it `occurred_at` — so the first `customer.subscription.updated` was compared against `None`, and `if occurred_at and last_sync and ...` is false when `last_sync` is `None`. An `incomplete` snapshot generated *before* the checkout could still land after it and undo the grant, for someone who had just paid.

Checkout now stamps the watermark too.

### Regression tests

- `test_a_late_failure_does_not_lock_out_a_lapsed_account` — cancelled, then a late failure; status must not read as paid
- `test_a_lapsed_account_can_check_out_again` — the whole point: a churned musician can come back
- `test_a_snapshot_older_than_the_checkout_cannot_undo_it` — pre-checkout `incomplete` snapshot arriving after; tier stays pro

### Running tally

Nine defects found before merge on this PR: one by the demo, two by `codex review`, six by the bot reviewers. Every one moved money, blocked payment, or displayed a wrong number. The pattern in all of them is the same — the acceptance criteria describe the happy path, and payment integrations fail in the cases no criterion mentions: a second click, a redelivery, an out-of-order event, a crash between two writes, a coupon, a dead subscription. Recorded in `tasks/lessons.md` as a checklist to run *before* opening the next one, starting with #332.

frankbria:
## Review triage — invoice/subscription correlation (1f884f4)

Correct, and a sharp catch: it identifies the exact hole my own `subscription_synced_at` fix left open.

> `_handle_invoice` looks the user up by `invoice["customer"]` and then branches purely on the user's *current* `subscription_tier` / `subscription_status` — it never checks `invoice.get("subscription")` against `user.stripe_subscription_id`.

Confirmed. The watermark I added guards `customer.subscription.*` events against out-of-order delivery, but invoice events reused the churn-lockout *tier* check instead of an *identity* check — so the class of bug the watermark exists to close was still open on the other half of the event stream.

The reviewer is also right about which direction is worse. A stale `payment_failed` produces a false warning, which is annoying. A stale `invoice.paid` **clears a genuine `past_due`** — telling a musician their real payment problem is resolved when their card still needs attention, and they find out when the subscription dies.

Fixed as suggested: compare `invoice["subscription"]` to `user.stripe_subscription_id`, and when they differ, record the event (history and idempotency both still want it) without applying it.

One thing added beyond the suggestion: **an invoice with no `subscription` passes through untouched.** One-off charges — credit top-ups, US-26.4, which is the very next issue — carry no subscription id, and a correlation check written as `!=` alone would have swallowed every one of them. Covered by `test_a_one_off_charge_still_applies` so #332 doesn't rediscover it.

The reviewer's note that the gap was uncovered was also right, and worth repeating: the existing churn tests ran checkout → deleted → failure and stopped at the resubscribe. The bug lives one step past where the tests stopped. New tests continue through checkout → deleted → **checkout again** → stale event.

49 billing tests, up from 46.

### Tally

Ten defects found and fixed before merge: one by the demo, two by `codex review`, seven by the bot reviewers. This one is the most instructive of the set — it was created by an *incomplete* fix of mine, and only visible if you keep pulling the thread past where the previous fix stopped.

frankbria:
## Review triage — replayed checkout (this commit)

Correct, and I have to own two things about it.

**First: I dropped this finding.** GLM posted it after `adc3801`, and my next triage comment only addressed the `claude-review` finding from the same round. The reviewer spotted the gap in my own triage, not just in the code. Two reviewers posting in one round is not an excuse — I should be reconciling every finding in a round against the triage, and I wasn't.

**Second: it is a regression I introduced.** The crash-safety reorder in `a16466f` moved every state change *before* the idempotency row so a crash between them could not lose an entitlement. That is right, but it is only safe while each mutation is idempotent **with respect to newer state** — and `_handle_checkout_completed` stamped the watermark without ever reading it back. So the one handler that *grants* Pro was the one the ordering guard did not protect.

The concrete failure is exactly as described: subscribe (T1), cancel and let the period end (T2), then Stripe redelivers the T1 checkout event — or an operator clicks "Resend webhook" in the dashboard — and the musician is back on Pro for free, while the response says `duplicate` as though nothing happened.

Fixed by giving the property a name. `_is_stale(user, occurred_at)` is now shared by both handlers that stamp the watermark, rather than one of them being expected to remember to check it. That is the actual defect: the invariant existed only in a docstring.

**The reviewer's caveat checked out.** They flagged that `test_a_redelivery_repairs_a_half_applied_event` might break, since its redelivery has no prior watermark. It passes — `_is_stale` is correctly a no-op when `subscription_synced_at` is `None`, which is the crash-window case. Verified rather than assumed.

New test: `test_a_replayed_checkout_cannot_resurrect_a_cancelled_subscription`.

50 billing tests, up from 49. **Eleven defects fixed before merge.** Three of them — this one, the churn lockout, and the invoice correlation gap — were caused or left open by *earlier fixes on this same PR*. That is the pattern worth naming: in payment code, each fix changes the state machine, and the next review has to be run against the new one rather than the original.

frankbria:
## Review triage — round reconciled as a set

Reconciling the **whole round** this time rather than the last comment I read, since dropping a finding is exactly what went wrong two rounds ago.

**GLM (commit 6):** no new defects. Verified `_is_stale` is a faithful extraction and the replay trace holds.

**claude-review (commit 6):** the latest fix checks out; confirmed `test_a_redelivery_repairs_a_half_applied_event` still passes with the guard in place. **One previously-flagged item still open**, explicitly kept alive so it would not be lost between commits.

### The open item is now closed (483f7ba)

> `_handle_invoice` has no ordering guard of its own. An out-of-order `invoice.payment_failed` can still land *after* a newer `invoice.paid` for the same subscription […] and leave `subscription_status` reading `past_due` on an account that's actually current.

Correct. The correlation check from `1f884f4` catches an invoice for a *different* subscription; this is reordering **within** the current one, which it does not cover.

Both reviewers called it non-blocking, and they are right that it self-heals and never moves the tier. I fixed it anyway: it had survived three rounds as a tracked note, and a banner telling a musician to fix a card that is fine is still wrong. Carrying a known defect past merge to save twenty minutes is the wrong trade in payment code.

**Deliberately a separate watermark from `subscription_synced_at`.** Sharing one would let an invoice advance it and then silently drop a legitimate subscription event stamped a moment earlier — trading this cosmetic bug for an entitlement one, which is precisely the shape of the last three regressions on this PR. `test_the_invoice_watermark_does_not_block_subscription_events` asserts they do not interfere.

A successful invoice advances the watermark even when it changes nothing, so a later-arriving older failure cannot re-open a grace period that payment already settled.

### One test bug worth naming

My first version of that isolation test failed, and it was the *test* that was wrong: the checkout stamped the subscription watermark at "now", so a cancellation posted at `now - 10` was genuinely stale for reasons unrelated to the invoice watermark. Fixed by controlling the checkout's timestamp. Worth stating plainly because a red test that turns out to be a bad fixture is exactly the moment it is tempting to weaken the assertion instead.

53 billing tests, up from 50. **Twelve defects fixed before merge.** No open findings remain from any reviewer.

frankbria:
## Review triage — round reconciled (31f38dd)

**claude-review:** the invoice-ordering fix is correct; no new defects. One documentation nit — the PR description was stale ("Seven defects", "43 billing tests"). Fair, and it matters because the description is what a future reader actually reads. **Updated**: thirteen defects, 55 tests.

**GLM:** one minor finding, and it is the most consequential one on the PR for what comes next.

> A subscription-less (one-off) invoice still mutates `subscription_status` — a paid credit top-up clears a real `past_due`, a failed one opens one […] Gated on US-26.4 one-off charges.

Confirmed, and **this is the landmine #332 would have walked into**. A credit top-up *is* an invoice with no `subscription`, so it fell past the identity check straight into the status logic:

- a **paid** top-up cleared a genuine `past_due` — telling a musician the card that is actually failing is fine, so they find out when the subscription dies;
- a **failed** top-up opened `past_due` on a perfectly healthy subscription — telling them to fix a card that works.

Fixed: an invoice with no subscription says nothing about the subscription, so it is recorded (it is still a charge, and billing history is about charges) and applied to nothing else.

Worth noting the reviewer flagged this *before* the code that triggers it exists. I had added `test_a_one_off_charge_still_applies` two commits ago specifically so #332 wouldn't rediscover the correlation check swallowing top-ups — I got the direction right and the consequence wrong, asserting only that the charge was recorded, never that it left the subscription alone.

### Two existing tests were wrong

The fix broke `test_a_failed_charge_does_not_downgrade`, and the test was at fault: it posted `invoice.payment_failed` with **no `subscription`**, which a real Stripe subscription invoice always carries. Checking the neighbours, `test_a_recovered_payment_clears_the_grace_flag` had the same fixture and had quietly become a **false pass** — it asserted `status == "active"` while `past_due` was never being set in the first place.

Fixtures corrected so both exercise the real path. Weakening the assertion instead would have been the easy move and would have deleted the coverage.

55 billing tests, up from 53. **Thirteen defects fixed before merge.** No open findings from any reviewer.

Issue #403, opened 2026-08-06
P1.4 Decide whether release bundling and local remaster are Pro (US-26.2 follow-up)


frankbria:
## Question

US-26.2 (#330, PR #399) gated every endpoint whose Pro status was unambiguous. Two routes do
work that is *arguably* Pro but have no gated sibling doing the same thing, so whether they
belong to the free tier is a product decision rather than an inconsistency to close quietly.
They were found during the US-26.2 demo and deliberately left alone — see
`docs/demos/us-26.2-subscription-tiers.md`.

### 1. `POST /releases/{release_id}/prepare/{target}`

Validates a release for a distribution target and, if it passes, **builds a distribution
bundle** and returns `bundle_url` plus manual-submission instructions. `POST
/releases/{id}/submit/{target}` then marks it submitted.

The gated `distribution` capability currently covers the SoundCloud upload
(`routers/distribution.py`) — "we publish it for you". The question is whether "you publish
it yourself, we assemble the bundle" is also Pro. Issue #330 lists the free tier as having
"no distribution" without distinguishing the two.

- Gate it → the free tier cannot produce a release bundle at all.
- Leave it → a free account can assemble a distributable bundle and upload it manually,
  which is most of the value of the distribution feature.

### 2. `POST /clips/{clip_id}/remaster`

A local loudness-normalisation edit to a `target_lufs`, living in `routers/editing.py`
alongside crop and speed, and costing credits (`REMASTER_COST`). It is **not** the
Dolby/LANDR mastering service that `/mastering/jobs` runs, which is what the `mastering`
capability gates today.

The tension: the capability is described to users as "master your tracks to a professional
loudness target", which is close to what remaster does. Gating it would also convert a
currently credit-priced action into a tier-locked one, which has revenue implications —
hence not a call to make in a review pass.

## Decision needed

For each: free, Pro, or (for remaster) Pro-only above some quality/level threshold. Once
decided, the change itself is small — `require_capability` / `require_tier_capability` plus
a test in `tests/test_tier_enforcement_api.py`, following the pattern established by #399.

## Acceptance criteria

- [ ] A decision recorded for release prepare/submit, and the endpoint matches it
- [ ] A decision recorded for `/clips/{id}/remaster`, and the endpoint matches it
- [ ] If either becomes Pro: a free-tier refusal test and matching UI copy in `web/lib/tiers.ts`

PR #421, opened 2026-08-07
feat: US-26.4 credit top-up packs (#332)

### LINKED ISSUE #332: P1.2 US-26.4: Credit Top-Up Purchase

frankbria:
Closes #332.

Credit top-up packs — **50 / $5, 100 / $9, 250 / $20** (10¢, 9¢, 8¢ per credit). Priced
deliberately dearer per credit than the Pro plan so a top-up never undercuts subscribing.

## Two buckets, not a bigger number

This is the whole design, and it is forced rather than preferred.

`apply_monthly_reset` grants `max(0, allocation - credits_balance)` — it tops the balance
*up to* the tier allocation rather than adding to it. So with a single field:

> buy 100 credits, spend 60, and your next monthly 50 **never arrives** — the balance is
> still above 50, so the reset grants nothing.

AC2 ("monthly consumed first") and AC3 ("purchased persist across resets") cannot both
hold in one number. Purchases now live in `purchased_credits`, which the reset never
touches — so **AC3 needed no change to the reset at all.**

## Keeping the deduction atomic

`deduct_credits` was one `find_one_and_update` filtered on `credits_balance >= cost`.
Spanning two buckets while staying **one** operation is an aggregation-pipeline update
filtered on `$expr: monthly + purchased >= cost`, computing both decrements from the
input document in a single `$set`:

```
take    = min(monthly, cost)
monthly = monthly - take
bought  = bought  - (cost - take)
```

Read-then-write would let two concurrent generations each see enough credit and overdraw
— the race the original single update existed to prevent. Ten concurrent requests against
a balance covering five: exactly five succeed.

## A top-up rides the same webhook as a subscription checkout

Only metadata distinguishes them, so `_handle_checkout_completed` branches on the pack id
**first**. Without that, a paid $9 credit pack grants Pro — the most expensive possible
way to misread a webhook. Test + negative control.

The write order is **inverted** relative to the subscription handlers, deliberately. A
tier flip is idempotent, so state goes first and a crash cannot lose an entitlement. A
credit grant is not — crediting twice is a real gift — so the ledger row decides. That
alone would lose credits on a crash, so the row carries `credits_granted_at` and a
redelivery can tell "done" from "started and abandoned".

## Every balance surface

Seven routers read `credits_balance` directly and would each have under-reported what a
musician paid for. All now use `credits.spendable()`, with a test that greps the router
package and fails if a **new** endpoint reads the raw field — the failure mode a
per-endpoint test cannot catch.

## Acceptance criteria

Demoed against a live API with real HMAC-signed webhooks —
`docs/demos/us-26.4-credit-topup.md`.

| | monthly | purchased | spendable |
|---|---|---|---|
| start (free tier) | 50 | 0 | 50 |
| buy the 100-pack | 50 | **100** | 150 |
| spend 30 | 20 | 100 | 120 |
| spend 40 | **0** | **80** | 80 |
| anniversary reset | **50** | **80** | 130 |

The last row is the argument for the design: a single balance computes
`max(0, 50 − 80) = 0`, so the musician silently loses a month **as a consequence of
having paid us**.

| AC | Evidence |
|----|----------|
| 1 — a pack credits the account | `topped_up`, +100 purchased, tier untouched |
| 2 — monthly consumed first | spend 40 drains the last 20 monthly before touching 20 purchased |
| 3 — purchases survive resets | full 50 arrives, 80 purchased intact |
| 4 — purchase visible in history | `$9.00` one-off row, `subscription_status` untouched |

Also live: a redelivered purchase returns `duplicate` and credits nothing.

## Three bugs found in pre-PR review

- **The purchase was broken end-to-end.** The proxy forwarded only method and auth, so
  `{pack_id}` never reached the backend. Nothing caught it: component tests stub `fetch`
  in the browser, API tests call FastAPI directly, and **the demo screenshotted the pack
  picker without clicking one** — I demoed rendering and called it a purchase flow. Proxy
  tests added for that hop.
- **The crash window lost paid credits** (see `credits_granted_at` above).
- **Refunds went to a bucket that cannot hold them.** Worse than reported: *every* refund
  into `credits_balance` is absorbed, because the reset tops up to the allocation. Refund
  5 into 10 against an allocation of 50 → next reset grants 35 instead of 40. Refunds now
  go to `purchased_credits`.

That last change altered five refund tests. They asserted `credits_balance` directly; the
shared helper now asks `spendable()` — the question a musician would ask — rather than
any assertion being weakened.

## Verification

- 158 Python tests across billing/credits/refunds/reset, 208 web files / 1670 tests
- `ruff`, `black`, `tsc --noEmit`, `eslint`, `prettier` clean
- Demoed end-to-end against a live API and in a browser

## Known limitations

- **Refunds are bucket-generous, not bucket-accurate.** A charge that came entirely from
  the monthly allowance is refunded as non-expiring credit. Being exact would mean
  recording the split on every deduction across ~10 call sites; approximating in the
  musician's favour is the safe direction, but it is an approximation.
- **The outbound Stripe calls remain unverified against the real API** — no account
  exists here. Everything webhook-driven is fully exercised.
- No pack picker in `UpgradeModal` yet; `/settings/billing` is the single entry point.


frankbria:
## CI failure fixed — and the reason is worth recording

CI failed on `test_mastering_tasks`, `test_clips_full_song_api` and `test_voice_models_api`. All three assert `credits_balance` after a refund, and all three broke because I changed refunds to land in the purchased bucket. I had verified against the credit and billing suites — **not the full one** — so I checked the files I happened to be thinking about rather than the ones that depend on the behaviour I changed.

## Reassessed rather than edited fourteen files

The finding was narrower than my fix. Before this story there was one bucket, so "a refund is absorbed by the next reset" was already true — **pre-existing, not a regression**. The genuine regression is only that a charge which consumed *purchased* credits refunds as *monthly*, turning non-expiring credit into expiring.

Sending every refund to `purchased_credits` fixes that, but redefines what `credits_balance` means for every caller, and over-corrects in the other direction: a charge funded entirely from the monthly allowance would come back as non-expiring credit, quietly inflating what the platform owes.

So refunds go back to the monthly bucket, the limitation is documented at the call site, and the real fix is filed as **#422 (P2.5)**. That fix records the bucket split on the deduction so a refund is the exact inverse of its charge — which means consolidating ledger-writing that ~10 routers currently each do themselves. That is a piece of work, not something to bolt onto this PR.

One part of the change was right independently and stays: the refund's `balance_after` reports **both** buckets. A musician's "balance after" is what they can spend, and with purchased credits in play a single field understates it.

142 tests pass locally across the three suites that failed, plus the credit, bucket and refund suites.

## What I'd do differently

After changing a shared behaviour, enumerate what depends on it — `grep -rl credits_balance tests/` returns fourteen files in a second — instead of running the suites already in mind. It is the top entry in my own lessons file, which is the annoying part.

frankbria:
## Review triage — four findings (6bdbbda)

All four correct, all fixed. Reconciling the round as a set.

### The double-grant race — the one that gives money away

> The "ledger row decides" invariant is sequential-only: recording the event and marking `credits_granted_at` are two separate writes across two collections, so a redelivery that lands while the first delivery is still inside `grant_purchased_credits` reads `credits_granted_at` as `None` and grants again.

Confirmed, and it was introduced by my own repair path two commits ago. I closed the crash window (recorded-but-never-granted) and opened a concurrency one in the same change.

The claim is now a single conditional `find_one_and_update` on
`{stripe_event_id, credits_granted_at: None}`. Exactly one caller can move the row from
unclaimed to claimed, whatever the concurrency, and only that one grants. A failed grant
releases the claim so a redelivery retries.

`test_concurrent_deliveries_of_one_purchase_credit_once` fires five simultaneous
deliveries of the same event; the account is credited once.

### The service-layer balance sites — my own guard was blind

> The router-level grep test doesn't catch these because they're in `services/`, not `routers/`.

This is the useful one. I wrote that grep test specifically to catch "a new call site
reads the raw field", and then scoped it to `routers/` — so it was blind to the same bug
one layer down, in three places. Fixing the three sites matters less than **widening the
guard**, which now covers `services/` too. `credits.py` is excluded, since it owns the
field and its reads are the implementation rather than a report.

### `apply_monthly_reset`'s ledger row

Fixed — it reported the monthly bucket alone while `deduct_credits` and `refund_credits`
both report the combined total, so a `monthly_reset` row under-reported against its
neighbours.

### `reverse_unrecorded_charge`

Taking the reviewer's own suggested minimum: a note at the call site pointing at #422,
rather than a second half-fix of the same problem in a different place. It is narrower
than the refund path — it only fires when job *creation* fails, before any ledger row
exists — and the real fix is the same work.

### One test changed

The repair-path case asserted a `"repaired"` status string that no longer exists: with an
atomic claim, a repair is **indistinguishable** from a first delivery. That is the point,
so the test asserts the credits landed rather than the label.

### A correction to my own earlier verification claim

I previously reported a 163-test pass for the service suites. **That number was void** — I
ran `git stash` and `git checkout` while the suite was executing, so it ran against a
working tree changing underneath it. Discarded and re-run on a stable tree: **246 passed**
across billing, billing-service, credit-buckets, refunds, credits-api, reset, mastering
and voice-models. Recorded in my own notes as: never stash or checkout while a run is in
flight — a green result whose inputs changed mid-run is worse than no result, because it
looks like evidence.

frankbria:
Stale docstring fixed.

> `_balance()`'s docstring still says "Refunds now land in the *purchased* bucket," which was true for `6be4889c` but was reverted in `84137362b`.

Correct. The revert commit updated the behaviour and the code comment in `credits.py`, but the test helper's docstring was corrected **on the wrong branch** during the parallel #403 work and never reached this one. No behavioural impact — the helper uses `spendable()`, which is bucket-agnostic, which is exactly why nothing failed and nothing caught it. A comment that survives its own revert is the kind of thing only a reader notices.

It now says refunds go to the monthly bucket and points at #422 for the real fix. 25 refund tests pass.

Issue #423, opened 2026-08-07
P2.6 Test users default to free tier, so every gating story breaks dozens of unrelated tests


frankbria:
## Problem

API tests create users through per-file `_make_user` helpers that leave
`subscription_tier` at its default of `"free"`. Any story that gates an existing endpoint
therefore breaks every test that exercised it — none of which are about tiers.

Measured, not estimated. Gating two endpoints in #403 broke **39 tests across two files**:

| File | Broken | Why |
| --- | --- | --- |
| `tests/test_distribution_prep_api.py` | 24 | free users hitting `POST /releases/{id}/prepare\|submit/{target}` |
| `tests/test_clips_edit_api.py` | 15 | free users hitting `POST /clips/{id}/remaster` |

US-26.2 (#330) paid the same tax on a larger scale. Every future gating story pays it
again, and the cost is not just the edit — it is that a wall of unrelated red obscures
whether the *gate itself* is right.

## Why this is worth fixing rather than absorbing

The repair is easy to get subtly wrong in a way that destroys coverage. The tempting fix
is to default the helper to Pro, which makes everything green immediately — and silently
removes the proof that the *ungated* operations are still ungated. In
`test_clips_edit_api.py` the operations are parametrized (`crop`, `speed`, `remaster`),
so a file-level Pro default would have meant crop and speed were no longer tested as a
free musician experiences them. That was avoided there with a `_tier_for(operation)`
helper, but it is a trap laid for whoever does this next.

## Suggested approach

Shared fixtures in `tests/conftest.py`, so the tier a test needs is stated rather than
inherited:

```python
@pytest.fixture
async def free_user(...): ...

@pytest.fixture
async def pro_user(...): ...
```

Two properties worth preserving:

- **A test that does not say which tier it wants should get `free`**, not Pro. Free is
  the default a real signup gets, and defaulting to Pro is what hides gating regressions.
- **Parametrized-by-capability tests need the tier to follow the parameter**, not the
  file — see `_tier_for` in `test_clips_edit_api.py` for the shape.

Worth doing as a single mechanical pass across the API test files rather than
opportunistically, so the convention is uniform and the next gating story can rely on it.

## Acceptance criteria

- [ ] Shared `free_user` / `pro_user` fixtures exist and are used by the API test files
- [ ] Per-file `_make_user` helpers that only differ in tier handling are removed
- [ ] Tests for ungated operations still run as a free user (no blanket Pro default)
[...]

PR #424, opened 2026-08-07
feat: Pro-gate release bundling and local remaster (#403)

### LINKED ISSUE #403: P1.4 Decide whether release bundling and local remaster are Pro (US-26.2 follow-up)

frankbria:
Closes #403.

**Decision: both are Pro.**

The issue asked for a product call on two endpoints that do arguably-Pro work but had no
gated sibling doing the same thing. Both are now gated, and the reasoning is recorded in
`services/tiers.py` beside each capability description so the next reader does not have
to find this PR.

## `POST /releases/{id}/prepare|submit/{target}` → `distribution`

"You publish it yourself, we assemble the bundle" is most of the value of distribution.
Gating only the SoundCloud upload left the free tier with nearly the whole feature while
the gate guarded the convenience. The free tier is specified as having no distribution;
this is distribution.

**`submit` is gated with `prepare`**, which the issue did not ask for. An account that
cannot build a bundle has nothing to submit, and an open `submit` would let it mark a
release *distributed* anyway.

## `POST /clips/{id}/remaster` → `mastering`

Loudness mastering by another route. The capability is described to musicians as "master
your tracks to a professional loudness target" — which is exactly what remaster does,
just locally instead of via Dolby/LANDR. Leaving it open made the gate on
`/mastering/jobs` a formality.

**The gate runs before the credit charge.** The issue flagged that gating remaster
converts a credit-priced action into a tier-locked one; charging someone and *then*
refusing them would be the worst version of that. Rather than assert this in isolation,
remaster joins the shared `_pro_actions` table in `test_tier_enforcement_api`, so it also
inherits the existing "costs nothing when refused" and "Pro is not refused" cases — the
two properties most likely to be missed when bolting a gate onto an endpoint that already
takes money.

## The UI had to move too

`remaster` was not `proOnly` in the shared registry, so a free musician would have clicked
a normal-looking menu item and got a 403 — US-26.2's "a menu is not a gate" problem in
reverse. Marking it in `lib/song-actions.ts` lights it up on song detail and the clip card
at once, via the shared renderer from #404.

## Acceptance criteria

- [x] A decision recorded for release prepare/submit, and the endpoints match it
- [x] A decision recorded for `/clips/{id}/remaster`, and the endpoint matches it
- [x] Free-tier refusal tests for both, and the UI marks remaster Pro

## Verification

134 tests across `test_clips_edit_api`, `test_distribution_prep_api` and
`test_tier_enforcement_api`; 1661 web tests; ruff/black/tsc/eslint clean.

## The test cost, and one trap avoided

The gates broke **39 pre-existing tests** across two files — all using free-tier users
against endpoints that are now Pro, none about tiers.

The tempting fix is to default the test helpers to Pro. That makes everything green and
**silently destroys the coverage proving the ungated operations are still ungated**:
`test_clips_edit_api` is parametrized over `crop`/`speed`/`remaster`, so a file-level Pro
default would have meant crop and speed were no longer tested as a free musician
experiences them. If either were accidentally gated later, the test would sail through.

So the tier follows the *operation* (`_tier_for`), not the file. `test_distribution_prep_api`
does default to Pro, because every endpoint in it is now gated and the gate itself is
covered once in `test_tier_enforcement_api` rather than 24 times there.

The recurring version of this cost is filed as **#423 (P2.6)** — shared `free_user` /
`pro_user` fixtures, with free as the default precisely because Pro-by-default is what
hides gating regressions.

## Known limitations

- **Revenue implications are unmodelled.** The issue notes gating remaster converts a
  credit-priced action into a tier-locked one. Free-tier musicians who were spending
  credits on remaster now cannot; whether that is net-positive is a business question
  this PR does not answer, only implements.
- No demo document: the behaviour is two 403s and a badge, all covered by the refusal
  tests and the existing US-26.2 demo of the 403 body shape.

### DIFF
--- a/src/acemusic/api/routers/editing.py
+++ b/src/acemusic/api/routers/editing.py
@@ -22,9 +22,15 @@ from pydantic import BaseModel, ConfigDict, Field, field_validator, model_valida
 from acemusic.audio import calculate_speed_multiplier
 from acemusic.utils import parse_time_string, snap_to_beat
 
-from ..auth.dependencies import CurrentUser, get_current_user, require_existing_user
+from ..auth.dependencies import (
+    CurrentUser,
+    get_current_user,
+    require_existing_user,
+    require_tier_capability,
+)
 from ..models import Clip
 from ..services import clips as clip_service, credits as credits_service, editing as editing_service
+from ..services.tiers import Capability
 
 logger = logging.getLogger(__name__)
 
@@ -221,7 +227,18 @@ async def remaster_clip(
     request: RemasterRequest,
     current: CurrentUser = Depends(require_existing_user),
 ) -> EditJobResponse:
-    """Enqueue a remaster of ``clip_id`` to ``target_lufs``; the original is preserved."""
+    """Enqueue a remaster of ``clip_id`` to ``target_lufs``; the original is preserved.
+
+    Pro-only (#403). This is loudness mastering by another route — the ``mastering``
+    capability is described to musicians as "master your tracks to a professional loudness
+    target", which is exactly what this does, just locally instead of via Dolby/LANDR.
+    Leaving it open would have made the gate on ``/mastering/jobs`` a formality.
+
+    The gate runs **before** the credit charge in ``_enqueue``: refusing someone after
+    taking their credits would be worse than not offering the feature.
+    """
+    await require_tier_capability(current.user_id, Capability.MASTERING)
+
     clip = await clip_service.get_owned_clip(clip_id, current.user_id)
     _require_wav(clip)
 
--- a/src/acemusic/api/routers/releases.py
+++ b/src/acemusic/api/routers/releases.py
@@ -20,7 +20,13 @@ from datetime import datetime
 from fastapi import APIRouter, Depends, HTTPException, status
 from pydantic import BaseModel, ConfigDict, field_validator, model_validator
 
-from ..auth.dependencies import CurrentUser, get_current_user, get_settings, require_existing_user
+from ..auth.dependencies import (
+    CurrentUser,
+    get_current_user,
+    get_settings,
+    require_existing_user,
+    require_tier_capability,
+)
 from ..models import DistributionStatus, Release, ReleaseStatus, VisibilityState
 from ..models.distribution import GUIDED_CHANNELS, SOUNDCLOUD_CHANNEL
 from ..services import (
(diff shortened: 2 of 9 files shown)


frankbria:
Gap closed.

> `submit` gating lacks an independent positive-path test. […] `test_a_pro_account_is_not_refused` only exercises `prepare`, not `submit`.

Correct, and worth closing rather than deferring: without it, a gate that refused
*everyone* would have passed. That is the failure mode positive controls exist for, and
it is the one I would least want to discover from a support ticket.

`prepare` and `remaster` inherit that check from the shared `_pro_actions` table.
`submit` takes a release id rather than a clip id, so it does not fit that table's shape
and needed its own — `test_a_pro_account_may_submit`.

5 tests in `TestReleaseBundlingIsPro` now: refusal for both endpoints, a positive control
for both, and one confirming reads stay open so the gate is not too wide.

PR #430, opened 2026-08-10
feat(#333): usage dashboard (US-26.5)

### LINKED ISSUE #333: P1.5 US-26.5: Usage Dashboard

frankbria:
Closes #333.

## What

`GET /api/v1/credits/usage?days=30` and a `/settings/usage` page that renders it: both
credit buckets, the next reset date, a daily consumption chart, a per-category breakdown,
the history table, and CSV export.

One endpoint backs the whole page. The three views are derived in Python from a single
read of the ledger window — a month of movements for one musician is tens of rows, and
fetching them apart would let the charts and the table disagree mid-render.

Two shaping rules carry the meaning:

- **Refunds cancel the charge they reverse.** `mastering_refund` nets against `mastering`.
  The blanket failure path labels refunds by *job* type, so `generate_refund` is mapped
  alongside the `song`/`sound` charges it actually reverses — otherwise a refunded song
  drifts into "other" and never cancels anything.
- **Grants are history, not consumption.** `monthly_reset` stays in the table (the balance
  jumped and the table has to say why) but out of the charts.

The endpoint applies the lazy monthly reset first, exactly as `/credits/balance` does —
otherwise the dashboard shows a stale balance to precisely the people who opened it to
check their credits.

## Choices worth flagging

- **No chart library.** `web/` has none today; two bar charts do not justify adding
  recharts. Hand-drawn SVG, the same idiom as the mastering EQ display.
- **No CSV endpoint.** The browser builds it from the payload already on screen, so a
  download cannot export something the musician never saw.
- **`credits.buckets()`** is new in `services/credits.py`. The dashboard needs the split
  that `spendable()` deliberately hides, and the bucket guardrail test is right that no
  other module should read `credits_balance` — so the split is served from the module
  that owns the field.

## Verification

Demo with outcome evidence: `docs/demos/us-26.5-usage-dashboard.md` — live stack, seeded
Pro account, every criterion cross-checked against the endpoint rather than read off the
screen. Highlights: 30 `<rect>` for a 30-day window with heights exactly proportional to
the API's daily credits; mastering showing the net 4 after an 8-credit charge and a
4-credit refund; `sum(daily) == sum(categories) == 42.5`; and the CSV captured off the
real download (19 lines = header + all 18 rows).

Tests: 23 service + 6 API (Python), 17 lib/proxy + 9 component (web). Full web suite
(1690) green, plus `test_credit_buckets`, `test_credit_refunds`, `test_credit_reset`,
`test_credits_api`, `test_users_api`. Web typecheck, lint and build run locally.

Reviewed pre-PR by `codex review` (opencode/GLM errored out); its one finding — a negative
category total emitting an invalid CSS width, which browsers drop so the bar renders *full*
width — is fixed with a regression test.

## Known limitations

- The ledger records a combined `balance_after`, not which bucket paid (#421/#422), so
  monthly-vs-purchased is reported for the **remaining** balance only.
- Free actions (crop, speed, export) and Stripe top-ups write no ledger row, so "complete
  usage data" means complete *credit-moving* data.
- History is capped at 1000 rows per window; the charts still total the whole window, so a
  heavy month reports honest numbers with a shortened table.
- Clip titles are best effort — a generation job only names its clip once it has completed.


frankbria:
## Third-party review (post-PR)

Reviewer: `codex review --base main` (cross-family). `opencode` (GLM) was tried first and
errored out with a server-side `UnknownError`, so codex is the pass of record.

**Verdict on the current HEAD — no findings.**

> The changes add a usage dashboard endpoint, service shaping logic, proxy route, UI, and
> tests without any clear correctness, security, or reliability regressions found in the
> reviewed diff. Type checking and targeted frontend tests pass; Python lint passes for the
> touched backend/test files.

### The pre-PR pass did find one thing, and it is fixed

On the first run codex flagged `UsageSettings.tsx` (category bars):

> **[P2] Handle negative net category totals before drawing bars** — When a refund falls
> inside the selected window but the original charge is just outside it (for example
> viewing 7 days after an 8-day-old job is refunded), `row.credits` can be negative. This
> width calculation then emits a negative CSS percentage when there is any positive
> category in the same window, which browsers treat as invalid/auto and can render the
> refund category as a full-width bar instead of a net credit-back entry.

Confirmed and fixed in `dfd6fb9`: the width is clamped at zero while the negative total
stays readable in the label, covered by `draws no bar for a category that nets credit back`.

Separately, the repo's own bucket guardrail (`test_credit_buckets.py::test_nothing_reports_a_balance_from_the_monthly_bucket_alone`)
caught `usage.py` reading `credits_balance` directly. The dashboard genuinely needs the
split that `spendable()` hides, so it now comes from a new `credits.buckets()` — the
monthly bucket keeps exactly one reader, which is what that guard exists to protect.

frankbria:
## Review triage — all findings addressed in `7d06e59`

### Fixed

**CSV sign convention** (GLM inline `web/lib/usage.ts:98`, and its summary comment).
Correct and worth fixing: the export wrote a refund as `-4` while the table showed `+4`,
under a docstring claiming the export "matched what the table shows". They answer
different questions — the table says *what happened to my balance*, the export says *what
did this cost* — and the export's convention is the one to keep, because it is the column
a spreadsheet sums. Kept the sign, renamed the column to **`credits_used`**, and fixed the
docstring so the two no longer disagree silently. Test updated to state the intent.

**CSV formula injection** (claude-review #2). A clip title opening with `= + - @` is
executed by Excel, and titles are user-controlled text landing in a file the musician
double-clicks. Those cells are now prefixed with `'`. Numbers deliberately skip the guard —
a refund's leading minus is arithmetic, and escaping it would break the column the fix
above exists to keep summable. Covered by `defuses a clip title a spreadsheet would run as
a formula`.

**`GRANT_CATEGORIES`** (claude-review #3). Fair — a one-element `frozenset` behind a single
`in` check. Now a plain string comparison.

### Considered, deliberately unchanged

**Net-negative days render as no bar** (claude-review #1). This is the conscious call:
showing it honestly needs a second visual direction (an inverted or differently-coloured
bar) and a legend to explain it, for a case that only arises when a refund lands on a
different day than its charge. The category card already surfaces the netting with the
signed number, so the information is on the page — just not in the daily chart. Flagged as
a known limitation rather than half-drawn.

**Empty `?days=` falls back to 30** (claude-review #4). The reviewer is right that this is
incidental rather than intended, but the behaviour it produces (a working dashboard instead
of a 422) is the better one, and reaching it requires hand-editing the query string.
Leaving it.

Demo doc re-captured from the live app so the CSV sample matches the new header.

Issue #427, opened 2026-08-09
P2.9 Video jobs can be re-claimed mid-render by a sibling worker (stale-requeue window is shorter than the handler)


frankbria:
## Context

Spun off from the review of #401 (free-tier video watermark). Not introduced by that PR — `src/acemusic/api/tasks/video.py:41-48` has documented the hazard since US-22.1 — but #401 widens it, so it is worth recording rather than leaving in a comment.

## The problem

`JobProcessor` treats a job as stale after `poll_timeout + 300s` (900s by default) and re-queues it. The video handler's worst-case wall time is `submit + POLL_TIMEOUT_S (600s) + download`, and since #401 also `+ the watermarking pass` (bounded by `_TIMEOUT_S = 1800` in `src/acemusic/video_watermark.py`). That comfortably exceeds 900s.

The sweep only runs at process **start**, so a single-process deployment is safe — a restart had already killed the previous worker. A multi-process deployment starting a sibling mid-render is not.

## Failure scenario

1. Worker A is inside the watermark step for a free-tier job.
2. Worker B starts; its stale sweep flips the job back to `queued`; B claims it.
3. Both submit to the provider — the render is billed twice.
4. Both insert a `Video` document (distinct `_id`s, same `storage_path`; `models/video.py` indexes only `clip_id + user_id`, so nothing rejects the duplicate).
5. B's upload overwrites A's at `{user}/{ws}/videos/{clip}/{job}.mp4`. If the owner's tier changed between A's and B's tier read, the surviving object may not match what A's `Video` document implies.

## Possible fixes

- Per-job-type `stale_after` (video needs a much larger one than the default).
- A heartbeat on the claimed job, so "stale" means "no progress", not "started long ago".
- A unique index on `Job.id` in the `videos` collection, so the duplicate insert fails loudly.

## Acceptance criteria

- [ ] A job actively being processed cannot be re-claimed by a sibling worker
- [ ] Two workers cannot both produce a `Video` document for the same job
- [ ] A test covers the requeue path for `video` jobs

## Dependencies

None; independent of #401, which is only affected in that it lengthens the window.

Issue #426, opened 2026-08-08
P2.7 Run web/ typecheck, lint, test and build in CI — a frontend-breaking change passed every check today


frankbria:
## Problem

CI does not build, typecheck, lint, or test `web/`. The `ci` job is Python-only, and the sole Node coverage is the `web-audit` job, which runs exactly two things:

- `npm audit --audit-level=high`
- `npm ls --package-lock-only --all` (lockfile consistency)

Neither executes a single line of frontend code. A change can break the entire web test suite and still show an all-green PR.

## This is not hypothetical — it happened today

PR #417 (`react` 19.2.7 → 19.2.8) bumped `react` and `@types/react` but left `react-dom` at 19.2.7. `web/package.json` pins both exactly, and React requires them to match:

```
Error: Incompatible React versions: The "react" and "react-dom" packages must have the exact same version. Instead got:
  - react:      19.2.8
  - react-dom:  19.2.7
```

**All 206 test files failed to load. Zero tests ran.**

That PR was green on every check this repo has: `ci (3.11)`, `ci (3.12)`, `web npm audit`, GitGuardian. It was caught only by running `npm test` locally during supply-chain triage.

The sharpest detail: **`tsc --noEmit` and `next build` both passed.** Only the vitest suite surfaces this failure mode. So the gap is not merely "no build in CI" — a build alone would not have caught it either. The test run is the thing that matters.

## Why this keeps recurring

Every frontend dependency bump currently depends on a human remembering to run four commands locally. Dependabot opens these continuously (four `web/` PRs in the 2026-08-07 batch alone). The failure mode is silent, and the reviewer's signal — green checks — actively points the wrong way.

## Proposed fix

Extend the existing `web-audit` job in `.github/workflows/ci.yml` (rename to something like `web`), reusing the Node setup already there:

```yaml
      - name: Install web dependencies
        working-directory: web
        run: npm ci

      - name: Typecheck
        working-directory: web
        run: npm run typecheck

      - name: Lint
        working-directory: web
        run: npm run lint

      - name: Test
        working-directory: web
        run: npm test

      - name: Build
        working-directory: web
        run: npm run build
```

Scripts already exist in `web/package.json` — `typecheck` (`tsc --noEmit`), `lint` (`eslint`), `test` (`vitest run`), `build` (`next build`). Local runtimes: full suite ~38s, plus `npm ci` and build. This should add roughly 2–4 minutes.

[...]

Issue #401, opened 2026-08-05
P1.3 Watermark free-tier 720p video output (US-26.2 follow-up)


frankbria:
## Context

US-26.2 specifies the free tier as **"720p watermarked video"**. PR #399 enforced the
resolution half — free accounts get 720p and are refused 1080p/4K — but the **watermark
was not implemented**, and neither #330 nor #400 addressed it.

Recording it rather than letting it disappear between two closed issues.

## Why it was not done in #330

The resolution gate is an authorization check: one call to `require_tier_capability` before
the charge. Watermarking is a *rendering* change — it belongs in the video pipeline
(`tasks/video.py` / the provider request), needs an asset and a placement decision, and has
no obvious right answer without knowing what the mark should say or look like.

## Open questions for whoever picks this up

- What does the mark look like, and where does it sit? (Corner bug, lower-third, full
  diagonal?)
- Is it applied by the provider or composited afterwards?
- Does an account that upgrades get a re-render of videos made while free, or do those keep
  their mark? The second is simpler and defensible; it should be a decision, not an accident.

## Acceptance criteria

- [ ] A 720p render requested by a free account carries the watermark
- [ ] A render requested by a Pro account carries none, at any resolution
- [ ] The behaviour of pre-existing videos after an upgrade is decided and documented
- [ ] Watermarking failure does not silently produce an unmarked video — it fails the job

## Dependencies

Depends on #330 (the tier table and the resolution gate).

frankbria:
Mark will be a corner bug consisting of the app icon and the app title "Cadenza" in the appropriate font. It will be applied afterwards as a separate function overlay on top of the video. Paid users cannot go back and re-generate videos without watermarks. Only videos generated during the period of payment are non-watermarked.

Issue #425, opened 2026-08-08
P2.8 Pin the remaining GitHub Actions by SHA, not mutable tags


frankbria:
## Problem

This repo pins GitHub Actions two different ways, and the inconsistency leaves credential-handling steps on mutable references.

**Already SHA-pinned** (correct — immutable digest plus a version comment):

| Action | Pin |
|---|---|
| `actions/checkout` | `@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1` |
| `actions/setup-node` | `@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0` |
| `anthropics/claude-code-action` | `@e0cf66d1d257526b5d07f141838c338921cb8455 # v1.0.182` |

**Still on mutable tags:**

| Action | Pin | Where | Notes |
|---|---|---|---|
| `docker/login-action` | `@v4.5.1` | `.github/workflows/docker-publish.yml:52` | **Runs with `DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN`** |
| `actions/setup-python` | `@v7.0.0` | `.github/workflows/ci.yml:31` | |
| `astral-sh/setup-uv` | `@v9.0.0` | `.github/workflows/ci.yml:36` | |

A git tag is mutable: it can be repointed to a different commit after the version has been reviewed and adopted. A digest cannot. The `docker/login-action` case is the one that matters most — that step receives Docker Hub push credentials, so a repointed tag there would run attacker-chosen code with registry write access in scope.

## Why this will not self-correct

Dependabot preserves whatever pinning style the file already uses. It updated all three of these actions in the current batch (#411, #408, #416) and kept them on tags each time. Without a manual change, every future bump re-pins to a tag indefinitely.

## Proposed fix

Convert the three remaining tag pins to `@<sha> # <version>` form, matching the existing convention. Dependabot understands this format and will keep both the digest and the trailing version comment updated on future bumps.

Resolve each digest from the upstream release tag, e.g.:

```bash
gh api repos/docker/login-action/git/ref/tags/v4.5.1  --jq .object.sha
gh api repos/actions/setup-python/git/ref/tags/v7.0.0 --jq .object.sha
gh api repos/astral-sh/setup-uv/git/ref/tags/v9.0.0   --jq .object.sha
```

Note that some actions publish annotated tags — if `.object.type` is `tag` rather than `commit`, dereference it to get the commit SHA rather than pinning the tag object.

[...]

PR #431, opened 2026-08-10
feat(#429): containerize the platform and ship it via CD

### LINKED ISSUE #429: P2.0 Containerize the platform deploy and ship it via CD (API + web + Mongo)
## Are we already on Docker?

Only for the GPU model, not the platform. `docker/Dockerfile` builds the **ACE-Step-1.5 worker image** (US-11.3, #124) and `docker-publish.yml` pushes it to Docker Hub for RunPod to pull. That is the inference engine.

Everything we actually ship has no container at all:

| Component | How it runs today |
| --- | --- |
| FastAPI platform API (`acemusic.api.main:app`) | `uv run uvicorn …` by hand on a shared VPS |
| Job processor (renders, mastering, video, stems…) | In-process inside the API's lifespan (`job_processor_enabled`) |
| SoundCloud poller | Same lifespan |
| Next.js web app | `npm run build && npm run start` by hand |
| MongoDB | Whatever is already on the box |

There is no compose file, no Procfile, no systemd unit, no deploy workflow, and no IaC. Provisioning is tribal knowledge, and the only record of what the host needs is prose scattered across `README.md` and `AGENTS.md`.

### And the one publish workflow we do have has never published anything

`DOCKERHUB_USERNAME` / `DOCKERHUB_TOKEN` are not set on this repo (secrets are only `CLAUDE_CODE_OAUTH_TOKEN` and `ZHIPU_API_KEY`). `docker-publish.yml` is written to warn-and-skip when they are absent, so every run reports **success** having skipped checkout, build and push:

```
Warn when Docker Hub credentials are missing: success
Run actions/checkout@v7: skipped
Log in to Docker Hub:    skipped
Build and push:          skipped
```

The skip-instead-of-fail behaviour was a deliberate choice (a red main on every Dependabot bump is worse), but the effect is a green check that proves nothing. Whatever registry this work lands on has to be actually wired up, and the "nothing configured" path should be visibly inert rather than passing.

## Why now

**A missing system package is now a product bug, not an inconvenience.** #401 made `ffmpeg` a hard runtime requirement for free-tier video watermarking — without it those jobs fail. It is also required for `flac`/`mp3` conversion, batch transcode, and Studio mixdowns. Nothing in the repo guarantees it is installed on the API host; CI installs it, production is on trust. That is exactly the class of drift an image removes.

Other things the current setup leaves to memory:

- The shared VPS means port conflicts are checked by hand before every bind.
- CORS origins, OAuth redirect URIs and `ACEMUSIC_API_*` settings are configured per-host with no manifest to diff against.
- `voice_training_root` defaults to a **relative** path (`./acemusic-voice-training`), so behaviour depends on the working directory the process happened to be started from.
- Local storage (`ACEMUSIC_STORAGE_BACKEND=local`) writes to a host path with no declared volume, so "where do the clips live" is per-deployment folklore.
- Rollback is "git pull the old commit and hope the dependencies still resolve".

## Scope — part 1: the images

1. **API image** — Python 3.11/3.12, `uv sync`, ffmpeg baked in, non-root user, healthcheck hitting `/api/v1/health`, `ACEMUSIC_API_*` config via environment only.
2. **Web image** — Next.js. Needs `output: "standalone"` in `web/next.config.ts` (not set today) to avoid shipping `node_modules`.
3. **Compose file** — API + web + MongoDB + named volumes, so `docker compose up` is the whole local stack and the same file documents production topology.
4. **Volumes and state** — a declared volume for local storage and for `voice_training_root`; make the latter absolute inside the container.
5. **Documentation** — replace the manual `uvicorn`/`npm start` instructions with the container path, and delete the host-requirement prose the image now encodes.

## Scope — part 2: continuous deployment

Merge to `main` should end with the new version running, without anyone SSH-ing anywhere.

6. **Build and publish on green main.** Tag every image with the commit SHA (deployable, traceable) plus a moving `latest`. Must run *after* CI passes, not in parallel with it — `main` is protected with `ci (3.11)` and `ci (3.12)` required and strict up-to-date enforcement, so a merge is already gated; the publish should key off that rather than re-deciding.
7. **Registry.** Recommend **GHCR** (`ghcr.io/frankbria/…`) over Docker Hub: authentication uses the built-in `GITHUB_TOKEN`, so there are no new credentials to set — which is precisely the failure mode that has kept `docker-publish.yml` inert. Docker Hub remains correct for the ACE-Step image, since RunPod pulls that one.
8. **Rollout to the VPS.** SSH to the host, `docker compose pull && docker compose up -d` against SHA-pinned tags. Needs a small set of secrets (host, user, SSH key) scoped to a GitHub **Environment** rather than repo-wide.
9. **Health gate and automatic rollback.** After `up -d`, poll `/api/v1/health` until healthy or a deadline; on failure, re-pin the previous SHA and bring the stack back up. A deploy that leaves the site down and reports green is worse than no CD.
10. **Idempotent and re-runnable.** Re-running the same deploy for the same SHA must be a no-op, not a second mutation.
11. **Concurrency.** One deploy at a time (`concurrency: production`), so two quick merges cannot interleave and land a mixed-version stack.

### Notes for whoever picks this up

- **No database migration step is needed.** Beanie creates indexes during `init_db` at API startup, so schema work happens as a side effect of the new container booting. That does mean index changes land at rollout time and are not separately gated — worth knowing before assuming a migration hook exists to hang things on.
- **Secrets to create:** VPS host / user / SSH private key, plus the production `ACEMUSIC_API_*` values. None of these exist today.
- **Zero-downtime is not free on a single VPS with compose.** Expect a short gap on restart unless we do blue/green or put a proxy in front. Fine to accept initially — but it should be an explicit choice, not a surprise.

## Decision needed: one container or two?

The job processor runs **in-process** inside the API today. There is already a `job_processor_enabled` flag, so splitting API and worker into separate containers is nearly free — but it is a real fork:

- **Single container (API + in-process worker).** Matches today's behaviour exactly. Simplest. Scaling the API scales the workers with it.
- **Split API and worker.** Better operationally, and lets the API scale independently. **But it makes #427 live** — the stale-requeue window is shorter than a video render's worst-case wall time, and multi-process workers is precisely the deployment that turns that latent race into double provider billing and racing `Video` documents. #427 should be fixed first, or the worker pinned to one replica.

Recommendation: **single container to start**, with the worker split deferred until #427 is closed.

## Decision needed: does a deploy wait for a human?

- **Fully automatic on green `main`.** Fastest, and CI is already a hard gate with two required checks.
- **Environment approval gate.** One click before production. Same workflow, one protection rule — cheap to add later, so this is reversible either way.

Recommendation: **fully automatic**, with the health gate and auto-rollback carrying the safety rather than a human staring at a diff. Revisit if a bad deploy ever gets through.

## Non-goals

Kubernetes, autoscaling, multi-region, and a separate staging environment. Single production target on the existing VPS.

## Acceptance criteria

**Images and local stack**

- [ ] `docker compose up` brings up API + web + MongoDB, and the web app can complete a real generation against the API
- [ ] The API image contains ffmpeg, verified by a free-tier video render succeeding inside the container (#401's watermark path)
- [ ] No component reads a relative filesystem path that depends on the working directory it was started from
- [ ] Local-storage clips and voice-training data survive a container restart via declared volumes
- [ ] All configuration is supplied by environment variables — no file edits needed to deploy to a different host

**Delivery**

- [ ] Merging to `main` publishes SHA-tagged images and deploys them, with no manual step
- [ ] The deployed API serves the new commit — verified by an endpoint that reports the running version, not by the workflow reporting success
- [ ] A deliberately broken deploy rolls back automatically and the workflow reports **failure**
- [ ] Re-running the same deploy is a no-op
- [ ] Two deploys cannot run concurrently
- [ ] Platform images do not collide with the `frankbria/ace-step` tags, and the ACE-Step publish keeps working (or is fixed) rather than silently skipping
- [ ] README/AGENTS.md describe the container deploy and the CD pipeline; the manual `uvicorn`/`npm start` instructions are removed or clearly marked development-only
- [ ] The single-vs-split worker decision and the approval-gate decision are recorded in the repo, not just in this issue

## Dependencies

- **#401** — established ffmpeg as a hard host requirement; the main driver for this work
- **#427** — must be resolved before splitting the worker into its own scalable container
- Related: **#426** (web/ is not in CI, so the web image would be built and deployed from code no pipeline typechecks — worth closing first, since CD makes that gap reach production automatically)

frankbria:
Closes #429.

## What

An API image, a web image, and one `compose.yaml` that is both the local stack and the
record of production topology — plus a CD pipeline that publishes SHA-tagged images on
green `main` and rolls them out with a health gate and automatic rollback.

Both decision forks in the issue are resolved as it recommends, and recorded in
`docs/deployment.md` rather than only here:

- **Single container** (API + in-process worker). Splitting the worker makes #427's
  stale-requeue race live; deferred until that closes.
- **Fully automatic deploys.** CI is already a hard gate with two required checks, and the
  health gate plus rollback carry the safety. An environment approval rule is one setting
  away if that proves wrong.

## Choices worth flagging

- **GHCR, not Docker Hub**, for the platform images. `GITHUB_TOKEN` authenticates it, so
  there are no new credentials — which is exactly the failure mode that has kept
  `docker-publish.yml` inert since it was written. Docker Hub stays correct for the
  ACE-Step image that RunPod pulls; the names do not collide.
- **The rollout logic is a script, not workflow YAML.** `scripts/deploy.sh` owns the health
  gate, the rollback, idempotence and a host-side lock, so it can be tested — and it is,
  with `docker` and `curl` stubbed on PATH so the suite needs no daemon. A rollback that
  has never run is not a rollback.
- **The gate is not "does something answer 200".** A crash-looping new image leaves the
  previous container answering happily. It waits for the API to report *the SHA just
  deployed*, which is why `GET /api/v1/health` gains `build_sha` — before it the only
  version reported was the static `0.1.0` nobody bumps.
- **Absolute data paths by environment, not by changing defaults.** `voice_training_root`
  is deliberately relative so the API and ACE-Step share a filesystem, and the value is
  sent verbatim to ACE-Step; changing the default would change what ACE-Step receives.
- **Unconfigured is now *skipped*, not green.** Both publish paths are gated at job level
  on a repository variable. `docker-publish.yml` previously reported success having skipped
  checkout, build and push.
- **`ci.yml` gains a `build images` job.** Otherwise the first place `docker build` runs is
  CD, on main, after the merge that broke it. It also runs `npm run typecheck` and
  `next build` inside the web image — the first time either runs in automation, though not
  a substitute for #426.

## Two bugs found by running the stack

Both fixed here, and both are the per-host tribal knowledge this issue is about:

1. Compose passed no compute-target URL, so every generation 503'd on the availability
   probe with the stack otherwise perfectly healthy.
2. Setting only `ACEMUSIC_API_LOCAL_URL` is not enough. The router's probe reads that
   (`ApiSettings`, `ACEMUSIC_API_` prefix); the job processor's client reads
   `ACEMUSIC_BASE_URL` (the CLI config, no `_API_`). With one set and not the other, every
   generation is **accepted and then fails**.

## Verification

`docs/demos/us-429-containerize-and-cd.md` — real daemon, real registry, real rollout:

- `docker compose up` → API + web + Mongo all healthy; a generation submitted through the
  containerised API **completes**, writing two clips to the mounted volume (ACE-Step itself
  is stubbed with the repo's own `scripts/demo_stub_acestep.py` — it is a separate CUDA
  image on a GPU host by design)
- ffmpeg verified by running #401's actual watermark path on a real 720p render inside the
  container, not by checking the binary exists
- Same container, three working directories, identical resolved paths
- Clips and voice-training data survive `--force-recreate`
- `deploy.sh goodsha` → live and reporting `build_sha=goodsha`; re-run → no-op, container
  not recreated; `deploy.sh badsha` (API exits on start) → health gate times out, rolls
  back to `goodsha`, **exit 1**; a second concurrent deploy → exit 3
- Port 3000 was already taken on this machine — resolved with `WEB_PORT` in `.env`, no file
  edit, which is the point

Tests: 1875 passed / 13 skipped (Python), 1692 (web), `black`, `ruff`, `actionlint`, web
typecheck and lint all clean.

## Known limitations

- **The SSH hop to a production VPS is not verified.** No host, user or key exists and I
  cannot create them, so the rollout is demonstrated by running `deploy.sh` exactly as the
  workflow runs it, against a registry the stack really pulls from. What is unproven is the
  transport, not the logic. Same for a live GHCR publish (it needs no new secrets, so it
  will run for real on the first merge) and the `PRODUCTION_HEALTH_URL` confirmation step.
- **The deploy job stays skipped until configured.** Set `DEPLOY_ENABLED=true` plus
  `DEPLOY_HOST` / `DEPLOY_USER` / `DEPLOY_SSH_KEY` on the `production` environment; full
  list in `docs/deployment.md`. Same for `DOCKERHUB_PUBLISH_ENABLED` to re-enable the
  ACE-Step publish.
- **A restart is a short outage.** Single VPS, single compose stack. Blue/green or a proxy
  would remove it; accepted deliberately rather than discovered later.
- **Index changes land at rollout.** Beanie creates indexes at startup, so there is no
  separate migration step to gate on.
- **The rollback restores images, not data.**
- The API image is ~1.1 GB, driven by the core scientific dependencies (numpy/scipy/
  librosa); the `audio-ml` extra is deliberately not installed.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,57 @@
+# Build context for the API image (#429).
+#
+# The first entries are the point of this file: `.env` and `.env.local` live at the repo
+# root and are gitignored, so nothing before this stopped a root-context build baking real
+# credentials into an image layer. .gitignore protects git; only this protects Docker.
+.env
+.env.*
+!.env.example
+
+# Secrets and local state
+.beads/
+*.pem
+*.key
+
+# Version control and CI
+.git/
+.github/
+.gitignore
+
+# Python build/test artifacts
+.venv/
+__pycache__/
+*.py[cod]
+.pytest_cache/
+.mypy_cache/
+.ruff_cache/
+.coverage
+coverage.xml
+htmlcov/
+dist/
+build/
+*.egg-info/
+
+# Node — the API image never needs these, and web/ builds from its own context
+node_modules/
+web/node_modules/
+web/.next/
+
+# Runtime data that must come from a volume, never from a layer. acemusic-voice-training/
+# alone is ~47 model directories of nothing the image should carry.
+storage/
+acemusic-voice-training/
+
+# The C++ plugin and its vendored JUCE build tree
+plugin/build/
+plugin/Builds/
+
+# Docs, demos and stray media — pure build-context weight
+docs/
+tasks/
+user-stories/
+*.md
+!README.md
+*.mp3
+*.wav
+*.png
+*.jpg
--- a/.env.docker.example
+++ b/.env.docker.example
@@ -1,0 +1,59 @@
+# Environment for the container stack (#429). Copy to `.env` beside compose.yaml.
+#
+#   cp .env.docker.example .env
+#   sed -i "s|^ACEMUSIC_API_JWT_SECRET_KEY=.*|ACEMUSIC_API_JWT_SECRET_KEY=$(openssl rand -hex 32)|" .env
+#   docker compose up
+#
+# This is the whole contract between the app and a host. Anything a deployment needs that
+# is not here belongs in the images, not in someone's shell history.
+
+# --- Required -----------------------------------------------------------------------
+# Tokens are signed with this. Compose refuses to start without it rather than falling
+# back to a shared default that would be identical on every deployment.
+ACEMUSIC_API_JWT_SECRET_KEY=
+
+# --- Local defaults, override per host ----------------------------------------------
+ACEMUSIC_API_MONGODB_DB_NAME=acemusic
+ACEMUSIC_API_CORS_ALLOW_ORIGINS=http://localhost:3000
+API_BASE_URL=http://api:8000
+API_PORT=8000
+WEB_PORT=3000
+# Bound to loopback on purpose — an exposed Mongo on a shared VPS is the whole database.
+MONGO_PORT=127.0.0.1:27017
+
+# --- Deploy-time (set by the CD workflow; leave blank locally) -----------------------
+# Registry images and the commit tag. Unset means compose builds from source instead.
+# API_IMAGE=ghcr.io/frankbria/auto-music-studio-api
+# WEB_IMAGE=ghcr.io/frankbria/auto-music-studio-web
+# IMAGE_TAG=<commit sha>
+
+# --- Optional integrations: unset means the feature is off --------------------------
+ACEMUSIC_API_GOOGLE_CLIENT_ID=
+ACEMUSIC_API_GOOGLE_CLIENT_SECRET=
+ACEMUSIC_API_GOOGLE_REDIRECT_URI=
+ACEMUSIC_API_DISCORD_CLIENT_ID=
+ACEMUSIC_API_DISCORD_CLIENT_SECRET=
+ACEMUSIC_API_DISCORD_REDIRECT_URI=
+ACEMUSIC_API_STRIPE_SECRET_KEY=
+ACEMUSIC_API_STRIPE_WEBHOOK_SECRET=
+ACEMUSIC_API_STRIPE_PRICE_ID_PRO=
+ACEMUSIC_API_OPENAI_API_KEY=
+ACEMUSIC_API_VIDEO_API_URL=
+ACEMUSIC_API_VIDEO_API_KEY=
+ACEMUSIC_API_RUNPOD_API_KEY=
+ACEMUSIC_API_RUNPOD_ENDPOINT_ID=
+
+# Where generation runs. The ACE-Step inference server is NOT part of this stack — it is a
+# separate CUDA image on a GPU host — so point this at it, or every generation 503s on the
+# availability probe with the rest of the stack perfectly healthy. `remote_only` with the
+# RunPod settings above is the other supported shape.
+ACEMUSIC_API_LOCAL_URL=http://localhost:8001
+ACEMUSIC_API_COMPUTE_PREFERENCE=local_first
+# Two config systems reach the same server: the availability probe reads the variable
+# above (ACEMUSIC_API_ prefix), the job processor's client reads this one (no _API_).
+# Leave blank to inherit ACEMUSIC_API_LOCAL_URL; set it only if they must differ.
+ACEMUSIC_BASE_URL=
+# Optional bearer for the ACE-Step server itself.
+ACEMUSIC_API_KEY=
+ACEMUSIC_STORAGE_BACKEND=local
+ACEMUSIC_API_SOUNDCLOUD_POLLER_ENABLED=false
--- a/.env.docker.example
+++ b/.env.docker.example
@@ -47,7 +47,10 @@ ACEMUSIC_API_RUNPOD_ENDPOINT_ID=
 # separate CUDA image on a GPU host — so point this at it, or every generation 503s on the
 # availability probe with the rest of the stack perfectly healthy. `remote_only` with the
 # RunPod settings above is the other supported shape.
-ACEMUSIC_API_LOCAL_URL=http://localhost:8001
+# `localhost` here would be the API container itself, not your machine — the compose file
+# defaults this to host.docker.internal (resolved via an extra_hosts entry) so an
+# ACE-Step server on the host is reachable. Point it at the real host in production.
+ACEMUSIC_API_LOCAL_URL=http://host.docker.internal:8001
 ACEMUSIC_API_COMPUTE_PREFERENCE=local_first
 # Two config systems reach the same server: the availability probe reads the variable
 # above (ACEMUSIC_API_ prefix), the job processor's client reads this one (no _API_).
--- a/.env.docker.example
+++ b/.env.docker.example
@@ -1,8 +1,11 @@
-# Environment for the container stack (#429). Copy to `.env` beside compose.yaml.
+# Environment for the container stack (#429). Copy to `.env.docker` beside compose.yaml.
 #
-#   cp .env.docker.example .env
-#   sed -i "s|^ACEMUSIC_API_JWT_SECRET_KEY=.*|ACEMUSIC_API_JWT_SECRET_KEY=$(openssl rand -hex 32)|" .env
-#   docker compose up
+#   cp .env.docker.example .env.docker
+#   sed -i "s|^ACEMUSIC_API_JWT_SECRET_KEY=.*|ACEMUSIC_API_JWT_SECRET_KEY=$(openssl rand -hex 32)|" .env.docker
+#   docker compose --env-file .env.docker up
+#
+# NOT `.env`: a checkout already has one holding the development configuration, which is a
+# different schema entirely — compose would load it and fail to parse it.
 #
 # This is the whole contract between the app and a host. Anything a deployment needs that
 # is not here belongs in the images, not in someone's shell history.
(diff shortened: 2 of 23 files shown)


frankbria:
## Third-party review (pre-PR)

Reviewer: `codex review --base main` (cross-family). `opencode`/GLM was tried on the last
PR and errored server-side, so codex is the pass of record here.

**Two findings, both confirmed against a running container and both fixed in `360fa58`.**

### [P2] `voice_training_root` never reached the worker

> `ACEMUSIC_API_VOICE_TRAINING_ROOT=/data/voice-training` is read by `ApiSettings` but
> never passed into `JobProcessor`, so the worker still uses its default
> `./acemusic-voice-training` under `/app` while running as the non-root `acemusic` user.

Correct, and worse than described. Verified in the container:

```
module default (was used) : ./acemusic-voice-training
setting (now passed on)   : /data/voice-training

$ mkdir -p ./acemusic-voice-training      # what the old behaviour would do
mkdir: cannot create directory './acemusic-voice-training': Permission denied
$ ls -ld /app
drwxr-xr-x 1 root root /app
```

So voice training in a container would not have merely bypassed the volume — it would have
failed outright. **My own volume check missed this**: it wrote the training file by hand,
which proves the mount and not that the application uses it. Fixed in `main.py`, pinned by
`tests/test_processor_wiring.py`, which also guards the sibling kwargs from being dropped
by a future edit to the same call.

### [P2] `localhost:8001` is the API container, not the host

> With `docker compose`, `localhost` here is the API container itself… both the
> availability probe and job processor will call the wrong container and generation will
> 503/fail.

Correct. The default is now `host.docker.internal` plus an `extra_hosts: host-gateway`
entry, since Linux does not resolve that name on its own. Verified:

```
host.docker.internal resolves inside the container -> 172.17.0.1
http://host.docker.internal:8013 -> 200
```

This is the third bug in this family on this branch — the first two (no compute URL passed
through at all; `ACEMUSIC_API_LOCAL_URL` and `ACEMUSIC_BASE_URL` being read by different
config systems) were found by running the stack. All three are the same shape: per-host
knowledge that was never written down, which is what the issue is about.

Re-verified after the fixes: 49 tests across the touched areas, `black`, `ruff`,
`actionlint`, and the stack brought back up healthy end to end.

frankbria:
## Review triage — `fca9990`

GLM: no defects, twice (initial pass and a re-check after `360fa58`). claude-review raised
three substantive items plus two minor ones, none blocking. All five addressed below.

### Fixed

**1. GHCR pull auth on the deploy host.** Correct and would have bitten on the first live
deploy: packages published with `GITHUB_TOKEN` are private by default regardless of repo
visibility, so `docker compose pull` 401s. Rather than provisioning a PAT — which would
undo the "no new credentials" property that is the whole reason for choosing GHCR — the
deploy now logs the host in with the job's own token before pulling, scoped to the run.
Making the packages public is documented as the alternative.

**5. `MONGO_PORT` footgun.** Right, and worth the two lines: this Mongo has no
authentication, so a bare `MONGO_PORT=27018` publishes the whole database on `0.0.0.0` of
a shared VPS. Caveat added next to both the compose entry and the `.env` template.

**2/3. Documented rather than silently assumed.** `docs/deployment.md` now records that
Compose **v2.20+** is required for the nested default that keeps the two ACE-Step URLs in
sync (verified on the version this stack was built with), and that **`build images` is not
yet a required check** — so a Dockerfile that fails to build shows red without blocking a
merge until branch protection is updated. Changing branch protection is a repo-admin action
I have deliberately not taken on your behalf.

### Not fixed, deliberately

**4. `docker-publish.yml` pins four actions by tag, not SHA.** Accurate, but that is
precisely the scope of **#425 (P2.8 Pin the remaining GitHub Actions by SHA)**. Doing half
of it here would leave that issue with a partial diff to reconcile.

### One more found while verifying the above

`docker compose` auto-loads `.env` from the project directory — and this repo already has
one, holding the *development* configuration under a different schema. `docker compose
config` fails outright on it, and the setup instructions I had written said
`cp .env.docker.example .env`, which would have overwritten a developer's file.

The stack's env file is now **`.env.docker`**, passed explicitly everywhere including
`deploy.sh`. Re-verified end to end with the documented command:

```
$ docker compose --env-file .env.docker up -d
api    Up (healthy)
mongo  Up (healthy)
web    Up (healthy)
$ curl …/api/v1/health
{"status":"ok","version":"0.1.0","uptime_seconds":14.08,"build_sha":"local"}
```

### Also noted

GLM flagged, as context rather than a defect, that `/data/voice-training` is a named volume
a host-side ACE-Step process would not see. Agreed on both counts — it is a topology
question for whoever binds that volume to a shared host path, and ACE-Step is out of scope
for this stack by design. It is recorded in `docs/deployment.md` alongside the other
known gaps.

PR #443, opened 2026-08-11
feat(#111): link multiple OAuth identities to one account

### LINKED ISSUE #111: P2.1 Auth: support multiple OAuth identities per account (robust cross-provider linking)
Follow-up from PR #109 (US-8.3 OAuth2 auth), raised by codex + Claude reviews.

## Problem
The `User` model holds a **single** OAuth identity (`oauth_provider` + `oauth_id`). When a verified email already exists under a different provider, PR #109 now **rejects the second-provider login with 409** (`src/acemusic/api/routers/auth.py` callback). This is deterministic and safe, but it means a user with the same email on, say, Google and Discord cannot sign in via the second provider.

## Fix direction
Support multiple OAuth identities per account:
- Move identities into a list/sub-collection (`[{provider, oauth_id}, ...]`) or a separate `OAuthIdentity` document keyed by `user_id`.
- On callback with a verified email that matches an existing account, **link** the new identity instead of 409.
- Migrate existing single-identity users.

## Status
Deferred from PR #109 by maintainer decision. Current behavior: 409 on second-provider email collision (no duplicate accounts, no 500).

frankbria:
Closes #111.

## What

A `User` now carries a list of `identities` (`{provider, oauth_id, linked_at}`), so the
same person can sign in with Google or Discord and land in the same account. A provider
reporting an email that already belongs to a different identity **links** that identity
instead of returning 409.

## The design fork, and why embedded

The issue offers "a list on `User`" or "a separate `OAuthIdentity` document" without
picking one. Embedded, for three reasons:

- **No index migration.** `oauth_provider`/`oauth_id` stay on the document as the *primary*
  identity, synced to `identities[0]`, so the existing partial-unique index keeps its
  meaning and the new one is purely additive. This matters more than it looks: Beanie
  creates indexes at startup but never rebuilds one whose spec changed, so redefining an
  index is a **silent no-op** on any database where it already exists — and a test database
  (dropped per run) would never reveal it.
- **Linking stays atomic.** One document, one `$push`. No cross-document consistency
  problem, and no second failure mode where an identity row outlives its user.
- **Direct repo precedent.** `Clip.visibility`/`is_public` does exactly this — a
  `model_validator(mode="before")` materialises the new shape from the legacy field, which
  is kept synced so existing queries keep matching without a migration.

## Migration: a read-time backfill, not a batch job

Accounts written before this have no `identities` key at all. Two things had to be true,
and both are pinned by tests:

1. **They stay reachable.** The identity lookup is an `$or` across the new array *and* the
   legacy pair — a query against the new shape alone would lock every existing user out of
   their own account.
2. **The array reaches the database.** The validator materialises it on read, but the
   linking lookup and the unique index query the *stored* array. It is persisted by a
   conditional update gated on the field still being absent, touching only that field —
   the `credits_reset_at` idiom, and for the same reason: a whole-document `save()` can
   revert a concurrent write.

## Security: this is a real trust transfer

Linking is gated on the provider having verified the address. `email_verified` defaults to
**False** in the service, so the ~50 existing callers keep today's refusal and only the
OAuth callback opts in — having already turned unverified addresses into a `403` one step
earlier.

Stated plainly, because it is the part worth disagreeing with: this treats *"the provider
says the email is verified"* as proof of the same human. It is the industry-standard trade
and it is what the issue asks for, but Discord's `verified` only means Discord confirmed
the address, and users can change their Discord email. Anyone who can get a supported
provider to verify a victim's address can link into that account.

Mitigated — not eliminated — by requiring verification and recording `linked_at`
provenance. The stronger alternative, requiring an authenticated session before attaching a
second provider, is a different user experience than the issue specifies, so it is flagged
rather than silently substituted.

## Verification

`docs/demos/us-111-multi-oauth-identity.md`, from `scripts/demo_us_111.py` against a real
MongoDB. Six scenarios, including the two that fail silently if wrong:

- A pre-#111 document inserted through the raw collection (`identities key present =
  False`) signs in, gets its array persisted, and then links a second provider.
- Six concurrent logins with the same new identity → one account, **one** array entry.

Plus: either provider signs into the same account; the primary identity does not move; an
unverified collision is still refused; both unique indexes exist and the second account
claiming an identity is rejected.

At the HTTP layer, `tests/test_auth_routes.py` drives the real FastAPI callback with a
stubbed provider exchange — the old `test_same_email_via_second_provider_rejected_409` is
now `..._links_to_one_account` (two callbacks, two providers, both 200, both tokens
carrying the same `sub`), and a new test asserts the unverified path still 403s before it
can reach the link.

Reviewed pre-PR by `codex review` (cross-family): no actionable findings.

## Known limitations

- **No unlink.** Nothing removes an identity. The issue does not ask for it, and a
  half-built unlink that can strand an account with zero identities is worse than none.
- **The API does not report which providers are linked.** `UserProfileResponse` maps fields
  explicitly and has never exposed `oauth_*`; a "signed in with" surface is a UI story.
- **The 409 detail is still discarded by the frontend** (`auth-context.tsx` throws a
  generic "Authentication failed"). Pre-existing, and now much rarer — the path is only
  reachable for an unverified collision, which the callback turns into a 403 anyway.
- **The old index is now partly redundant.** `(oauth_provider, oauth_id)` guards the
  primary identity and `(identities.provider, identities.oauth_id)` guards all of them.
  Keeping both is deliberate — dropping an index is the one operation Beanie will not do
  for you, so it wants a considered migration rather than a silent spec change.


frankbria:
## Third-party review (pre-PR)

Reviewer: `codex review --base main` (cross-family).

> The multi-identity OAuth changes preserve legacy lookup paths, gate linking on verified
> email, and add appropriate persistence and uniqueness coverage. Targeted service tests
> passed; no actionable correctness issues were identified in the diff.

**No findings.**

## Test scope, and one thing I stopped rather than finished

- **Unit: 1875 passed, 13 skipped** — the same count as `main`, so the model change breaks
  nothing that was passing.
- **Integration: 217 passed**, over every suite that constructs `User` documents:
  `test_auth_routes`, `test_auth_models`, `test_user_service`, `test_oauth_identities`,
  `test_users_api`, `test_workspaces_api`, `test_credits_api`, `test_credit_buckets`,
  `test_credit_reset`, `test_tier_enforcement_api`, `test_mongodb`.

I started a full local `-m "integration or not integration"` run and **killed it at 32%**.
It had stopped making progress for ~12 minutes, and the local integration set reaches real
external services (ElevenLabs among them) that CI does not have configured — so it was
neither representative nor free. The bounded run above covers the risk this diff actually
carries, and CI runs the complete integration set in a clean environment.

Flagging it rather than quietly reporting "tests pass": the number above is a subset I
chose, not the whole suite.

frankbria:
## Review triage — `b0d2708`

Both reviewers independently found the same defect, and they are right. Fixed.

### The finding

> Full-document `user.save()` in the known-identity branch reverts a concurrently-linked
> identity (`$push` from `_link_identity`): silent data loss of the linked provider and its
> `linked_at` provenance.

Verified before fixing rather than taken on trust: `User.Settings` sets no
`use_state_management`, so Beanie's `save()` is a replace from the in-memory model, not a
diff. And that `save()` runs on **every** login — the branch is entered whenever the
reported email is unowned or already ours, which is the normal repeat-login case, not just
an email change.

So the window is: request A loads the account, request B links a second provider, request A
finishes and writes back its pre-link `identities`. The user simply stops being able to
sign in with that provider, and nothing errors.

The race is newly reachable because this branch introduces the concurrent `$push` writer —
before it, nothing else touched the array. What makes it a fair hit rather than an unlucky
one is that **the principle was already written down two functions above**, in
`_persist_identities`' own docstring ("A full `save()` would push a whole stale document
back and could revert a concurrent write"), and I did not apply it ten lines below.

### What changed

- `_refresh_profile()` — a targeted `$set` of `email`/`name`/`updated_at`, replacing the
  `save()`. The in-memory instance is kept consistent for the caller.
- **`update_user_profile()` got the same treatment**, which neither review flagged. It had
  the identical hazard: a profile PATCH racing an OAuth login writes back a pre-link
  `identities` array and drops the provider the same way. Narrower window, identical silent
  loss — not worth leaving as a known-bad twin of the bug being fixed.
- `test_a_repeat_login_does_not_clobber_a_concurrently_linked_identity` pins it.

### The test has teeth

Mutation-checked rather than assumed: restoring the whole-document `save()` fails that test
and **only** that test (`1 failed, 10 passed`); with the fix, `94 passed` across
`test_oauth_identities`, `test_user_service`, `test_users_api` and `test_auth_routes`.

### Also noted

Both reviews analyzed and discarded an unhandled `DuplicateKeyError` on `_link_identity`'s
`$push` as unreachable — an OAuth identity reports one email, so the same
`(provider, oauth_id)` cannot be claimed by a second account under a different email; the
email unique index conflicts first. Agreed, and left alone.

frankbria:
## Review triage — `f2c9eae`

GLM is right, and the finding lands harder than it reads: my previous commit message said a
known-bad twin was "not worth leaving" **while leaving six of them in `billing.py`**. I swept
`services/users.py`, not the codebase.

### Fixed

All six whole-document `await user.save()` calls in `billing.py` now go through one
`_persist_user_fields()` helper doing a targeted `$set`.

`ensure_customer` is the one that matters most, exactly as described: a Stripe round-trip
sits between loading the user and writing it back, so a musician who links Discord in
another tab during a checkout loses the link when that call returns. The five webhook
handlers each hold a `User` across at least one `await` — narrower window, identical silent
loss.

A grep for `user.save()` across `src/` is now empty. The remaining `.save()` calls in the
codebase are on other document types (`Clip`, `Job`, `VoiceModel`, `Workspace`, …), which
`_link_identity` never touches, so they are out of this hazard class.

### Why it needed its own test

`tests/test_billing_service.py` + `tests/test_billing_api.py` — **66 tests — pass either
way**. They never race a link, so green there proved nothing about this. Covered instead by
`test_a_billing_write_does_not_clobber_a_concurrently_linked_identity`, which exercises the
widest window directly.

That is the second time on this PR that the useful signal came from a test written *for the
race*, not from the suite being green.

### Also re-confirmed

The `DuplicateKeyError`-on-`$push` candidate stays discarded for the reason all three
reviews gave independently: an OAuth identity reports one email, so the same
`(provider, oauth_id)` cannot be claimed by a second account under a different email — the
email unique index conflicts first.

